PR #33, opened 2026-04-22
Add OAuth login flow with JolliAuthService, config store, and UI banners

jordangreen-jolli:
## E2E Test (8)
<details>
<summary><strong>1. Sign in to Jolli account from the tool window</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a valid Jolli account. Make sure you are currently signed out (no active session in the IDE).

**Steps:**
1. Open the Jolli Memory panel in the IDE (look for it in the side panel or bottom bar).
2. Check that a sign-in banner appears at the top of the panel.
3. Click the "Sign In" button in the banner.
4. Complete the login flow in the browser window that opens.
5. Return to the IDE and look at the Jolli Memory panel.

**Expected Results:**
- The sign-in banner disappears after a successful login.
- A status bar appears showing that you are signed in, along with a "Sign Out" button.

</blockquote>
</details>
<details>
<summary><strong>2. Sign out of Jolli account from the tool window</strong></summary>
<br>
<blockquote>

**Preconditions:** Be signed in to a Jolli account in the IDE.

**Steps:**
1. Open the Jolli Memory panel in the IDE.
2. Confirm the signed-in status bar is visible with a "Sign Out" button.
3. Click the "Sign Out" button.
4. Observe the panel.

**Expected Results:**
- The signed-in status bar disappears.
- The sign-in banner reappears at the top of the panel.

</blockquote>
</details>
<details>
<summary><strong>3. Sign in to Jolli account from the Settings page</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a valid Jolli account. Make sure you are currently signed out.

**Steps:**
1. Open Settings and navigate to Tools > Jolli Memory.
2. Find the Account section on the settings page.
3. Check that a "Sign In" button and a status label are visible.
4. Click the "Sign In" button and complete the login in the browser.
5. Return to the Settings page.

**Expected Results:**
- The status label updates to show you are signed in.
- The button changes to "Sign Out".
- The Jolli API Key field label updates to indicate the key is optional now that an account is active.

</blockquote>
</details>
<details>
<summary><strong>4. Sign out of Jolli account from the Settings page</strong></summary>
<br>
<blockquote>

**Preconditions:** Be signed in to a Jolli account in the IDE.

**Steps:**
1. Open Settings and navigate to Tools > Jolli Memory.
2. Find the Account section and confirm the "Sign Out" button is visible.
3. Click "Sign Out".
4. Observe the Account section.

**Expected Results:**
- The status label updates to show you are signed out.
- The button changes back to "Sign In".
- The Jolli API Key field label no longer shows the "optional" indicator.

</blockquote>
</details>
<details>
<summary><strong>5. Generate a commit message using a Jolli account (no Anthropic key)</strong></summary>
<br>
<blockquote>

**Preconditions:** Be signed in to a Jolli account. Ensure no Anthropic API key is set in Settings or as an environment variable. Have at least one file staged for commit in a Git repository.

**Steps:**
1. Stage a change in your project.
2. Open the commit dialog or trigger the AI commit message action.
3. Wait for the AI to generate a commit message.
4. Check the result.

**Expected Results:**
- The commit message is generated successfully without any error about a missing API key.
- The generated message appears in the commit message field.

</blockquote>
</details>
<details>
<summary><strong>6. Error shown when no credentials are available at commit time</strong></summary>
<br>
<blockquote>

**Preconditions:** Be signed out of Jolli. Ensure no Anthropic API key is configured in Settings and no environment variable is set. Have at least one file staged for commit.

**Steps:**
1. Stage a change in your project.
2. Trigger the AI commit message action.
3. Observe what happens.

**Expected Results:**
- An error dialog appears stating that no credentials are available.
- The message directs you to sign in to Jolli or configure an Anthropic API key in Settings.
- No commit message is generated.

</blockquote>
</details>
<details>
<summary><strong>7. OAuth login rejects a tampered or replayed callback (CSRF protection)</strong></summary>
<br>
<blockquote>

**Preconditions:** Be signed out of Jolli. Have a way to send a custom web request to localhost (e.g. a browser or a tool like Postman).

**Steps:**
1. Trigger the Sign In flow from the tool window or Settings page.
2. Before completing the login in the browser, manually open a new browser tab and try to visit the local callback URL with a made-up or blank state value in the address bar.
3. Observe the response in the browser tab.
4. Check whether the IDE shows a successful sign-in.

**Expected Results:**
- The manually crafted callback is rejected and the browser shows an error (such as a 403 or similar failure message).
- The IDE does not show a signed-in status — the sign-in banner remains visible.
- The local callback server shuts down after the failed attempt and does not remain open.

</blockquote>
</details>
<details>
<summary><strong>8. Settings save does not erase the stored auth token</strong></summary>
<br>
<blockquote>

**Preconditions:** Be signed in to a Jolli account so an auth token is stored.

**Steps:**
1. Open Settings and navigate to Tools > Jolli Memory.
2. Change any unrelated setting (for example, update the model selection or a text field).
3. Click OK or Apply to save.
4. Close and reopen Settings.
5. Check the Account section.

**Expected Results:**
- The account status still shows you are signed in after saving.
- The auth token has not been cleared by the save operation.

</blockquote>
</details>

---

## Topics (5)
<details>
<summary><strong>01 · OAuth login flow with CSRF state parameter protection</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The login callback accepted any GET request containing a token and wrote it straight to disk, with no verification that the callback came from the browser tab the plugin opened. Any local process or browser tab could silently sign the IDE into an attacker-controlled account, after which the plugin would push diffs and transcripts to attacker-owned storage.

- **State embedded in callback URL, not as a top-level login param**: The Jolli web app stores the entire `cli_callback` URL and appends `token=` to it on redirect, so a separate `state=` param on the login URL would be lost. Embedding state inside the callback URL ensures it survives the round-trip automatically.
- **Shutdown on mismatch, not just rejection**: Leaving the server open after a failed state check would keep the exposure window alive until timeout. Shutting down immediately closes the window as soon as an invalid request is detected.
- **`finally` block for shutdown**: An early `return` inside the callback handler would skip `shutdown()`, leaving the server running. The `finally` block guarantees cleanup regardless of which code path exits.

**✅ What Was Implemented**

- `JolliAuthService.kt` (new): full OAuth callback server — generates a 128-bit cryptographically random state token before opening the browser, embeds it in the callback URL (URL-encoded inside `cli_callback` so the web app forwards it back), and rejects any callback where the returned state doesn't match with a 403 response
- State mismatch and error paths both call `shutdown()` via a `finally` block so the server never stays open past a failed check
- 60-second timeout was already present; this change adds the state validation on top of it
- `JolliAuthServiceTest.kt` (new): verifies the timeout fires `onError` and never fires `onSuccess`

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliAuthService.kt`
- `intellij/src/test/kotlin/ai/jolli/jollimemory/services/JolliAuthServiceTest.kt`

</blockquote>
</details>
<details>
<summary><strong>02 · Jolli account sign-in UI in tool window and settings page</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

There was no way for users to sign in to their Jolli account from inside the IDE. Credentials had to be configured manually as raw API keys, with no guided flow and no visible account status.

- **Banner in tool window, not just Settings**: Placing the sign-in prompt where users already work (the tool window) makes it discoverable without requiring them to navigate to preferences first, improving adoption.
- **Auth listener pattern for UI sync**: Rather than polling or requiring manual refresh, `JolliAuthService` maintains a listener list so all UI surfaces update immediately when auth state changes, keeping the signed-in bar and banner in sync without coupling them directly.
- **Jolli API Key field kept but marked optional when signed in**: Removing the field entirely would break existing users who configured a key manually. Relabeling it as optional preserves backward compatibility while communicating the new preferred path.

**✅ What Was Implemented**

- `JolliMemoryToolWindowFactory.kt`: adds a sign-in banner at the top of the tool window (shown when signed out) and a signed-in status bar with a Sign Out button (shown when signed in); both toggle via `JolliAuthService.addAuthListener`
- `JolliMemoryConfigurable.kt`: adds an Account section to the Settings page with a dynamic Sign In / Sign Out button and status label; the Jolli API Key label updates to indicate it is optional when an account is active
- `StatusPanel.kt`: removes the warning row that appeared when no Jolli API key was configured, since sign-in is now the primary path
- `SummaryPanel.kt`: updates the missing-key warning message to direct users to sign in or use Settings

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/JolliMemoryToolWindowFactory.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/settings/JolliMemoryConfigurable.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/StatusPanel.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/auth/JolliConfigStore.kt`

</blockquote>
</details>
<details>
<summary><strong>03 · LLM proxy routing layer to support Jolli-managed credentials</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

All LLM calls were hardwired to the Anthropic API and required users to supply their own API key. There was no way to route calls through the Jolli backend for users who sign in with an account instead of providing a key directly.

- **Proxy sends action key + params, not a pre-built prompt**: The Jolli backend owns the prompt templates in proxy mode, so the plugin only needs to send structured data. This lets the backend iterate on prompts without plugin releases, and avoids sending large prompt strings over the wire when the backend can construct them.
- **Anthropic key takes priority over proxy**: Users who already have a direct key configured get the same behavior as before; the proxy is only used as a fallback, minimising disruption to existing setups.
- **Single routing object rather than per-call branching**: Centralising the credential resolution and dispatch in one place makes the priority logic easy to audit and change, rather than duplicating it across five call sites.

**✅ What Was Implemented**

- `LlmClient.kt` (new): central routing object with `resolveCredentialSource` (priority: Anthropic config key > `ANTHROPIC_API_KEY` env var > Jolli proxy) and `callLlm` that dispatches to either direct Anthropic or `JolliApiClient.callLlmProxy`
- `JolliApiClient.kt`: adds `callLlmProxy` (POST to `/api/push/llm/complete` with action key + params) and `resolveToken` (prefers explicit key, falls back to stored OAuth token)
- `Summarizer.kt`: all five LLM call sites (`generateSummary`, `generateCommitMessage`, `generateE2eTest`, `translateToEnglish`, `generateSquashMessage`) refactored to go through `LlmClient.callLlm`; each now accepts an optional `jolliApiKey` param
- `PlanProgressEvaluator.kt`: same refactor applied
- `CommitAIAction.kt`, `SquashAction.kt`, `PostCommitHook.kt`: credential check and error messages updated to reflect that either an Anthropic key or a Jolli account is sufficient

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/core/LlmClient.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/core/Summarizer.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/core/PlanProgressEvaluator.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/hooks/PostCommitHook.kt`

</blockquote>
</details>
<details>
<summary><strong>04 · Windows PATH fix to make git hooks find shell utilities</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

When IntelliJ runs a git commit, the JVM process that launches git hooks does not inherit the user's shell PATH. On Windows this means tools like `sed` — which live in Git's `usr/bin` directory — are not found, causing hook scripts to fail silently.

- **Locate Git root dynamically via `where git`**: Hard-coding a path like `C:\Program Files\Git` would break for users with non-standard install locations. Deriving the root from the actual `git.exe` location works regardless of where Git is installed.
- **Append rather than replace PATH**: Replacing the entire PATH would drop other entries the JVM inherited. Appending `usr/bin` is the minimal, safe change that adds only what is missing.

**✅ What Was Implemented**

`GitOps.kt`: added `resolveWindowsPath()` — on Windows, runs `where git` to locate `git.exe`, walks up two directory levels to the Git install root, checks for `usr/bin`, and appends it to the inherited `PATH` in the `ProcessBuilder` environment. The existing Unix path-resolution logic is unchanged.

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/GitOps.kt`

</blockquote>
</details>
<details>
<summary><strong>05 · Global config store for auth token and space, shared with CLI</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The OAuth token received after sign-in needed a persistent home that both the IntelliJ plugin and the CLI tool could read, and the existing per-project config file was not the right place for account-level credentials.

- **Separate file for space (`space.json`) vs. token (`config.json`)**: The CLI already uses this layout, so matching it avoids a migration and lets both tools share state without coordination.
- **Merge-on-save in settings dialogs**: Previously, saving settings constructed a fresh config object, which would silently erase any fields the dialog didn't manage (including the new auth token). Switching to load-then-copy preserves all fields not explicitly changed by the user.

**✅ What Was Implemented**

- `JolliConfigStore.kt` (new): read/write helpers for `~/.jolli/jollimemory/config.json` (auth token) and `~/.jolli/space.json` (space name), both shared with the CLI
- `Types.kt`: added `authToken` field to `JolliMemoryConfig`
- `SessionTracker.kt`: `mergeConfig` now preserves `authToken` when merging partial updates
- `JolliMemoryConfigurable.kt` and `SettingsDialog.kt`: settings saves now load the existing global config and use `copy()` to merge changes, preventing unrelated fields (including `authToken`) from being wiped on save

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/auth/JolliConfigStore.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/core/Types.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/settings/JolliMemoryConfigurable.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/SettingsDialog.kt`

</blockquote>
</details>

---

*Generated by Jolli Memory · April 28, 2026 at 6:06 PM*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,8 @@ Thumbs.db
 .codex/
 .opencode/
 .DS_Store
+.idea/
+*.iml
 **/dist/
 **/coverage/
 **/node_modules/
--- a/intellij/build.gradle.kts
+++ b/intellij/build.gradle.kts
@@ -375,6 +375,17 @@ tasks {
         }
     }
 
+    // Workaround: IntelliJ Platform Gradle Plugin 2.5.0 fails to parse the Java
+    // version from the downloaded IDE runtime, producing "JavaLanguageVersion must
+    // be a positive integer, not ''". Explicitly set the JVM launcher for affected tasks.
+    withType<JavaExec> {
+        javaLauncher.set(
+            project.the<JavaToolchainService>().launcherFor {
+                languageVersion.set(JavaLanguageVersion.of(21))
+            }
+        )
+    }
+
     test {
         useJUnitPlatform()
     }
--- a/intellij/src/main/kotlin/ai/jolli/jollimemory/actions/CommitAIAction.kt
+++ b/intellij/src/main/kotlin/ai/jolli/jollimemory/actions/CommitAIAction.kt
@@ -59,9 +59,9 @@ class CommitAIAction : AnAction() {
         }
 
         val config = SessionTracker.loadConfig(cwd)
-        if (config.apiKey.isNullOrBlank()) {
+        if (config.apiKey.isNullOrBlank() && config.jolliApiKey.isNullOrBlank() && System.getenv("ANTHROPIC_API_KEY").isNullOrBlank()) {
             Messages.showErrorDialog(project,
-                "Anthropic API key not configured.\nSet it in the STATUS panel or Settings > Tools > Jolli Memory.",
+                "No LLM credentials available.\nSign in to Jolli or configure an Anthropic API key in Settings > Tools > Jolli Memory.",
                 "Jolli Memory")
             return
         }
@@ -116,13 +116,14 @@ class CommitAIAction : AnAction() {
                     val branch = git.getCurrentBranch() ?: "unknown"
 
                     // Step 5: Generate AI message
-                    indicator.text = "Calling Anthropic API..."
+                    indicator.text = "Generating AI commit message..."
                     val message = Summarizer.generateCommitMessage(CommitMessageParams(
                         stagedDiff = diff,
                         branch = branch,
                         stagedFiles = selectedPaths,
                         apiKey = config.apiKey,
                         model = config.model,
+                        jolliApiKey = config.jolliApiKey,
                     ))
 
                     // Step 6: Show dialog on EDT
--- a/intellij/src/main/kotlin/ai/jolli/jollimemory/actions/SquashAction.kt
+++ b/intellij/src/main/kotlin/ai/jolli/jollimemory/actions/SquashAction.kt
@@ -114,7 +114,7 @@ class SquashAction : AnAction() {
                     }
                     try {
                         log.info("Generating squash message via AI…")
-                        message = Summarizer.generateSquashMessage(commitData, ticketId, isFullSquash, config.apiKey, config.model)
+                        message = Summarizer.generateSquashMessage(commitData, ticketId, isFullSquash, config.apiKey, config.model, config.jolliApiKey)
                         log.info("Squash message generated: %s", message)
                     } catch (ex: Exception) {
                         log.error("Failed to generate squash message: %s", ex.message)
--- a/intellij/src/main/kotlin/ai/jolli/jollimemory/auth/JolliConfigStore.kt
+++ b/intellij/src/main/kotlin/ai/jolli/jollimemory/auth/JolliConfigStore.kt
@@ -1,0 +1,68 @@
+package ai.jolli.jollimemory.auth
+
+import ai.jolli.jollimemory.core.JolliMemoryConfig
+import ai.jolli.jollimemory.core.SessionTracker
+import com.google.gson.Gson
+import com.google.gson.GsonBuilder
+import java.nio.file.Files
+import java.nio.file.Path
+
+/**
+ * Read/write store for general Jolli config files under ~/.jolli/ (shared with CLI).
+ * Auth token and space are stored in ~/.jolli/jollimemory/config.json.
+ */
+object JolliConfigStore {
+
+    private val gson: Gson = GsonBuilder().setPrettyPrinting().create()
+
+    private fun jolliDir(): Path =
+        Path.of(System.getProperty("user.home"), ".jolli")
+
+    private fun spacePath(): Path = jolliDir().resolve("space.json")
+
+    private data class SpaceConfigFile(
+        val space: String? = null,
+    )
+
+    /** Load auth token from env var or ~/.jolli/jollimemory/config.json. */
+    fun loadAuthToken(): String? {
+        System.getenv("JOLLI_AUTH_TOKEN")?.takeIf { it.isNotBlank() }?.let { return it }
+
+        val config = SessionTracker.loadConfigFromDir(SessionTracker.getGlobalConfigDir())
+        return config.authToken?.takeIf { it.isNotBlank() }
+    }
+
+    /** Save auth token to ~/.jolli/jollimemory/config.json. */
+    fun saveAuthToken(token: String) {
+        val globalDir = SessionTracker.getGlobalConfigDir()
+        val existing = SessionTracker.loadConfigFromDir(globalDir)
+        SessionTracker.saveConfigToDir(existing.copy(authToken = token), globalDir)
+    }
+
+    /** Clear auth token from ~/.jolli/jollimemory/config.json. */
+    fun clearAuthToken() {
+        val globalDir = SessionTracker.getGlobalConfigDir()
+        val existing = SessionTracker.loadConfigFromDir(globalDir)
+        SessionTracker.saveConfigToDir(existing.copy(authToken = null), globalDir)
+    }
+
+    /** Load space from ~/.jolli/space.json. */
+    fun loadSpace(): String? {
+        val path = spacePath()
+        if (!Files.exists(path)) return null
+        return try {
+            val json = Files.readString(path)
+            gson.fromJson(json, SpaceConfigFile::class.java)?.space
+        } catch (_: Exception) {
+            null
+        }
+    }
+
+    /** Save space to ~/.jolli/space.json. */
+    fun saveSpace(space: String) {
+        val dir = jolliDir()
+        Files.createDirectories(dir)
+        val data = SpaceConfigFile(space = space)
+        Files.writeString(spacePath(), gson.toJson(data))
+    }
+}
--- a/intellij/src/main/kotlin/ai/jolli/jollimemory/auth/JolliUrlConfig.kt
+++ b/intellij/src/main/kotlin/ai/jolli/jollimemory/auth/JolliUrlConfig.kt
@@ -1,0 +1,17 @@
+package ai.jolli.jollimemory.auth
+
+/**
+ * Resolves the Jolli login URL.
+ * Priority: JOLLI_URL env var > jolli.url system property > default.
+ */
+// TODO: test comment for jollimemory summary verification
+object JolliUrlConfig {
+
+    private const val DEFAULT_URL = "https://jolli.ai"
+
+    fun getJolliUrl(): String {
+        return System.getenv("JOLLI_URL")?.takeIf { it.isNotBlank() }
+            ?: System.getProperty("jolli.url")?.takeIf { it.isNotBlank() }
+            ?: DEFAULT_URL
+    }
+}
--- a/intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/GitOps.kt
+++ b/intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/GitOps.kt
@@ -22,16 +22,50 @@ class GitOps(private val projectDir: String) {
      */
     private val shellPath: String by lazy {
         try {
-            val shell = System.getenv("SHELL") ?: "/bin/zsh"
-            val proc = ProcessBuilder(shell, "-l", "-c", "echo \$PATH")
+            if (System.getProperty("os.name").lowercase().contains("win")) {
+                resolveWindowsPath()
+            } else {
+                val shell = System.getenv("SHELL") ?: "/bin/zsh"
+                val proc = ProcessBuilder(shell, "-l", "-c", "echo \$PATH")
+                    .redirectErrorStream(true)
+                    .start()
+                val output = proc.inputStream.bufferedReader().use { it.readText().trim() }
+                proc.waitFor(5, TimeUnit.SECONDS)
+                if (output.isNotBlank()) output else System.getenv("PATH") ?: ""
+            }
+        } catch (_: Exception) {
+            System.getenv("PATH") ?: ""
+        }
+    }
+
+    /**
+     * On Windows, the system PATH usually has Git's cmd/ dir but not usr/bin/
+     * where tools like sed and awk live. Find the Git install root and append
+     * usr/bin so git hooks that need those tools work correctly.
+     */
+    private fun resolveWindowsPath(): String {
+        val basePath = System.getenv("PATH") ?: ""
+        try {
+            val proc = ProcessBuilder("where", "git")
                 .redirectErrorStream(true)
                 .start()
             val output = proc.inputStream.bufferedReader().use { it.readText().trim() }
             proc.waitFor(5, TimeUnit.SECONDS)
-            if (output.isNotBlank()) output else System.getenv("PATH") ?: ""
-        } catch (_: Exception) {
-            System.getenv("PATH") ?: ""
-        }
+
+            // Find Git install root from the first git.exe path
+            // e.g. "C:\Program Files\Git\cmd\git.exe" -> "C:\Program Files\Git"
+            val gitExePath = output.lines().firstOrNull { it.endsWith("git.exe") }
+            if (gitExePath != null) {
+                val gitRoot = File(gitExePath).parentFile?.parentFile
+                if (gitRoot != null) {
+                    val usrBin = File(gitRoot, "usr${File.separator}bin")
+                    if (usrBin.isDirectory) {
+                        return "$basePath${File.pathSeparator}${usrBin.absolutePath}"
+                    }
+                }
+            }
+        } catch (_: Exception) { }
+        return basePath
     }
 
     /**
--- a/intellij/src/main/kotlin/ai/jolli/jollimemory/core/LlmClient.kt
+++ b/intellij/src/main/kotlin/ai/jolli/jollimemory/core/LlmClient.kt
@@ -1,0 +1,110 @@
+package ai.jolli.jollimemory.core
+
+import ai.jolli.jollimemory.services.JolliApiClient
+
+/**
+ * LlmClient — Routes LLM calls between direct Anthropic API and Jolli proxy.
+ *
+ * Credential priority: Anthropic API key > ANTHROPIC_API_KEY env var > Jolli API key (proxy).
+ * Direct mode: caller provides a pre-built prompt, we call AnthropicClient.
+ * Proxy mode: we send action + params to Jolli backend which owns the prompt template.
+ */
+object LlmClient {
+
+    private val log = JmLogger.create("LlmClient")
+
+    enum class CredentialSource { ANTHROPIC_CONFIG, ANTHROPIC_ENV, JOLLI_PROXY }
+
+    data class LlmCallResult(
+        val text: String?,
+        val model: String?,
+        val inputTokens: Int,
+        val outputTokens: Int,
+        val apiLatencyMs: Long,
+        val stopReason: String?,
+    )
+
+    /**
+     * Determines which credential source to use.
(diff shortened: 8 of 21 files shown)


flyerli2025:
Critical (must fix before merge)
C1. OAuth callback has no state / PKCE — local CSRF + token injection
[JolliAuthService.kt:43-107](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliAuthService.kt#L43-L107)
The login URL only carries cli_callback, and the /callback handler accepts any GET containing token= and writes it straight to disk. Any local process that can probe ports, or any browser tab the user has open (/callback?token=ATTACKER"> works because it's an unauthenticated GET), can silently sign the IDE in to an attacker's account — after which the plugin will push diffs/transcripts to attacker-controlled storage. Fix: generate a 128-bit state before opening the browser, embed it in loginUrl, and require strict equality on callback. Add a timeout to close the exposure window. Ideally upgrade to PKCE.

C2. PR description does not match the implementation — jolliAutoApiKey / resolveTokenList() don't exist
[Types.kt:284-291](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/core/Types.kt#L284-L291) only adds authToken; grep -rn "jolliAutoApiKey\|resolveTokenList" intellij/src/main/kotlin/ returns zero hits. The OAuth callback writes the server-returned key into the same jolliApiKey field used by the manual settings form ([JolliAuthService.kt:74-87](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliAuthService.kt#L74-L87)). Every UX/security claim in summary section 03 of the PR description is false against the code. Either ship the missing commit or remove that section from the description.

C3. signOut() wipes the user's manually entered Jolli API key
[JolliAuthService.kt:120-125](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliAuthService.kt#L120-L125) — existing.copy(authToken = null, jolliApiKey = null). Because there is no auto/manual split (see C2), every sign-out destroys whatever the user typed in Settings. Fix: don't touch jolliApiKey here, or land the field-split first.

C4. Tool-window sign-in failure is fully silent — msg is dropped
[JolliMemoryToolWindowFactory.kt:244-249](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/JolliMemoryToolWindowFactory.kt#L244-L249)


onError = { msg -> SwingUtilities.invokeLater { isEnabled = true; text = "Sign In" } }
msg is never shown. Compare to the settings page at [JolliMemoryConfigurable.kt:111-116](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/settings/JolliMemoryConfigurable.kt#L111-L116) which calls Messages.showErrorDialog — inconsistent UX. Fix: use Notifications.Bus.notify or the same dialog.

Important (should fix before merge)
I1. Settings "Sign In" button stays disabled forever
[JolliMemoryConfigurable.kt:101-117](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/settings/JolliMemoryConfigurable.kt#L101-L117) — on click it sets isEnabled = false; text = "Signing in...", but neither branch of updateAccountUI() resets isEnabled back to true. After a failed login the user has to close and reopen Settings. One-line fix: set accountButton?.isEnabled = true unconditionally at the top of updateAccountUI().

I2. JolliAuthService.addAuthListener has no removal API — listener leak
[JolliAuthService.kt:25-29](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliAuthService.kt#L25-L29) is a process singleton; every createToolWindowContent call at [JolliMemoryToolWindowFactory.kt:268](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/JolliMemoryToolWindowFactory.kt#L268) appends a closure that captures the panel and Project. Across project reloads, listeners accumulate and keep disposed Swing components alive. Fix: return a Disposable from addAuthListener (or expose removeAuthListener) and register it on the tool window's Disposer.

I3. HTTP server executor is never shut down
[JolliAuthService.kt:52,127-130](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliAuthService.kt#L52) — shutdown() only calls server?.stop(0). Executors.newSingleThreadExecutor() is non-daemon, so each login attempt leaks a thread; the IDE may stall on shutdown. Fix: keep the ExecutorService reference and call shutdownNow() inside shutdown().

I4. No callback timeout
If the browser never returns, the server stays open forever and the UI button is stuck on "Signing in..." (tool window) or isEnabled = false (settings) indefinitely. Fix: schedule a 5-minute ScheduledExecutorService task in login() that calls shutdown(); onError("Login timed out — no callback received from browser.").

I5. server.stop(0) can truncate the response
[JolliAuthService.kt:91-104](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliAuthService.kt#L91-L104) — finally { shutdown() } calls stop(0), which closes the listening socket immediately and does not wait for in-flight exchanges. Even though responseBody.use {} closed the application stream, the OS socket buffer may not have flushed; the browser can see a connection reset / blank page. Fix: stop(1) gives 1 second of graceful drain.

I6. SummaryPanel push gate ignores authToken — signed-in users with no jolliApiKey are blocked
[SummaryPanel.kt:252-260](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/SummaryPanel.kt) still checks if (config.jolliApiKey.isNullOrBlank()) and derives the URL from parseJolliApiKey(config.jolliApiKey!!). If the OAuth response omits jolli_api_key (the callback treats it as optional), the user is "Signed in" but cannot push. Fix: route through JolliApiClient.resolveToken(config.jolliApiKey) and fall back to JolliUrlConfig for the base URL when no key meta is available.

I7. After sign-out, the Settings jolliApiKeyField is not refreshed
signOut() writes jolliApiKey = null (see C3), but the Configurable's savedJolliApiKey and text field don't update. Next apply() writes the stale field value back. Fix: have Configurable register an auth listener that calls loadFromConfig() on sign-in/out (currently only the tool window registers).

I8. URL building uses URI.toASCIIString() instead of percent-encoding
[JolliAuthService.kt:58](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliAuthService.kt#L58) — works in the current shape, but : and / remain in a query value. As soon as anyone adds a second query param to the callback, the server-side parse breaks. Fix: URLEncoder.encode(callbackUrl, StandardCharsets.UTF_8).

Suggestions
S1. Zero tests added
intellij/src/test/kotlin already has ~22 test files (including JolliApiClientTest.kt, SessionTrackerTest.kt) — testing is the convention here. This PR adds none. Highest-value targets:

JolliApiClient.resolveToken priority chain
JolliAuthService.parseQuery edge cases (empty segments, %ZZ, duplicate keys)
getErrorMessage for known codes
signOut() semantics — would catch C3 immediately
JolliConfigStore.loadAuthToken env-var priority
S2. JolliConfigStore.loadSpace and JolliApiClient.parseJolliApiKey swallow exceptions silently
[JolliConfigStore.kt:55-58](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/auth/JolliConfigStore.kt#L55-L58) — catch (_: Exception) { null } treats corrupt JSON, IO errors, and permission failures all as "not configured." At minimum log via log.warn(..., e) so users can find the problem in the IDE log.

S3. saveSpace failure leaves a partial-success state
[JolliAuthService.kt:84-89](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliAuthService.kt#L84-L89) — order is saveConfigToDir(authToken=...) then saveSpace. If the second step throws, the outer catch reports "Unexpected error" but the token has already been persisted. Fix: wrap saveSpace in its own try/catch; keep the token but surface "Signed in but failed to persist space."

S4. build.gradle.kts workaround has no tracking link
[build.gradle.kts:378-387](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/build.gradle.kts#L378-L387) — pinning the JDK 21 launcher around an IntelliJ Platform Gradle Plugin 2.5.0 bug is reasonable, but add a TODO with an upstream issue link so this can be removed once the plugin is fixed.

Strengths
Server binds 127.0.0.1 rather than 0.0.0.0 ([JolliAuthService.kt:46](vscode-webview://0n7r23ldi54otglcapt981i045ccakvn04oiei60ttf2ss1t7lnn/intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliAuthService.kt#L46)) — correctly limits exposure to local processes.
Random port InetSocketAddress(..., 0) — avoids port conflicts.
Shares the ~/.jolli/ config directory with the CLI — consistent across tools.
JolliUrlConfig's env / system-property / default precedence is clean and easy to override for dev/staging.
Error HTML escapes < and > (though & and " are still missing).
Recommended order
Resolve the PR-description / implementation mismatch first (C2, C3): are you adding the missing commit, or rewriting the description? This decides the scope of the rest.
Fix C1 (state/PKCE) + C4 (display msg) + I1 (button disabled).
Fix the resource-lifecycle cluster (I2 listener, I3 executor, I4 timeout, I5 stop(0)).
Add 5–8 high-value tests (S1) — they'll catch C3 and I6 as regressions for free.
Polish: S2/S3 logging, I8 encoding, S4 TODO.

summerfang-jolli (approved):

PR #49, opened 2026-04-26
Add local Knowledge Base explorer with folder storage, dual-write, migration, and UI

summerfang-jolli:
## Summary

- Extract storage abstraction layer (StorageProvider interface, OrphanBranchStorage, FolderStorage)
- Implement dual-write mode: orphan branch + KB folder with markdown generation
- Add MigrationEngine for bulk data migration from orphan branch to folder
- Knowledge Base explorer panel in JOLLI tool window with C/P/N badges
- Right-click context menu: New Folder, New File, Import, Rename, Move, Delete
- Drag and drop support for files and folders
- Auto-migration on plugin startup
- Settings UI: KB folder path, sort order, Migrate button
- Fix CommitsPanel "disabled" after Enable on empty branches
- Fix panel refresh on branch switch
- Technical journey documentation (Chinese + English)

## Test plan

- [ ] Install plugin, verify Knowledge Base tab appears in JOLLI tool window dropdown
- [ ] Enable dual-write (storageMode: dual-write in config.json)
- [ ] Make a commit with AI tool, verify markdown appears in KB folder
- [ ] Click Migrate to Knowledge Base in Settings, verify old data migrated
- [ ] Browse KB tree, click commit file to open JCEF webview
- [ ] Right-click actions: create folder, create file, rename, delete
- [ ] Drag and drop files between folders
- [ ] Disable/Enable Jolli Memory, verify all panels recover correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Plans

- JOLLI-1309: Local Knowledge Base — IntelliJ Implementation Plan

### 1. Plugin builds and installs successfully after storage refactor

**Preconditions:** Have IntelliJ IDEA open with the JolliMemory plugin project loaded. Have access to the plugin installation settings.

**Steps:**
1. Open a terminal in the plugin project directory and run the build command to compile the plugin.
2. Wait for the build to complete and verify the output shows "BUILD SUCCESSFUL" with no errors or failures.
3. Locate the built plugin distribution file (should be a .zip file in the build output folder).
4. In IntelliJ, open Settings → Plugins.
5. Click the gear icon and select "Install Plugin from Disk".
6. Navigate to and select the plugin .zip file, then click "Install".
7. When prompted, restart IntelliJ to activate the plugin.
8. After restart, open Settings → Plugins and search for "JolliMemory" to confirm it is installed and enabled.

**Expected Results:**
- The build completes with zero failures and displays "BUILD SUCCESSFUL".
- The plugin .zip file is created in the expected build output location.
- The plugin installs without errors.
- After restart, JolliMemory appears in the Plugins list as an installed and enabled plugin.
- The JolliMemory tool window appears in the IDE (usually on the right side or bottom).

### 2. Configure knowledge base folder location in settings

**Preconditions:** Have IntelliJ open with the JolliMemory plugin installed and enabled. Have a folder ready where you want to store your knowledge base (e.g., ~/Documents/jollimemory/).

**Steps:**
1. Open IntelliJ Settings (Preferences on macOS).
2. Search for "JolliMemory" or navigate to Tools → JolliMemory Settings.
3. Locate the "Knowledge Base" section.
4. Click the "Browse" button next to the folder path field.
5. Navigate to your desired knowledge base folder and select it.
6. Click "OK" to confirm the folder selection.
7. Click "Apply" and then "OK" to save the settings.
8. Verify the folder path now appears in the Knowledge Base field.

**Expected Results:**
- The Settings dialog opens and displays a "Knowledge Base" section.
- A folder path field and Browse button are visible.
- After selecting a folder, the path is populated in the field.
- The settings are saved without errors.
- The configured folder path persists after closing and reopening Settings.

### 3. View knowledge base files in the Explorer panel

**Preconditions:** Have a knowledge base folder configured in settings. Have at least one commit summary or file already stored in the knowledge base (or create one by making a commit in the repository).

**Steps:**
1. Open the JolliMemory tool window in IntelliJ.
2. Click on the "Memories" tab (or the first tab if tabs are visible).
3. Expand the branch folder tree to view the contents.
4. Scroll through the file list to see all stored summaries and files.
5. Look for files with badges (e.g., [C] for commit, [P] for plan, [N] for note).
6. Click on one of the commit summary files to open it.

**Expected Results:**
- The Memories tab displays a tree view of the knowledge base folder structure.
- Branch folders are listed and can be expanded to show their contents.
- Files are displayed with appropriate badges indicating their type (commit, plan, note).
- Clicking a commit summary file opens it in a readable format (either a webview or text editor).
- The tree updates automatically if files are added or removed externally.

### 4. Import a user file into the knowledge base

**Preconditions:** Have the JolliMemory tool window open with the Memories tab visible. Have a file on your computer (e.g., a .txt, .md, or .pdf file) that you want to add to the knowledge base.

**Steps:**
1. In the Memories tab, right-click on a branch folder or the root of the knowledge base.
2. Select "Import File(s)..." from the context menu.
3. A file chooser dialog opens. Navigate to and select the file you want to import.
4. Click "Open" or "Import" to confirm.
5. Wait a moment for the file to be copied into the knowledge base folder.
6. Verify the imported file now appears in the tree view under the selected folder.

**Expected Results:**
- A context menu appears with an "Import File(s)..." option.
- The file chooser dialog opens and allows you to browse and select files.
- The selected file is copied into the knowledge base folder.
- The imported file appears in the Memories tree view without a type badge (indicating it is a user file, not AI-generated).
- The file can be opened by clicking on it in the tree.

### 5. Commit code and verify summary appears in knowledge base folder

**Preconditions:** Have the knowledge base folder configured. Have a Git repository open in IntelliJ with the JolliMemory plugin active. Have valid API credentials configured (Anthropic or Jolli OAuth).

**Steps:**
1. Make a code change in the repository (e.g., edit a file and save it).
2. Stage and commit the change using IntelliJ's Git interface or terminal.
3. Wait 5–10 seconds for the post-commit hook to run and generate a summary.
4. Open the JolliMemory tool window and click the "Memories" tab.
5. Expand the current branch folder in the tree view.
6. Look for a new markdown file with a [C] badge (indicating a commit summary).
7. Click on the file to open and read the generated summary.

**Expected Results:**
- After committing, a new markdown file appears in the knowledge base folder under the current branch.
- The file has a [C] badge indicating it is an AI-generated commit summary.
- The file name includes the commit hash and a slug (e.g., "abc12345-add-oauth-flow.md").
- Opening the file displays the summary with a YAML header (commit hash, branch, author, date) followed by the summary text.
- The summary is readable and relevant to the code changes committed.

### 6. Privacy policy clarifies data handling for OAuth and API key authentication

**Preconditions:** Have access to the JolliMemory privacy policy (typically in a help menu, settings, or on the website).

**Steps:**
1. Open the JolliMemory privacy policy document or page.
2. Search for or scroll to the section about authentication and data handling.
3. Read the paragraph that explains the three authentication paths.
4. Verify the text clearly distinguishes between: (a) using your own API key, (b) using Jolli OAuth, and (c) pushing to Jolli Space.
5. Check that the policy explicitly states that code diffs and AI conversations are not stored or cached.

**Expected Results:**
- The privacy policy contains a clear section explaining authentication and data handling.
- The policy distinguishes three scenarios: own API key (data sent directly to provider), Jolli OAuth (data transmitted through Jolli servers but not stored), and Jolli Space (summaries stored per normal practices).
- The policy explicitly states that AI conversations and code diffs are never retained after summary generation.
- The language is clear and non-technical, suitable for end users.

### 7. Switch storage mode from orphan branch to folder-based storage

**Preconditions:** Have the plugin installed and a knowledge base folder configured. Have existing summaries stored in the orphan branch (from previous commits).

**Steps:**
1. Open IntelliJ Settings and navigate to JolliMemory settings.
2. Locate the "Storage Mode" or "Backend" option (if visible in settings).
3. Change the setting from "Orphan Branch" to "Folder-Based" or similar.
4. Click "Apply" and "OK" to save.
5. Open the JolliMemory tool window and click the "Memories" tab.
6. Verify that the knowledge base folder still displays all previously stored summaries.
7. Make a new commit and verify the new summary appears in the folder (not in the orphan branch).

**Expected Results:**
- The Settings dialog includes a storage mode option.
- Changing the mode from orphan branch to folder-based completes without errors.
- All previously stored summaries remain accessible in the knowledge base folder.
- New summaries are written to the folder-based storage, not the orphan branch.
- The plugin continues to function normally with the new storage mode.

### 8. Three-tab layout displays Memories, Branch, and Status information

**Preconditions:** Have the JolliMemory tool window open in IntelliJ.

**Steps:**
1. Look at the JolliMemory tool window and identify the tabs at the top.
2. Click on the first tab labeled "Memories" and verify it shows the knowledge base folder tree.
3. Click on the second tab labeled "Branch" (or similar) and verify it displays branch-related information (plans, changes, commits).
4. Click on the third tab labeled "Status" and verify it shows plugin status and sync information.
5. Switch between tabs a few times to confirm they are responsive and display the correct content.
6. Make a commit and verify the Status tab updates to reflect the new summary.

**Expected Results:**
- The JolliMemory tool window displays three tabs at the top.
- The first tab (Memories) shows the knowledge base folder tree with branch folders and files.
- The second tab (Branch) displays plans, changes, and commits for the current branch.
- The third tab (Status) shows plugin status, sync status, and migration progress (if applicable).
- Tabs are clickable and switch content smoothly.
- The Status tab updates in real-time when new summaries are generated.

### 9. Verify all unit tests pass after storage abstraction refactor

**Preconditions:** Have the plugin project open in IntelliJ or a terminal with access to the Gradle build tool.

**Steps:**
1. Open a terminal in the plugin project directory.
2. Run the unit test command (e.g., `gradle test` or `./gradlew test`).
3. Wait for all tests to complete.
4. Review the test output to verify all tests passed.
5. Check the summary line at the end (should show "X tests passed, 0 failed").
6. If any tests failed, review the error messages and verify they are unrelated to the storage refactor.

**Expected Results:**
- All unit tests run without errors.
- The test output shows "BUILD SUCCESSFUL" or "All tests passed".
- The summary indicates 22 tests passed (or the expected number for the SummaryStore test suite).
- Zero test failures are reported.
- The test results confirm that the storage abstraction refactor did not break existing functionality.

### 10. Verify plugin distribution file is created and ready for installation

**Preconditions:** Have completed a full Gradle build of the plugin project.

**Steps:**
1. Open a file browser and navigate to the plugin project's build output directory (typically `intellij/build/distributions/`).
2. Look for a .zip file with the plugin name and version (e.g., `jollimemory-intellij-0.97.9.1.zip`).
3. Verify the file size is reasonable (should be a few megabytes, e.g., 3–5 MB).
4. Check the file's modification timestamp to confirm it was created during the recent build.
5. Optionally, right-click the file and view its properties to confirm it is a valid ZIP archive.

**Expected Results:**
- The plugin distribution .zip file exists in the build output directory.
- The file name includes the plugin name and version number.
- The file size is between 2–5 MB (indicating it contains compiled code and resources).
- The file's timestamp matches the recent build time.
- The file is a valid ZIP archive and can be opened (if tested).

---

#### 01 · Make Branch tab the default view when tool window opens

**⚡ Why This Change**

When opening the Jolli Memory tool window in IntelliJ, the Knowledge Base tab was shown by default instead of the Branch tab. The user wanted the Branch tab to appear first.

- **Ordering over explicit selection**: IntelliJ's content manager automatically selects the first added tab, so simply reordering the two `addContent` calls was sufficient — no extra API calls or state management needed, keeping the change minimal and safe.

**✅ What Was Implemented**

Swapped the order in which the two content tabs are registered with the tool window's content manager in `JolliMemoryToolWindowFactory.kt`. The Branch content is now added before the Knowledge Base content, causing IntelliJ to select it by default on open.

#### 02 · Fix CommitsPanel showing "disabled" after Enable on empty branches

**⚡ Why This Change**

After clicking Enable in the JolliMemory status panel, the Plans and Changes panels would correctly restore their content, but the CommitsPanel continued to display "Jolli Memory is disabled." even though all hooks were successfully installed and `areAllHooksInstalled()` returned true.

The `emptyLabel` text reset in `updateCommitList()` was the true minimal fix — the label was being reused across enabled/disabled states without resetting its content. The `@Synchronized` on `refreshStatus()` and the `installProtectionUntil` guard were added as defense-in-depth against the `GIT_REPO_CHANGE` flapping race, which was confirmed by desktop debug logs showing the `enabled` value oscillating `true → false → true` within 2 seconds of install. A `forceRefresh()` path was added to bypass `refreshVersion` because the version-increment mechanism, while correct in theory, was vulnerable to the concurrent status flap causing the final pooled-thread result to show `disabled`. Direct `JOptionPane` dialogs and desktop file logging were used during diagnosis because IntelliJ's notification system was unreachable from the pooled thread context and the normal log files were not being written by the version under test.

**✅ What Was Implemented**

Root cause was identified through systematic debug logging as a two-part problem: (1) `refreshStatus()` was called concurrently from multiple pooled threads after `GIT_REPO_CHANGE` events fired when `.git/hooks/` was modified during install, causing the `enabled` status to flap `true → false → true`; (2) even when the final `refreshFromGit()` call read `enabled=true` and called `updateCommitList()` with an empty commit list, the `emptyLabel` text still read "Jolli Memory is disabled." because `showDisabled()` had mutated it and `updateCommitList()` never reset it. Fixes applied: added `@Synchronized` to `refreshStatus()` to eliminate concurrent-write races; added a 3-second `installProtectionUntil` guard in `JolliMemoryService` to suppress `enabled→disabled` flaps immediately after install; added a `forceRefresh()` method on `CommitsPanel` that bypasses the `refreshVersion` stale-discard mechanism; reset `emptyLabel.text` to the "Start coding" message inside `updateCommitList()` before displaying it; added a second `service.refreshStatus()` call in `StatusPanel.onToggle` after `install()` returns; added install-time verification logging to `HookInstaller` and debug logging to `refreshFromGit`.

#### 03 · Fix KB path change handling to auto-migrate data

**⚡ Why This Change**

When users change the knowledge base folder path in Settings and apply changes, the panel did not switch to the new path and data was not migrated from the old location.

Always re-resolving the KB root on refresh is simpler than adding a config change listener, and ensures the panel stays in sync with Settings changes without requiring a restart. Auto-migration on Settings Apply is transparent to users and prevents data loss when switching paths. The migration is wrapped in a try-catch so Settings Apply succeeds even if migration fails (folder write errors don't block the primary config save).

**✅ What Was Implemented**

Modified `KBExplorerPanel.refresh()` to always call `resolveKBRoot()` instead of caching it, ensuring the panel re-reads the config on every refresh. Updated `SettingsDialog` Apply button to auto-trigger migration: after writing the new config, it checks if the orphan branch exists and runs `MigrationEngine.runMigration()` to copy historical data to the new KB folder path.

#### 04 · Improve drag-and-drop support for KB folder file operations

**⚡ Why This Change**

Drag-and-drop capability was missing or incomplete; internal tree drags and external Finder drops were not working reliably.

`DropMode.ON_OR_INSERT` is more forgiving than `ON` and matches user expectations for tree drop targets. The native `DropTarget` is necessary because IntelliJ's `Tree` class may override or suppress default Swing DnD behavior, so a dual approach (TransferHandler + DropTarget) ensures compatibility with both internal and external drag sources. Refresh after move ensures the manifest and UI stay in sync with filesystem changes.

**✅ What Was Implemented**

Changed `DropMode` from `ON` to `ON_OR_INSERT` for more permissive drop zones. Added `exportDone()` callback to refresh the tree after a move operation completes. Implemented a native `DropTarget` adapter in addition to the `TransferHandler` to catch system-level drag-and-drop events from Finder that the TransferHandler might miss. Both internal (tree-to-tree) and external (Finder-to-folder) drags now trigger `backgroundRefresh()` to update the UI.

#### 05 · Document local knowledge base technical journey and architecture

**⚡ Why This Change**

User requested a detailed, non-abstract explanation of the entire technical evolution from Step 1.1 to the current dual-write architecture, including the relationship between file formats and orphan branch design.

Structured the narrative chronologically by development phase rather than by component, making it easier to understand how each piece solved a specific problem. Used concrete file paths and JSON examples to ground abstract concepts. Included both Chinese and English versions to serve the bilingual development team. Emphasized the relationship between orphan branch (primary immutable JSON), `.jolli/` (hidden backup JSON), and visible markdown files, clarifying why three copies exist and how they enable data recovery.

**✅ What Was Implemented**

Created two comprehensive technical journey documents (English and Chinese) covering: the original orphan-branch-only architecture, the StorageProvider interface extraction (Step 1.1), FolderStorage and MetadataManager implementation (Steps 1.2–1.3), KB path configuration (Step 1.4), DualWriteStorage dual-write phase, the FolderStorage refactor from "dumb write" to "smart write" (JSON to hidden `.jolli/` + markdown to visible folders), data migration via MigrationEngine, the Memories panel UI, the three-copy data model, data recovery chains, and future evolution toward folder-only mode. Included concrete examples of file structures, code snippets, and a bug-fix table.

#### 06 · Fix custom KB path to append repo name as subdirectory

**⚡ Why This Change**

User set custom knowledge base path to ~/Documents/test and discovered all repositories' data was being written to the same folder instead of separate subfolders. Additionally, the VSCode plugin was also writing to the same incorrect location, causing data from multiple repos to mix together.

Custom paths are now treated identically to the default path behavior (~/Documents/jolli/ becomes ~/Documents/jolli/{repoName}/). This ensures consistency across all storage configurations and prevents data mixing when multiple repositories use the same custom parent directory. The fix applies globally to both IntelliJ plugin and VSCode plugin since they share the same PostCommitHook JAR and configuration.

**✅ What Was Implemented**

Modified KBPathResolver.resolve() to treat customPath as a parent directory rather than the final KB root. The repoName is now always appended to the custom path, ensuring each repository gets its own subfolder (e.g., ~/Documents/test/repo1/, ~/Documents/test/repo2/). Updated SettingsDialog tooltip and placeholder text to clarify that the custom path is a parent folder. Updated KBPathResolverTest to verify the new behavior. Bumped version to 0.97.9.52.

#### 07 · Fix drag-and-drop functionality in KB explorer tree

**⚡ Why This Change**

Drag-and-drop appeared to initiate but drops were not being processed, preventing users from moving files and folders within the Knowledge Base.

Native DropTarget takes precedence over TransferHandler on the same component, causing the TransferHandler's drop logic to be ignored. Unified approach eliminates this conflict. Used DragGestureListener for explicit drag initiation control rather than relying on implicit dragEnabled flag. Added selectionRows null guard because macOS requires valid selection state to generate drag images.

**✅ What Was Implemented**

Replaced conflicting TransferHandler + DropTarget dual implementation with unified DragSource + DropTarget approach. Added null check for selectionRows before startDrag to prevent macOS NPE in CDragSourceContextPeer. Implemented proper drop target detection using getClosestRowForLocation with visual feedback (row highlighting during drag-over). Supports both internal tree moves (with metadata sync via renameBranchFolder/updatePath) and external Finder drops.

#### 08 · Add visual selection highlighting and single-selection mode to tree

**⚡ Why This Change**

Users had no indication of which file or folder was currently selected, making it difficult to understand the tree state and target of operations.

Single-selection mode matches standard file explorer UX and prevents confusion from multi-select. Visual feedback during drag-over helps users understand where drops will land. Selection preservation across refresh maintains user context when tree rebuilds.

**✅ What Was Implemented**

Set TreeSelectionModel to SINGLE_TREE_SELECTION mode. Added visual highlighting of drop target nodes during drag-over via setSelectionRow. Preserved selection state across tree refresh by storing and restoring selectionPath.

#### 09 · Enable folder and file creation in KB root via right-click empty area

**⚡ Why This Change**

Right-clicking empty space in the tree selected the nearest node instead of clearing selection, making it impossible to create items in the root directory.

getRowForLocation provides precise hit detection; getClosestRowForLocation was too forgiving. Using isPopupTrigger (platform-aware) instead of manual button checks handles both mousePressed and mouseReleased correctly. Null data parameter signals "root context" to context menu logic. Showing empty folders improves visibility of newly created directories.

**✅ What Was Implemented**

Replaced getClosestRowForLocation (which always returns nearest row) with getRowForLocation (returns -1 if click is not exactly on a row). Refactored right-click handling into showPopup method using isPopupTrigger flag. When row is -1, clear selection and pass null data to context menu, which defaults targetDir to root. Removed childCount > 0 filter that was hiding empty folders from tree display.

#### 10 · Open new markdown files with write access, bypassing non-project protection

**⚡ Why This Change**

Newly created markdown files opened as read-only due to IntelliJ's non-project file protection, requiring manual intervention to edit.

OpenFileDescriptor + openTextEditor bypasses the non-project file read-only guard that openFile enforces. Calling isWritable triggers VFS state refresh before editor opens. This approach is used internally by IntelliJ for similar scenarios.

**✅ What Was Implemented**

Replaced FileEditorManager.openFile with OpenFileDescriptor + openTextEditor pattern. Added explicit vFile.isWritable call to trigger VFS refresh. Applied same pattern to openFile helper method for consistency.

#### 11 · Bump plugin version to 0.97.9.51

**⚡ Why This Change**

Multiple bug fixes and improvements (DnD, KB path handling, migration) required a version increment to reflect the changes.

Incremented by 2 minor versions to account for the two intermediate fixes (0.97.9.50 for DnD improvements, 0.97.9.51 for KB path + migration).

**✅ What Was Implemented**

Updated `build.gradle.kts` version from `0.97.9.49` to `0.97.9.51`.

#### 12 · Improve Finder integration with reveal-and-select behavior

**⚡ Why This Change**

"Open in Finder" opened the folder but did not select or highlight the specific file/folder, requiring users to manually locate it.

The -R flag to macOS open command selects the file in Finder, matching standard file explorer behavior. This is more discoverable and user-friendly than opening a folder. Explicit path targeting (not parent) respects user intent.

**✅ What Was Implemented**

Renamed menu item to "Reveal in Finder". Replaced Desktop.getDesktop().open() with Runtime.exec(["open", "-R", path]) which uses macOS native reveal-and-select behavior. Targets the specific file/folder (data.path) rather than always opening the parent directory.

#### 13 · Improve Move To dialog to open at KB root directory

**⚡ Why This Change**

File chooser dialog for "Move to..." did not default to or navigate to the Knowledge Base root, requiring manual navigation.

refreshAndFindFileByNioFile performs VFS refresh before lookup, ensuring the file is found even if VFS cache is stale. This is more reliable than findFileByNioFile alone, especially for recently created or externally modified directories.

**✅ What Was Implemented**

Changed LocalFileSystem.findFileByNioFile(root) to refreshAndFindFileByNioFile(root) when obtaining the initial folder for FileChooser. This ensures the KB root is located and passed as the starting directory.

#### 14 · Rename Memories tab to Knowledge Base

**⚡ Why This Change**

User requested clarification on terminology: should the tab be labeled "Memories" or "Knowledge Base" (singular vs. plural).

Singular "Knowledge Base" is standard terminology (like "Database", not "Databases"). Matches industry convention used by IntelliJ, Confluence, Notion. Singular form better represents a single unified KB rather than multiple memory stores.

**✅ What Was Implemented**

Changed tab title from "📚 Memories" to "📚 Knowledge Base" (singular). Updated only the Content tab label; other internal references to MEMORIES panel left unchanged.

#### 15 · Display current Git branch name in Branch tab with auto-update

**⚡ Why This Change**

User wanted the Branch tab to show the active feature branch name and update automatically when switching branches.

Used GitRepositoryChangeListener instead of polling to avoid performance overhead. Emoji prefix (🌿) provides visual distinction in the tab bar since IntelliJ Content tabs don't render SVG icons. SwingUtilities.invokeLater ensures UI updates happen on the EDT.

**✅ What Was Implemented**

Modified JolliMemoryToolWindowFactory to read current branch via GitOps, display it in the Branch content tab title (e.g., "🌿 feature/sign-up"), and subscribe to GitRepository.GIT_REPO_CHANGE events to auto-update the tab title when the branch changes.

#### 16 · Add readable commit message titles to Memories explorer tree

**⚡ Why This Change**

Memories panel displayed raw markdown filenames (e.g., "add-login-feature-abc12345.md") instead of human-readable commit messages like "Add login feature".

Storing title in manifest allows decoupling display name from filesystem naming. This preserves proper case, removes hash/slug artifacts, and makes the tree immediately readable. Fallback to filename ensures backward compatibility with user-imported files that have no manifest entry.

**✅ What Was Implemented**

Added `title` field to ManifestEntry in KBTypes.kt. Updated FolderStorage.generateMarkdown() to store the commit message as title in the manifest. Modified KBExplorerPanel to build a titleMap from manifest entries and display title instead of filename in the tree. Updated tree cell renderer to show displayName (title) when available, falling back to filename for user files.

#### 17 · Backfill missing titles in existing manifest entries during migration

**⚡ Why This Change**

Existing manifest entries created before the title field was added lacked readable titles, so the Memories tree still showed filenames for old commits.

Backfilling during migration is non-destructive and idempotent—it only updates the title field if missing. This avoids requiring users to manually re-migrate or delete old data. The approach leverages the existing orphan storage to read summary JSON.

**✅ What Was Implemented**

Added backfillTitle() method to MigrationEngine that detects existing manifest entries without a title field and reads the CommitSummary JSON to extract the commitMessage. Modified the migration idempotency check to call backfillTitle() for entries that already exist but lack a title.

#### 18 · Open commit markdown files in formatted JCEF webview instead of plain text editor

**⚡ Why This Change**

Double-clicking a commit markdown file in the Memories tree opened raw markdown text instead of the formatted HTML view available in the CommitsPanel eye icon.

Reusing SummaryVirtualFile ensures consistent formatting and behavior across the UI. Reading from the hidden JSON (not the markdown file) guarantees access to structured data needed for rendering. Background thread execution prevents UI blocking during file I/O.

**✅ What Was Implemented**

Added openCommitSummary() method to KBExplorerPanel that reads the hidden summary JSON from .jolli/summaries/{hash}.json, parses it into a CommitSummary object, and opens it via SummaryVirtualFile (the same JCEF webview used by CommitsPanel). Modified the tree double-click handler to detect files with "C" badge and route them to openCommitSummary() instead of openFile().

#### 19 · Reorganize tool window content tabs: Memories first, Branch second

**⚡ Why This Change**

User wanted to reorder the tab layout to prioritize the Memories explorer panel.

Memories is the primary feature (KB exploration), so it appears first in the tab bar. Emoji icons provide visual distinction and are the only reliable way to add icons to Content tabs in IntelliJ. The 📚 and 🌿 symbols are intuitive and match the semantic meaning of each tab.

**✅ What Was Implemented**

Swapped the order of content registration in JolliMemoryToolWindowFactory: Memories tab (with 📚 emoji) is now added first, Branch tab (with 🌿 emoji) second. Both tabs use emoji prefixes since IntelliJ Content tabs don't render SVG icons.

#### 20 · KB explorer right-click context menu for file and folder operations

**⚡ Why This Change**

Phase 4.2 requires users to create folders, add markdown files, move files, and delete items within the Knowledge Base through the UI. Previously, the KB explorer was read-only.

Context menu appears on right-click and targets the selected node or parent directory if a file is selected. Operations run on background threads to avoid blocking the UI. Rename and Move operations automatically update manifest paths via MetadataManager.updatePath(). Delete operations remove entries from manifest via MetadataManager.removeFromManifest() or MetadataManager.removeBranchFolder() for directories. This keeps metadata in sync with filesystem state without requiring manual reconciliation.

**✅ What Was Implemented**

Added context menu to KBExplorerPanel with operations: New Folder, New Markdown File, Import File(s), Rename, Move to, Delete, and Open in Finder. Each operation updates the filesystem and triggers metadata synchronization via MetadataManager. New files open automatically in the editor after creation.

#### 21 · Metadata synchronization for branch folder rename and removal operations

**⚡ Why This Change**

When users rename or delete branch folders through the UI, the manifest and branches.json metadata must be updated to reflect the changes, otherwise stale references accumulate.

Metadata updates are atomic (single write per operation) to avoid partial state. Branch folder operations affect multiple manifest entries, so they are handled separately from single-file operations (updatePath, removeFromManifest). This separation keeps the API clear and testable.

**✅ What Was Implemented**

Added two new methods to MetadataManager: renameBranchFolder(oldFolder, newFolder) updates branches.json mappings and all manifest entries whose paths start with the old folder name; removeBranchFolder(folder) removes the folder from branches.json and deletes all manifest entries under that folder. Both methods return a count of affected entries and log the operation.

#### 22 · External change detection and automatic manifest reconciliation

**⚡ Why This Change**

Users may modify KB files outside the IDE (via Finder, terminal, etc.). The manifest can become stale with deleted or moved files. The system needs to detect and repair these inconsistencies automatically.

Fingerprint-based move detection allows recovery of files moved outside the UI without user intervention. Reconciliation runs before tree building so the UI always displays current state. The method is idempotent and logs all changes for debugging. Fingerprints are computed using FolderStorage.sha256() on file content.

**✅ What Was Implemented**

Added MetadataManager.reconcile(kbRoot) method that scans the filesystem, builds a fingerprint map of all .md files, and compares against manifest entries. Deleted files are removed from manifest. Moved files are detected by matching fingerprints and their paths are updated. Reconcile is called automatically on load() and refresh() in KBExplorerPanel before building the tree.

#### 23 · Comprehensive test coverage for metadata operations and reconciliation

**⚡ Why This Change**

New metadata operations (rename, remove, reconcile) require test coverage to ensure correctness and prevent regressions.

Tests use the existing tempDir fixture and makeEntry() helper. Reconciliation tests create actual filesystem state (directories and files) to verify the walk and fingerprint matching logic. Tests verify both the filesystem state and metadata state after each operation.

**✅ What Was Implemented**

Added 7 new test cases to MetadataManagerTest: renameBranchFolder updates manifest and branches.json, does not affect other branches; removeBranchFolder removes entries and mappings; reconcile detects deleted files and removes them, detects moved files by fingerprint and updates paths, and returns 0 when no changes are needed.

#### 24 · Enable drag and drop for files and folders in KB explorer tree

**⚡ Why This Change**

User requested drag and drop support for the knowledge base explorer, enabling both internal file/folder movement within the tree and external file imports from the file system.

Used Swing's TransferHandler pattern rather than custom mouse listeners to leverage native drag-and-drop semantics and OS integration. Implemented dual DataFlavor support (javaFileListFlavor for external, stringFlavor for internal) to distinguish drag sources and apply appropriate file operations. Initially restricted to files only, but expanded to folders after user feedback that folder moves should be supported—this required adding recursive manifest update logic to handle nested entries.

**✅ What Was Implemented**

Implemented a custom TransferHandler on the JTree with support for: (1) internal drag-and-drop of files and folders between tree nodes with automatic manifest path updates; (2) external drag-and-drop from Finder with automatic detection of KB-internal vs external sources (move vs copy); (3) folder move handling that recursively updates all manifest entries under the moved branch via renameBranchFolder(). Drop target is any folder node; drop mode set to DropMode.ON.

#### 25 · Fix Memories panel auto-refresh on commit and file changes

**⚡ Why This Change**

KBExplorerPanel only loaded once at startup and never refreshed when new commits were dual-written to the KB folder or when files changed externally. Users had to disable/enable the plugin to see new memories.

Status listener provides immediate feedback for the common case (new commits via IDE), while periodic polling handles edge cases (external file edits, dual-write delays). 3-second interval balances responsiveness vs. performance. Pooled thread execution prevents UI blocking on large KB folders.

**✅ What Was Implemented**

Added dual refresh mechanism to KBExplorerPanel: (1) Service status listener that triggers immediate refresh when PostCommitHook completes, (2) 3-second periodic timer as fallback to catch external file system changes. Both paths call `refresh()` on pooled thread. Cleanup in `dispose()` stops timer and removes listener.

#### 26 · Restore dual-write mode after Settings dialog overwrites config

**⚡ Why This Change**

Settings dialog was overwriting config.json with only fields it manages (apiKey, model, knowledgeBaseSort), losing unmanaged fields like `storageMode: "dual-write"`. This caused all subsequent commits to use orphan-only mode, breaking KB folder writes.

Merge-on-save approach is safer than adding UI fields for every config option. Preserves user intent for fields outside the dialog's scope. Allows future config fields to persist without dialog changes.

**✅ What Was Implemented**

Modified SettingsDialog to merge new config with existing config before saving. Before `saveConfigToDir()`, load existing config and preserve `storageMode`, `logLevel`, and `logLevelOverrides` fields. Also restored `storageMode: "dual-write"` to the user's config.json.

#### 27 · Add VCS_CONFIGURATION_CHANGED listener for branch switch detection

**⚡ Why This Change**

Branch tab title and Commits/Changes panels only refreshed on `GIT_REPO_CHANGE` events, which don't fire reliably when branches are created/switched via terminal (`git checkout -b`, `git switch`). IntelliJ's VFS doesn't monitor `.git/HEAD` in real-time.

Layered detection approach ensures reliability across all branch creation methods (IDE UI, terminal, external tools). 2-second poll interval is fast enough for user perception while avoiding excessive git calls. VCS listener catches most cases; polling is safety net.

**✅ What Was Implemented**

Added `VCS_CONFIGURATION_CHANGED` listener to CommitsPanel and ChangesPanel as secondary detection path. Added 2-second periodic polling in JolliMemoryToolWindowFactory to detect branch changes via `git rev-parse --abbrev-ref HEAD`. Branch title now updates via 4 paths: GIT_REPO_CHANGE, VCS_CONFIGURATION_CHANGED, periodic poll, and service status listener.

#### 28 · Rename tool window from "JOLLI MEMORY" to "JOLLI"

**⚡ Why This Change**

User requested simplification of the tool window identifier in the IDE sidebar.

Single-word identifier is cleaner and more concise while remaining recognizable. No functional impact.

**✅ What Was Implemented**

Changed tool window id in plugin.xml from "JOLLI MEMORY" to "JOLLI".

#### 29 · Change markdown filename format from hash-slug to slug-hash

**⚡ Why This Change**

User questioned whether prefixing filenames with commit hash was appropriate, preferring a more readable format.

Slug-first format is more human-readable in file browsers while still maintaining uniqueness via the hash suffix. Hash at the end is less visually prominent but still prevents collisions.

**✅ What Was Implemented**

Updated FolderStorage.generateMarkdown() to generate filenames as `{slug}-{hash8}.md` instead of `{hash8}-{slug}.md`. Updated documentation comments to reflect the new format.

#### 30 · Extract StorageProvider abstraction layer from SummaryStore

**⚡ Why This Change**

SummaryStore.kt directly calls GitOps for all orphan branch I/O, tightly coupling storage implementation to git plumbing. To support multiple storage backends (orphan branch, folder-based), an abstraction layer is needed.

Chose interface-based abstraction over conditional branching to enable clean swapping between storage backends. Extracted git plumbing into OrphanBranchStorage rather than keeping it in SummaryStore to isolate implementation details and make the interface contract explicit. Kept GitOps in SummaryStore for non-storage git operations (cat-file for tree hash extraction) because those are read-only queries against the actual repo, not storage operations. Secondary constructor ensures zero breaking changes to existing code during the refactor phase.

**✅ What Was Implemented**

Created StorageProvider interface with readFile, writeFiles, listFiles, exists, ensure methods. Extracted all git plumbing from SummaryStore into OrphanBranchStorage (hash-object, mktree, commit-tree, update-tree logic). Refactored SummaryStore to inject StorageProvider via constructor, replacing 14 direct gitOps calls. Added backward-compatible secondary constructor that auto-creates OrphanBranchStorage, keeping all 6 existing call sites unchanged. All 22 unit tests pass without modification.

#### 31 · Create comprehensive IntelliJ implementation plan for JOLLI-1309 local knowledge base

**⚡ Why This Change**

VS Code plan was written in TypeScript; IntelliJ plugin is a complete Kotlin rewrite with different architecture (JAR subprocess, NIO file locking, Swing UI). Need IntelliJ-specific plan mapping all 15 sub-issues to exact Kotlin files, classes, and methods.

Chose Strangler Fig migration pattern (dual-write shadow, then read switchover) over big-bang cutover to eliminate data loss risk. Kept orphan branch as primary during dual-write phase to validate FolderStorage correctness before switching reads. Identified 3 independent parallel tracks (KB settings UI, UI redesign, Space backend API) to maximize team throughput. Documented JAR subprocess as critical concern because it runs as separate Java process and must bundle new storage classes. Used NIO FileChannel.tryLock() instead of file-existence checks for `.jolli/lock` because it's more robust across platforms.

**✅ What Was Implemented**

Wrote two parallel plans (English and Chinese) targeting IntelliJ Kotlin codebase. Mapped 6 phases across 19 files: Phase 1 (StorageProvider + FolderStorage + MetadataManager + KB settings), Phase 2 (DualWriteStorage + MigrationEngine), Phase 3 (PostCommitHook rewire), Phase 4 (ExplorerPanel + import UI + 3-tab redesign), Phase 5 (SyncEngine), Phase 6 (Space API integration). Included execution timeline showing parallelism, IntelliJ-specific considerations (JAR subprocess bundling, NIO WatchService, FileChannel locking, Kotlin coroutines), risk mitigation table, and verification checklist.

#### 32 · Implement KB Explorer UI panel for browsing local knowledge base files

**⚡ Why This Change**

Users needed a way to visually browse and open markdown files stored in the local Knowledge Base folder within IntelliJ, separate from the existing orphan branch UI.

Added as a new independent tool window rather than modifying existing UI to avoid regression risk. Used background thread initialization to prevent blocking EDT during KB root resolution and tree building. Implemented lazy loading with "Loading..." message that auto-refreshes until KB root is resolved, ensuring robustness when service initialization timing is uncertain. Filtered out hidden directories (.jolli, summaries, transcripts) and metadata files from display to show only user-visible content.

**✅ What Was Implemented**

Created KBExplorerPanel as an independent tool window content showing a tree view of KB folder structure organized by branch. Panel displays branch folders and markdown files with badges (C for commit, P for plan, N for note), supports double-click to open files in editor, and auto-loads on background thread after service initialization. Integrated into JolliMemoryToolWindowFactory as a second content tab ("Memories") alongside the existing "Branch" tab.

#### 33 · Implement dual-write storage with automatic migration from orphan branch to folder

**⚡ Why This Change**

Phase 2 of the KB implementation required safely transitioning from orphan branch storage to folder-based storage without data loss, using a dual-write pattern where new commits write to both locations while reads remain on the primary (orphan branch).

Dual-write approach chosen over direct cutover to minimize risk — orphan branch remains source of truth while folder storage is validated. Shadow write failures are non-fatal to preserve backward compatibility. Migration is idempotent (already-migrated entries skipped by checking manifest) to allow safe re-runs. Auto-migration runs silently on service init rather than requiring user action, with manual "Migrate to Knowledge Base" button in Settings as fallback. StorageFactory pattern allows mode switching via config without code changes.

**✅ What Was Implemented**

Implemented DualWriteStorage class that writes to both OrphanBranchStorage (primary) and FolderStorage (shadow), with shadow write failures logged as warnings but not blocking primary writes. Created StorageFactory to instantiate the correct provider based on storageMode config ("orphan", "dual-write", or "folder"). Implemented MigrationEngine to batch-migrate existing orphan branch data to folder storage, generating markdown files and tracking progress in .jolli/migration.json. Added auto-migration trigger in JolliMemoryService.initialize() that runs migration if orphan branch exists and migration status is not "completed".

#### 34 · Implement FolderStorage with markdown generation and metadata management

**⚡ Why This Change**

KB folder storage needed to write both hidden JSON data files and visible markdown files organized by branch, with metadata tracking for AI-generated files and branch name mapping.

Two-layer storage (hidden JSON + visible markdown) chosen to keep programmatic data separate from user-readable content. Markdown generation happens automatically during write rather than as separate step to ensure consistency. MetadataManager handles all .jolli/ operations atomically with file locking to prevent concurrent corruption. Branch name transcoding uses safe character replacement rather than hashing to keep folder names human-readable.

**✅ What Was Implemented**

Created FolderStorage class that routes data files to .jolli/ hidden directory and auto-generates visible markdown files with YAML frontmatter when summaries are written. Implemented MetadataManager to track manifest.json (AI-generated files), branches.json (git branch to folder name mapping), config.json (KB settings), and migration.json (migration progress). Added branch name transcoding to convert unsafe characters (/, :, *, ?, ~, ^) to safe folder names. Implemented atomic file writes via temp file + move to prevent corruption.

#### 35 · Implement KB path resolution with collision detection and repo identity tracking

**⚡ Why This Change**

KB folder location needed to be configurable with sensible defaults, handle naming collisions when multiple repos have same name, and track repo identity via remote URL to distinguish between different projects.

Remote URL used as primary identity key (normalized by removing .git suffix and trailing slash) because it uniquely identifies a repo across machines. Folder name matching by repoName as fallback for local repos without remotes. Collision detection happens at resolve time rather than requiring user intervention. initializeKBFolder() is idempotent and backfills repoName for legacy folders created before this field existed.

**✅ What Was Implemented**

Created KBPathResolver with resolve() method that returns ~/Documents/jolli/{repoName}/ by default, respects user custom path from config, and detects collisions by comparing remoteUrl in existing .jolli/config.json. Implemented initializeKBFolder() to write repo identity (remoteUrl, repoName) to .jolli/config.json. Added getRemoteUrl() to extract origin URL via git command. Collision handling appends -2, -3 suffixes when same repo name exists with different remote URL.

#### 36 · Add comprehensive test coverage for KB storage, migration, and metadata layers

**⚡ Why This Change**

New KB storage, migration, and metadata management code needed test coverage to ensure correctness and catch regressions.

Tests use @TempDir for isolated filesystem state. FolderStorage tests use real file I/O to catch atomic write and locking issues. MigrationEngine tests use InMemoryStorage adapter to avoid git plumbing. Comprehensive edge case coverage (empty orphan, root commits, corrupted JSON) to ensure resilience.

**✅ What Was Implemented**

Implemented 12 test classes with 100+ test cases covering: FolderStorage (read/write/delete, markdown generation, manifest tracking, file locking), MetadataManager (manifest ops, branch mapping, index rebuild, config), KBPathResolver (repo name extraction, path resolution, collision handling), MigrationEngine (basic migration, idempotency, progress tracking, branch organization), and DualWriteStorage (read/write contract, shadow failure resilience, factory modes).

#### 37 · Build and verify IntelliJ plugin distribution after storage refactor

**⚡ Why This Change**

After completing Phase 1.1 refactor (StorageProvider extraction), need to verify plugin builds cleanly and all tests pass before proceeding to Phase 2.

Ran full test suite before plugin build to catch any regressions early. Built distribution artifact to enable manual end-to-end testing in real IntelliJ environment (reads/writes to orphan branch, post-commit hook flow, UI panels). Version bump (0.97.9) reflects incremental progress within pre-1.0 development.

**✅ What Was Implemented**

Ran full test suite (22 unit tests in SummaryStoreTest) — all passed. Ran full Gradle build (15 tasks) — BUILD SUCCESSFUL with zero failures. Built plugin distribution: jollimemory-intellij-0.97.9.zip (3.1 MB) at intellij/build/distributions/. Plugin ready for manual IDE testing and installation via Settings → Plugins → Install Plugin from Disk.

#### 38 · Revise privacy policy paragraph for Jolli OAuth and API key usage

**⚡ Why This Change**

Privacy policy text was unclear about data handling when users authenticate via Jolli OAuth vs. providing their own third-party API key. Needed clarification that Jolli does not store, cache, or retain code diffs or AI conversations after summary generation.

Separated the three auth flows explicitly to avoid ambiguity. Clarified that Jolli acts as a proxy (not cache) when using Jolli credentials. Emphasized non-retention of transient data (diffs, conversations) to address privacy concerns while distinguishing from persistent summaries pushed to Space.

**✅ What Was Implemented**

Rewrote privacy paragraph to distinguish three authentication paths: (1) own API key — data sent directly to third-party provider, Jolli never sees it; (2) Jolli OAuth/API key — data transmitted through Jolli servers to provider, used only for summary generation, not stored or cached; (3) push to Jolli Space — summaries stored per normal customer data practices. Emphasized that AI conversations and code diffs are never retained.

#### 39 · Add Knowledge Base settings UI with folder path and sort order configuration

**⚡ Why This Change**

Users needed UI to configure custom KB folder path and sort order, with visual feedback showing default path and ability to manually trigger migration.

Folder path field left blank to use default rather than pre-filling, reducing visual clutter. Browse button uses FileChooserDescriptor for native folder picker. Migration button placed in Settings rather than main UI to keep it discoverable but not prominent. Button is disabled during migration and shows "Migrating..." text for feedback.

**✅ What Was Implemented**

Added Knowledge Base section to SettingsDialog with TextFieldWithBrowseButton for custom folder path (shows default ~/Documents/jolli/{project}/ as placeholder), ComboBox for sort order (date/name), and "Migrate to Knowledge Base" button. Button triggers MigrationEngine.runMigration() and shows result dialog. Settings are saved to global config and KB folder is auto-initialized on Apply.

#### 40 · Fix squash action to handle root commits without parent reference

**⚡ Why This Change**

Squash action failed when oldest commit in selection was a root commit (no parent), causing git rev-parse to fail and blocking the squash operation.

Explicit root commit detection via git rev-parse rather than catching exception, making intent clear. Using update-ref -d HEAD is the standard git way to reset to orphan state. Preserving existing soft-reset logic for non-root case minimizes risk of regression.

**✅ What Was Implemented**

Modified SquashAction to detect root commits by checking if git rev-parse {hash}^ returns null. For root commits, use git update-ref -d HEAD to delete HEAD and enter orphan state, then git commit creates new root. For non-root commits, use existing git reset --soft logic. Updated fork point resolution to handle null case.

#### 41 · Wire StorageFactory throughout codebase to enable storage mode switching

**⚡ Why This Change**

All SummaryStore instantiations needed to use StorageFactory to respect the configured storage mode (orphan, dual-write, or folder) rather than always using OrphanBranchStorage.

StorageFactory.create() called at each instantiation site rather than injecting a singleton, allowing per-project storage mode configuration. Backward-compatible 2-arg SummaryStore constructor kept for tests but not used in production code.

**✅ What Was Implemented**

Updated SummaryStore constructor to accept StorageProvider parameter. Replaced all 7 SummaryStore instantiations (in JolliMemoryService, PostCommitHook, PostRewriteHook, SummaryPanel, StatusPanel) to call StorageFactory.create(git, projectPath) instead of directly creating OrphanBranchStorage. Made OrphanBranchStorage open (non-final) to allow subclassing in tests.

#### 42 · Add storageMode and knowledgeBase config fields to JolliMemoryConfig

**⚡ Why This Change**

Global configuration needed to persist user's storage mode choice and KB folder path across IDE restarts.

Fields are nullable with sensible defaults (null = use default path, "date" sort, "orphan" mode) to maintain backward compatibility. SessionTracker merge logic ensures partial updates don't lose other config fields.

**✅ What Was Implemented**

Added three fields to JolliMemoryConfig: knowledgeBasePath (custom KB folder, null = default), knowledgeBaseSort (date/name), storageMode (orphan/dual-write/folder). Updated SessionTracker.saveConfig() to merge these fields when saving. Updated SettingsDialog to read/write these fields.

---

*Generated by JolliMemory · April 25, 2026 at 10:03 PM*


flyerli2025:
1. `[P1]` `intellij/src/main/kotlin/ai/jolli/jollimemory/core/MetadataManager.kt`
Transcoded branch folders can collide. We persist `transcodeBranchName(branchName)` without checking whether that folder name is already mapped to a different branch, so branches like `feature/foo` and `feature-foo` can be merged into the same KB directory.

2. `[P1]` `intellij/src/main/kotlin/ai/jolli/jollimemory/core/MigrationEngine.kt`
Plan/note migration never produces visible KB entries. `MigrationEngine` writes plain `FileWrite(path, content)` records, but `FolderStorage` can only generate visible markdown after it knows the branch. During migration `.jolli/index.json` is copied at the very end, so even hash-suffixed plans cannot resolve a branch yet, and note ids are not hash-based at all.

3. `[P1]` `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/KBExplorerPanel.kt`
Multi-repo explorer actions still use the current repo context. The panel renders nodes from `cachedRepos`, but open/action paths still derive `kbRoot` and `metadataManager` from the current project only, so actions on another repo can fall back to raw markdown or update the wrong manifest.

4. `[P2]` `intellij/src/main/kotlin/ai/jolli/jollimemory/core/KBPathResolver.kt`
Custom KB collisions fall back to the default parent. `resolve()` starts from `customPath`, but the collision path calls `findAvailablePath()` which always builds candidates under `KB_PARENT`, so a collision under a custom KB parent can silently create the fallback folder under `~/Documents/jolli/...` instead.

jordangreen-jolli (approved):

PR #60, opened 2026-05-01
Add space binding chooser flow with modernized push contract; Part of JM-SPACE: Simplify push titles to use only content without metadata prefix

foster-han:
## E2E Test (10)
<details>
<summary><strong>1. First-time push triggers space binding chooser</strong></summary>
<br>
<blockquote>

**Preconditions:** - You have the JolliMemory VS Code extension installed
- You are in a repository that has never been connected to a JolliMemory space
- You have a valid Jolli API key configured in the extension settings

**Steps:**
1. Open VS Code and navigate to the JolliMemory panel in the sidebar
2. Write or open a commit summary and click the **Push to Jolli** button
3. Wait for the binding chooser panel to appear (it should open automatically)
4. Select an existing memory space from the list shown in the "Existing space" tab
5. Click **Confirm** to bind your repository to that space

**Expected Results:**
- The binding chooser panel opens automatically after clicking Push
- Your push completes successfully after confirming the space selection
- A success message appears confirming the memory was published

</blockquote>
</details>
<details>
<summary><strong>2. Create a new memory space from the binding chooser</strong></summary>
<br>
<blockquote>

**Preconditions:** - You have the JolliMemory VS Code extension installed
- You are in a repository that has never been connected to a JolliMemory space

**Steps:**
1. Open the JolliMemory panel and click **Push to Jolli**
2. When the binding chooser opens, click the **New space** tab
3. Type a name for your new memory space in the name field
4. Click **Confirm** to create the space and bind your repository

**Expected Results:**
- The new space is created and your repository is bound to it
- The push completes successfully and a success confirmation appears
- No error messages are shown

</blockquote>
</details>
<details>
<summary><strong>3. Binding chooser shows correct tabs and hides unavailable options</strong></summary>
<br>
<blockquote>

**Preconditions:** - You have the JolliMemory VS Code extension installed
- You are in a repository with no existing memory space binding

**Steps:**
1. Open the JolliMemory panel and click **Push to Jolli**
2. When the binding chooser opens, look at the tab bar at the top of the panel
3. Check whether the **Existing space** tab shows a count (e.g. "Existing space (3)")
4. If no spaces exist yet, observe which tab is active by default

**Expected Results:**
- Two tabs are visible: "Existing space (N)" and "New space"
- If no existing spaces are available, the "Existing space" tab is greyed out and the "New space" tab is selected automatically
- Only one tab's content is visible at a time — the other is fully hidden

</blockquote>
</details>
<details>
<summary><strong>4. Cancelling the binding chooser shows a clear cancellation message</strong></summary>
<br>
<blockquote>

**Preconditions:** - You have the JolliMemory VS Code extension installed
- You are in a repository that has never been connected to a JolliMemory space

**Steps:**
1. Open the JolliMemory panel and click **Push to Jolli**
2. When the binding chooser panel opens, close it or click **Cancel** without selecting a space

**Expected Results:**
- An error message appears explaining that the push was cancelled
- No memory is published
- The extension does not attempt to push again automatically

</blockquote>
</details>
<details>
<summary><strong>5. Binding chooser does not reopen if the retry push also fails</strong></summary>
<br>
<blockquote>

**Preconditions:** - You have the JolliMemory VS Code extension installed
- You are in a repository that has never been connected to a JolliMemory space
- The backend is configured to reject bindings (for testing purposes)

**Steps:**
1. Open the JolliMemory panel and click **Push to Jolli**
2. When the binding chooser opens, select a space and click **Confirm**
3. Observe what happens after the retry push attempt

**Expected Results:**
- The binding chooser opens only once — it does not reopen after the retry fails
- A standard error message is shown describing the failure
- The user is not stuck in a loop of repeated chooser dialogs

</blockquote>
</details>
<details>
<summary><strong>6. Plugin version is correctly identified in requests (IntelliJ)</strong></summary>
<br>
<blockquote>

**Preconditions:** - You have the JolliMemory IntelliJ plugin installed
- You have a valid Jolli API key configured

**Steps:**
1. Open IntelliJ and navigate to the JolliMemory tool window
2. Select a commit summary and click **Push to Jolli**
3. Check the network request in your browser's developer tools or a proxy tool (e.g. Charles, Proxyman) for the outgoing push request

**Expected Results:**
- The request includes a header named `x-jolli-client` with a value starting with `intellij-plugin/` followed by the plugin version number
- No `pluginVersion` field appears inside the request body

</blockquote>
</details>
<details>
<summary><strong>7. Plugin version is correctly identified in requests (VS Code)</strong></summary>
<br>
<blockquote>

**Preconditions:** - You have the JolliMemory VS Code extension installed
- You have a valid Jolli API key configured
- You have a network inspection tool available (e.g. Proxyman, Charles)

**Steps:**
1. Open VS Code and navigate to the JolliMemory panel
2. Push a commit summary by clicking **Push to Jolli**
3. Inspect the outgoing HTTP request headers in your network tool

**Expected Results:**
- The request includes a header named `x-jolli-client` with a value starting with `vscode-plugin/` followed by the extension version number
- No `clientInfo` or `pluginVersion` field appears inside the request body

</blockquote>
</details>
<details>
<summary><strong>8. Hidden elements in the binding chooser do not flash on load</strong></summary>
<br>
<blockquote>

**Preconditions:** - You have the JolliMemory VS Code extension installed
- You are in a repository with no existing memory space binding

**Steps:**
1. Open the JolliMemory panel and click **Push to Jolli**
2. Watch the binding chooser panel closely as it first appears
3. Look for any brief flicker of empty boxes or banners before the panel fully loads

**Expected Results:**
- The binding chooser panel appears cleanly with no visible flicker or flash of empty boxes
- The conflict banner and repository hint area are not visible until they are needed
- The panel looks correct immediately on opening

</blockquote>
</details>
<details>
<summary><strong>9. Legacy subfolder field no longer sent on push (modern routing)</strong></summary>
<br>
<blockquote>

**Preconditions:** - You have the JolliMemory VS Code extension installed
- You have a network inspection tool available (e.g. Proxyman, Charles)
- You have a valid Jolli API key configured

**Steps:**
1. Open VS Code and navigate to the JolliMemory panel
2. Push a commit summary by clicking **Push to Jolli**
3. Inspect the outgoing push request body in your network tool

**Expected Results:**
- The request body does not contain a `subFolder` field
- The push completes successfully and the memory appears on the Jolli web frontend
- No routing errors are returned from the server

</blockquote>
</details>
<details>
<summary><strong>10. Race condition banner appears when another user already bound the same repo</strong></summary>
<br>
<blockquote>

**Preconditions:** - Two users share the same Jolli organisation
- Both users are in the same repository with no existing space binding
- Both attempt to bind the repository at nearly the same time

**Steps:**
1. As the second user, open the JolliMemory panel and click **Push to Jolli**
2. When the binding chooser opens, select a space and click **Confirm**
3. Observe the chooser panel after the confirmation attempt

**Expected Results:**
- A banner appears inside the chooser informing you that another binding was already created by someone else
- The banner shows the name of the winning space that was bound
- The tab controls are hidden and replaced by the banner, preventing a duplicate binding

</blockquote>
</details>

---

## Topics (20)
<details>
<summary><strong>01 · Add API service for JM space listing, creation, and repo binding</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The plugin needed a way to call three new backend endpoints — list existing memory spaces, create a new one, and register a repo-to-space binding — as part of the first-commit chooser flow. No HTTP client for these endpoints existed anywhere in the codebase.

- **Scope strictly limited to three endpoints**: Binding move/delete, space rename/delete, and listing bindings were explicitly ruled out of scope for the plugin; those belong to the jolli.ai web frontend. This keeps the plugin's surface area minimal and avoids duplicating management UI.
- **Reuse existing header builder rather than inline**: Extracting `buildJolliApiHeaders` into a shared export avoids drift between push and the new endpoints — any future header change (e.g. new tenant header) only needs one edit.
- **No client-side cache**: An earlier design considered caching the repo→space binding locally to skip a round-trip. This was dropped because the server is the single source of truth, cache invalidation adds complexity disproportionate to the UX benefit, and a 412 from the server self-heals any inconsistency anyway.

**✅ What Was Implemented**

- Created `vscode/src/services/JolliMemoryApiService.ts` with three exported functions: `listJolliMemorySpaces` (GET), `createJolliMemorySpace` (POST), and `createJolliMemoryBinding` (POST)
- Reuses `buildJolliApiHeaders` from `JolliPushService` for Authorization, Content-Type, tenant-slug, and org-slug headers — no header logic duplicated
- Handles both flat-array and `{ spaces: [...] }` envelope shapes from the list endpoint
- Maps `426` → `PluginOutdatedError` and `409 binding_already_exists` → `BindingAlreadyExistsError` centrally inside `sendJson`, so callers stay clean
- Supports both `http://` and `https://` via the same `node:http` / `node:https` pattern used by the push service

**📋 Future Enhancements**

- Server must implement the three endpoints (`GET /api/jolli-memory/spaces`, `POST /api/jolli-memory/spaces`, `POST /api/jolli-memory/bindings`) and the 412/409 response contracts before this client code is exercisable end-to-end.

**📁 FILES**
- `vscode/src/services/JolliMemoryApiService.ts`

</blockquote>
</details>
<details>
<summary><strong>02 · Extend push payload and add binding error classes to push service</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The existing push service sent a minimal payload with no repo identity or client version structure. The new backend contract required the plugin to identify itself with a structured client descriptor, send the canonical repo URL, and handle two new error codes (412 for missing binding, 409 for binding race) so the chooser flow could be triggered.

- **Keep `pluginVersion` for one release**: Dropping it immediately would break pushes to servers not yet upgraded to recognize `clientInfo`. Sending both fields in parallel lets the rollout be decoupled — old servers use `pluginVersion`, new servers use `clientInfo`.
- **Error classes carry structured data, not just a message**: `BindingRequiredError` holds `repoUrl` and `BindingAlreadyExistsError` holds the full winner body so the chooser panel can act on them without re-parsing strings. This makes the call-site logic straightforward and type-safe.
- **412 falls back to payload's `repoUrl` when server omits it**: The server should echo the URL, but if it doesn't, the client already has it in the payload — using it as a fallback avoids a broken error object with an empty URL.

**✅ What Was Implemented**

- Added `clientInfo: { kind, version }`, `repoUrl`, and `relativePath` fields to `JolliPushPayload`; marked `pluginVersion` as deprecated but kept it for one-release forward-compat with older servers
- Added `BindingRequiredError` (thrown on `412 binding_required`) carrying the `repoUrl` the server echoes back
- Added `BindingAlreadyExistsError` (thrown on `409 binding_already_exists`) carrying the winner binding body for the chooser's race-resolution banner
- Extracted `buildJolliApiHeaders` as a named export so `JolliMemoryApiService` can share it without duplication
- `BindingExistsBody` and `ClientInfo` interfaces exported for use by the chooser panel

**📁 FILES**
- `vscode/src/services/JolliPushService.ts`

</blockquote>
</details>
<details>
<summary><strong>03 · Design and scope the JM space binding chooser flow for the plugin</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The team needed to decide exactly what the plugin should do when a user pushes from a repo that has no memory space assigned yet — and equally important, what it should NOT do — before any implementation began.

- **No client-side cache at all**: Initially planned to cache repo→space mappings locally to avoid repeated 412 round-trips. Dropped because the server already stores this authoritatively, cache invalidation logic (TTL, stale detection, 4 inconsistency scenarios) added complexity far exceeding the one saved network call.
- **Chooser as a separate singleton panel, not embedded in the summary panel**: The summary panel is not a singleton — multiple commit tabs can be open simultaneously. A separate singleton chooser prevents two panels from each rendering their own chooser and racing to create the same binding; the second caller simply reveals the already-open chooser.
- **No rename detection or workspace state tracking**: Detecting git remote URL changes and branch renames was designed, then cut. The server handles repo identity by URL; if the URL changes, the next push either succeeds (server migrates) or triggers a new 412 chooser flow — no client tracking needed.
- **Only first-bind, no management UI**: Keeping the plugin to a single new action (first bind) avoids scope creep and keeps the implementation small and auditable.

**✅ What Was Implemented**

The session produced a finalized scope decision (captured in a plan file) rather than code changes in this diff. Key outcomes:
- Plugin responsibility narrowed to exactly one new action: show a chooser when the server returns 412, let the user pick or create a JM space, register the binding, then retry the push once
- All binding lifecycle management (move, delete, rename, list) explicitly moved out of scope to the jolli.ai web frontend
- Client-side binding cache design was proposed, debated, and then fully removed — server is authoritative
- Rename detection (`workspaceState` tracking, branch/repo rename endpoints) was designed and then cut entirely for simplicity
- `BindingChooserWebviewPanel` implemented as a singleton separate from `SummaryWebviewPanel`

**📁 FILES**
- `vscode/src/views/BindingChooserWebviewPanel.ts`

</blockquote>
</details>
<details>
<summary><strong>04 · Remove legacy subfolder field from plugin-to-backend push contract</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The plugin was still sending an old subfolder field on every push, which caused the backend to route those requests through a legacy code path instead of the newer one. The goal was to clean up the contract so the plugin only sends the fields the backend's modern routing logic expects.

- **Scope limited to VS Code only**: IntelliJ still sends the legacy fields and has no `clientInfo`/`repoUrl`, so it continues to hit the backend's legacy path. Touching IntelliJ was explicitly deferred to a separate ticket to avoid scope creep and unreviewed risk on a different platform.
- **No migration shim needed**: All VS Code push call sites already included `clientInfo`, `repoUrl`, and `relativePath`, so `subFolder` was purely redundant. Dropping it immediately (rather than keeping it during a rollout window) was safe and kept the contract clean.
- **Test discriminator switched to `relativePath`**: Since `subFolder` was the only distinguishing field in those test assertions, the tests were updated to use `relativePath` — the field that now carries the same semantic role on the wire.

**✅ What Was Implemented**

- Removed `readonly subFolder?: string` from `JolliPushPayload` interface in `JolliPushService.ts`.
- Dropped `subFolder: "Plans & Notes"` from the plan push call in `SummaryWebviewPanel.ts` (~line 1090) and the note push call (~line 1161).
- Updated two test assertions in `SummaryWebviewPanel.test.ts` to discriminate satellite calls by `relativePath === "main"` instead of `subFolder === "Plans & Notes"`.

**📋 Future Enhancements**

Port the same contract change to IntelliJ (`JolliApiClient.kt`, `SummaryPanel.kt`) so its pushes also include `clientInfo` and `repoUrl` and stop using the legacy backend path.

**📁 FILES**
- `vscode/src/services/JolliPushService.ts`

</blockquote>
</details>
<details>
<summary><strong>05 · Fix webview hidden elements flashing due to blocked inline styles</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The binding chooser webview used inline style attributes to hide elements on initial render, but the VS Code webview security policy only allows styles applied through a nonce-tagged style block. This meant the hidden conflict banner and repo hint could appear as empty visible boxes before JavaScript ran.

- **CSS class over CSSOM mutation**: Although Chromium does allow JavaScript CSSOM mutations (`element.style.display = ...`) even under strict CSP, using a CSS class eliminates the flash of unstyled content on initial paint — the element is hidden before any script runs, not after.
- **`!important` on `.hidden`**: Matches the existing convention in the sibling summary CSS builder, ensuring the utility class wins over component-level specificity without requiring selector gymnastics.
- **Inline margin moved to a compound rule**: Rather than a generic override, the margin was scoped to `.error-message.general-error` to avoid unintended side effects on other error message elements.

**✅ What Was Implemented**

- `BindingChooserCssBuilder.ts`: added `.hidden { display: none !important; }` utility class and a `.error-message.general-error { margin-top: 8px; }` rule to replace the inline margin
- `BindingChooserHtmlBuilder.ts`: replaced `style="display:none;"` on `#repoHint` and `#banner` with the `hidden` class; replaced `style="margin-top:8px;"` on `#generalError` with the `general-error` class
- `BindingChooserScriptBuilder.ts`: replaced `element.style.display = 'block'/'none'` assignments with `classList.remove/add('hidden')` calls

**📁 FILES**
- `vscode/src/views/BindingChooserCssBuilder.ts`
- `vscode/src/views/BindingChooserHtmlBuilder.ts`
- `vscode/src/views/BindingChooserScriptBuilder.ts`

</blockquote>
</details>
<details>
<summary><strong>06 · Add behavioral tests for 412 binding-required chooser retry flow</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The first-time push flow — where a 412 response triggers the binding chooser and then retries the push — had no direct behavioral tests. A regression in this path could silently break the most critical user action (publishing a memory for the first time) while all other unit tests stayed green.

- **Three cases, not one**: The retry path has three meaningfully different outcomes (success, cancel, infinite-loop prevention); collapsing them would leave the anti-loop guard untested, which is the subtlest failure mode.
- **Test the anti-loop guard explicitly**: The guard that prevents a second chooser from opening on a repeated 412 is easy to break silently; a dedicated test makes the contract visible and regression-proof.

**✅ What Was Implemented**

Added three test cases in `SummaryWebviewPanel.handlePush.test.ts` under a new `describe` block covering the 412 retry path:
- **Happy path**: 412 on first push → chooser opens with correct derived params (`baseUrl`, `apiKey`, `repoUrl`, `suggestedRepoName`) → retry succeeds → `pushToJolliResult` posted with `success: true`
- **Cancel path**: chooser resolves `undefined` → no retry → `showErrorMessage` called with the cancellation copy → `pushToJolliResult` posted with `success: false`
- **Anti-loop guard**: both push attempts return 412 → chooser opens exactly once (second 412 surfaces as a normal failure, not another chooser invocation)

**📁 FILES**
- `vscode/src/views/SummaryWebviewPanel.handlePush.test.ts`

</blockquote>
</details>
<details>
<summary><strong>07 · Redesign binding chooser panel with segmented tab layout</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The binding chooser UI had a confusing structure: the "Add to existing space" heading appeared even when no spaces existed, the "Create new" input was always visible (just greyed out), and the repo display-name field was a top-level required section that confused users who just wanted to attach to an existing space.

- **Mode as a toggle, not two sections**: the user is making a single binary choice, so a segmented control maps directly to that mental model and eliminates the contradictory "Add to existing / No spaces yet" state.
- **Hide inactive pane entirely instead of disabling**: hiding removes visual noise (grey empty inputs) and makes the active path unambiguous, at the cost of slightly more JS state management — judged worthwhile for clarity.
- **Repo display-name field placed inline per pane, not in Advanced**: keeping it always visible avoids an extra click for users who want a meaningful name, while its subordinate visual position (indented, lighter border) signals it is secondary to the main choice.
- **No changes to host↔webview message contract**: all existing `init`, `confirm`, `acceptWinner`, and `error` message shapes were preserved so `BindingChooserWebviewPanel.ts` and all integration tests required zero changes.

**✅ What Was Implemented**

Full rebuild of the three builder files for the binding chooser webview:
- `BindingChooserHtmlBuilder.ts`: replaced the two-section-plus-divider layout with a segmented tab control (`Existing space (N)` / `New space`) and two independent mode panes; repo display-name field moved inside each pane as an `inline-attr` sub-field; banner given `role="status"`.
- `BindingChooserScriptBuilder.ts`: introduced a `mode` state variable and `setMode()` function; split the single `repoName` input into `repoNameExisting` and `repoNameNew`; empty-spaces case now locks the Existing tab and auto-switches to New; race-winner banner hides tabs and both panes instead of just disabling the confirm button; validation rewritten per-mode.
- `BindingChooserCssBuilder.ts`: added `.mode-tabs`, `.mode-tab`, `.tab-count`, `.mode-pane`, `.field-row.inline-attr`; removed `.divider`, `.section`, `.section-title`, `.create-new-row`.

**📁 FILES**
- `vscode/src/views/BindingChooserHtmlBuilder.ts`
- `vscode/src/views/BindingChooserScriptBuilder.ts`
- `vscode/src/views/BindingChooserCssBuilder.ts`

</blockquote>
</details>
<details>
<summary><strong>08 · Diagnose retry-after-binding-chooser still returning 412 error</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

After completing the binding chooser flow and confirming a space, the subsequent push still failed with the same "binding required" error, leaving the user unable to push.

- **Root cause attributed to server, not client**: log evidence showed both push attempts used identical `repoUrl` strings, ruling out the client-side normalisation mismatch hypothesis and pointing squarely at the dangling binding row.
- **Upsert over insert for binding creation**: allowing duplicate `(orgId, repoUrl)` bindings creates an unresolvable ambiguity at query time; an upsert ensures the chooser always overwrites the stale record atomically.
- **Client defensive fix still recommended despite not being the root cause**: removing reliance on the server-returned URL is low-risk and eliminates an entire class of future mismatch bugs, so it was flagged as worth doing independently.

**✅ What Was Implemented**

- Investigated three candidate causes: stale deployed extension (ruled out — new UI was visible), mismatched `repoUrl` strings between binding creation and retry push, and a server-side data integrity problem.
- Server logs revealed the real cause: a historical binding (id=3) pointing at a deleted space (id=17) was never cleaned up; the push router kept resolving to that dangling record instead of the newly created binding (id=8 → space 16).
- Identified two required server-side fixes: add a unique constraint on `(orgId, repoUrl)` with upsert semantics on the bindings endpoint, and cascade-delete bindings when a space is deleted.
- Provided an immediate unblock: manually delete or update the dangling binding row in the database.
- Noted a separate defensive client-side improvement: stop passing the server-returned `repoUrl` from the 412 error body to the chooser; always use the locally normalised value to eliminate a potential string-mismatch vector.

**📋 Future Enhancements**

- Server: add unique constraint on `(orgId, repoUrl)` in the bindings table and change the binding creation endpoint to use upsert semantics.
- Server: cascade-delete bindings when their target JM space is deleted.
- Client (`SummaryWebviewPanel.ts`): change the `repoUrl` passed to the chooser to always use the locally normalised value, ignoring the server-returned `err.repoUrl`.

**📁 FILES**

</blockquote>
</details>
<details>
<summary><strong>09 · Move client identity from request body to HTTP header across all clients</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Client identification info (plugin kind and version) was being sent inside the JSON request body, mixed in with business data. The user wanted to move it to a dedicated HTTP header so it could be read by server middleware without parsing the body, and applied consistently across every endpoint automatically.

- **Header over body**: Client identity is transport-layer metadata, not business payload — putting it in a header lets server middleware log and gate on it without touching the body, and one change covers all endpoints automatically.
- **Internal constant, not a parameter**: The VS Code services are VS Code-only (CLI and IntelliJ each have their own clients), so threading `clientInfo` as a function parameter through every call site added complexity with no real benefit; reading it internally from a module-level constant is simpler and eliminates caller mistakes.
- **Delete the compatibility bridge immediately**: The deprecated `pluginVersion` body field was introduced less than a day before this change with no external dependents, so carrying a transition period would only add noise; it was removed outright.
- **Staged rollout required**: The server still validates the old body schema, so the backend must be updated to read the header before these client changes go live — a three-step plan (server adds header support → clients switch → server drops body field) was agreed but deferred to a separate backend PR.

**✅ What Was Implemented**

- New `vscode/src/services/ClientInfo.ts`: defines `ClientInfo` interface and `VSCODE_CLIENT_INFO` constant (reads version from `package.json`)
- `vscode/src/services/JolliPushService.ts`: `buildJolliApiHeaders` now injects `x-jolli-client: vscode-plugin/<version>` automatically; removed `clientInfo` and `pluginVersion` fields from `JolliPushPayload`; updated module doc-comment
- `vscode/src/services/JolliMemoryApiService.ts`: all three binding endpoints inherit the header for free via the shared header builder
- `vscode/src/views/SummaryWebviewPanel.ts`: removed `clientInfo` and `pluginVersion` from the three push callsites; removed the `package.json` version import
- `cli/src/core/LlmClient.ts`: added `x-jolli-client: cli/<__PKG_VERSION__>` to `callProxy` fetch headers using the existing Vite-injected version pattern

**📋 Future Enhancements**

Backend (`PushRouter.ts` in the jolli main repo) needs a coordinated update: remove `pluginVersion` from the zod schema, switch the `MIN_PLUGIN_VERSION` gate to read `x-jolli-client` header, and update corresponding backend tests. Client changes should not be activated until the backend ships first.

**📁 FILES**
- `vscode/src/services/ClientInfo.ts`
- `vscode/src/services/JolliPushService.ts`
- `vscode/src/services/JolliMemoryApiService.ts`
- `cli/src/core/LlmClient.ts`

</blockquote>
</details>
<details>
<summary><strong>10 · Move plugin version from request body into HTTP header on every API call</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The plugin version was being sent inside the JSON body of push and delete requests, but the server needs it available on every request — including deletes, which have no body — so it can identify the caller and gate access by version.

- **Header over body**: Sending the version as a header rather than a JSON field makes it uniformly available on all request types (including DELETE, which carries no body), and lets the server inspect it without parsing the payload.
- **Caller-supplied version**: The version is passed in by the call site rather than resolved inside the API client because the IntelliJ runtime cannot read it statically — it must be fetched from the workspace's bundled metadata at runtime, which is already handled by the UI layer.

**✅ What Was Implemented**

- Removed `pluginVersion` field from `JolliPushPayload` data class in `JolliApiClient.kt`
- Added `clientVersion: String` parameter to both `pushToJolli` and `deleteFromJolli` functions
- Both functions now send `x-jolli-client: intellij-plugin/<version>` as an HTTP header on every request
- Updated all call sites in `SummaryPanel.kt` to pass `pluginVersion` as the new explicit argument instead of embedding it in the payload
- Added a class-level KDoc comment to `JolliApiClient` explaining the header's purpose and why the version is passed by the caller rather than resolved statically

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/SummaryPanel.kt`

</blockquote>
</details>
<details>
<summary><strong>11 · Bake IntelliJ plugin version into JAR via Gradle processResources</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The IntelliJ plugin was reading the VS Code extension's `package.json` from the user's workspace to get a version string for the `x-jolli-client` header — a fragile approach that sent the wrong version number since the backend gates each client type independently.

- **Build-time resource injection over `PluginManagerCore` runtime API (Plan B over Plan A)**: Keeps `JolliApiClient` free of IntelliJ Platform dependencies, matching its stated "pure HTTP client" posture analogous to the VS Code TypeScript port; also makes unit testing trivial without IDE fixtures or mocks.
- **Fallback value `0.0.0` instead of a real version string**: Intentionally fails the server's minimum-version gate loudly to surface packaging mistakes, rather than silently shipping a misleading hardcoded version like the previous `0.90.160`.
- **Version resolved inside `JolliApiClient` rather than passed by callers**: Callers (`SummaryPanel`) had IDE context but `JolliApiClient` is a plain `object`; moving the lookup inside the client eliminates the parameter from every call site and keeps the concern co-located with the HTTP layer.

**✅ What Was Implemented**

- Added `intellij/src/main/resources/jollimemory-plugin-version.txt` template containing `${version}`
- Wired `processResources` in `build.gradle.kts` to expand the template with `project.version` at build time, with `inputs.property` to bust Gradle's incremental cache on version bumps
- Added `pluginVersion: String by lazy { loadPluginVersion() }` to `JolliApiClient` object, reading from classpath resource `/jollimemory-plugin-version.txt` with fallback `0.0.0`
- Removed `clientVersion: String` parameter from `pushToJolli` and `deleteFromJolli`; both now use `pluginVersion` directly
- Deleted `getPluginVersion()` method and `FALLBACK_PLUGIN_VERSION = "0.90.160"` constant from `SummaryPanel`
- Added `PluginVersion` nested test asserting the resolved version matches semver and is not `0.0.0`

**📁 FILES**
- `intellij/build.gradle.kts`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/SummaryPanel.kt`
- `intellij/src/main/resources/jollimemory-plugin-version.txt`

</blockquote>
</details>
<details>
<summary><strong>12 · Remove create-space flow from plugin, bind to existing spaces only</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Creating a JM Space from inside the VS Code plugin was deemed high-risk because it pushes governance decisions (permissions, naming, org ownership) into the IDE. The product boundary should be: plugin selects an existing space, all space management lives on the web frontend.

- **Remove from API client, not just UI**: Hiding the button would leave a live `POST /spaces` code path in the plugin. Deleting the function enforces the product boundary at the code level and prevents accidental re-exposure.
- **Auto-derive repoName in extension host**: Letting the user edit the repo display name inside the IDE is another governance decision that belongs on the web. Using `suggestedRepoName` automatically keeps the plugin's responsibility minimal.
- **Disable bind button when no spaces exist**: Rather than redirecting to a create flow, the empty state now explains the web-first path. This avoids any IDE-side creation affordance while still giving the user a clear next step.
- **Regression tests on generated HTML/JS strings**: Since the builders produce raw strings, snapshot-style `not.toContain` assertions are the most direct way to prevent the removed UI from creeping back in.

**✅ What Was Implemented**

- Removed `createJolliMemorySpace` / `POST /api/jolli-memory/spaces` from `JolliMemoryApiService` entirely (not just hidden in UI)
- Removed the "New space" / "Existing space" segmented tab UI from `BindingChooserHtmlBuilder` and all associated CSS (`.mode-tabs`, `.mode-tab`, `.field-row`, text inputs)
- Stripped `setMode`, `mode` state, and all "new" branch logic from `BindingChooserScriptBuilder`; webview now only posts `{ command: 'confirm', jmSpaceId }` — no `repoName`, no `choice`
- `repoName` for the binding payload is now derived server-side from `suggestedRepoName` in the extension host (`BindingChooserWebviewPanel`), not entered by the user
- Empty-space state now shows "Create one on jolli.ai" guidance and disables the Bind button instead of auto-switching to a create form
- Updated `WebviewMessage` union type to remove the `choice: 'new'` variant
- Added regression tests in `BindingChooserHtmlBuilder.test.ts` and `BindingChooserScriptBuilder.test.ts` asserting no `New space`, `newSpaceName`, or `repoName` surface in generated output

**📁 FILES**
- `vscode/src/services/JolliMemoryApiService.ts`
- `vscode/src/views/BindingChooserWebviewPanel.ts`
- `vscode/src/views/BindingChooserHtmlBuilder.ts`
- `vscode/src/views/BindingChooserScriptBuilder.ts`
- `vscode/src/views/BindingChooserCssBuilder.ts`

</blockquote>
</details>
<details>
<summary><strong>13 · Flatten push relativePath to bare branch slug for all doc types</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Summary pushes were going to `<branchSlug>/commits` while plan/note pushes went to `<branchSlug>`, creating an asymmetric on-disk layout. The goal was to unify all three doc types under a single flat per-branch folder.

- **Merge two helpers into one**: The two path helpers were functionally identical after the flatten decision; a single `buildBranchRelativePath` removes the asymmetry and eliminates the risk of a future caller accidentally using the wrong helper
- **No server-side migration in this commit**: Existing summary docs already stored at `<branchSlug>/commits/<x>` are left in place; the plan (§3 #9) calls for a separate server migration job to reparent them — doing it here would mix client and server concerns
- **Alt B chosen over Alt A**: Rather than keeping `<branchSlug>/commits` for summaries and adding `docType` as redundant metadata (Alt A), the team chose full flattening (Alt B) with a planned server migration, accepting the transition-window complexity for a cleaner long-term layout

**✅ What Was Implemented**

- Replaced `buildCommitRelativePath` (returned `<branchSlug>/commits`) and `buildSatelliteRelativePath` (returned `<branchSlug>`) with a single `buildBranchRelativePath` that returns just `sanitizeBranchSlug(branch)` for all doc types
- Updated all three `pushToJolli` call sites in `SummaryWebviewPanel.ts` (summary, plan, note) to use `buildBranchRelativePath`
- Updated `relativePath` JSDoc to reflect the new flat contract

**📋 Future Enhancements**

- Server must implement the `<branchSlug>/commits` → `<branchSlug>` reparent migration (JM-SPACE plan §3 #9) before this reaches prod; without it, existing summary docs update in place at the legacy path and the "flat" promise only holds for new pushes
- IntelliJ plugin still uses the old `subFolder` field with no `relativePath` — needs a separate modernization pass to match the VS Code push contract

**📁 FILES**
- `vscode/src/views/SummaryUtils.ts`

</blockquote>
</details>
<details>
<summary><strong>14 · Add required docType field to push payload to disambiguate flat-path docs</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Once all doc types share the same flat `<branchSlug>` path, the server has no structural way to distinguish summaries from plans from notes. A new body field was needed to carry that distinction for `sourceMetadata` and frontend TreeItem icons.

- **`docType` over `kind`**: The field was initially proposed as `kind`, then `docKind`, and finally settled on `docType`. `kind` clashed semantically with the existing `x-jolli-client: <kind>/<version>` header; `docType` is self-documenting and avoids the ambiguity without being verbose
- **Required, not optional**: Making `docType` optional would allow a buggy client to silently omit it, causing the server to mis-tag every push. Requiring it forces every call site to be explicit at compile time
- **String literal enum over numeric or URL-based discrimination**: Self-documenting on the wire, trivial to debug from network logs; URL-based discrimination (`/push/jollimemory/summary`) would multiply endpoint surface and split error-handling logic three ways
- **Single flat interface over discriminated union**: The payload shape is identical across all three doc types today (only `title` differs, already handled by separate builder functions), so splitting into a discriminated union would be over-engineering for no compile-time narrowing benefit

**✅ What Was Implemented**

- Added `readonly docType: "summary" | "plan" | "note"` as a required field on `JolliPushPayload` interface
- Each of the three `pushToJolli` call sites in `SummaryWebviewPanel.ts` now passes the appropriate literal (`"summary"`, `"plan"`, `"note"`)
- Updated the JOLLI-1335 contract docstring to clarify that `<kind>` in the `x-jolli-client` header is *client* kind, distinct from the body's `docType`
- Added a new `JolliPushService` test that round-trips all three `docType` values through the serialized request body

**📋 Future Enhancements**

- Backend must accept `docType` in `PushRequestSchema` and write it to `sourceMetadata.docType` before this ships
- Frontend TreeItem icon routing on `sourceMetadata.docType` is not yet implemented (deferred per JM-SPACE plan §4.6.a)
- IntelliJ `JolliPushPayload` data class does not yet include `docType`

**📁 FILES**
- `vscode/src/services/JolliPushService.ts`
- `vscode/src/views/SummaryUtils.ts`

</blockquote>
</details>
<details>
<summary><strong>15 · Remove dead field that falsely implied a re-initialization guard existed</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A field in the binding chooser panel was being set and then immediately discarded with a comment claiming it prevented repeated initialization on transient errors. No code actually read the field, so the guard did not exist — it was misleading scaffolding left over from generation.

- **Delete rather than implement**: The guard was not needed by any current caller, so adding real logic would be premature. The right place for deduplication of repeated `ready` messages would be in the message handler if the need arises — YAGNI applies here.
- **Remove the comment too**: Leaving a comment that describes a non-existent guard is worse than no comment; it actively misleads future readers about the panel's behavior.

**✅ What Was Implemented**

Removed the `spacesLoaded` field declaration, its assignment after the spaces API call, and the `void this.spacesLoaded` expression plus its misleading comment from `BindingChooserWebviewPanel.ts`.

**📁 FILES**
- `vscode/src/views/BindingChooserWebviewPanel.ts`

</blockquote>
</details>
<details>
<summary><strong>16 · Audit whether VS Code plugin monitors repo name and branch name changes</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Developer needed to understand whether the plugin could detect `git remote set-url` or branch rename events in order to notify the backend via `/jollimemory/rename-rep` and `/jollimemory/rename-branch`.

- **No implementation change made**: The audit was informational; the developer did not request a fix in this session. Adding a `.git/config` watcher plus API calls was identified as the required approach if needed in future.

**✅ What Was Implemented**

- Confirmed branch changes ARE monitored: `.git/HEAD` is watched via `watchFile` with create+change events (handles Windows atomic rename via `HEAD.lock`)
- Confirmed repo name (remote URL) changes are NOT monitored: `GitRemoteUtils.ts` only reads `remote.origin.url` on demand; no `FileSystemWatcher` on `.git/config` exists
- Neither `rename-rep` nor `rename-branch` endpoints are called anywhere in the plugin

**📋 Future Enhancements**

Adding `.git/config` FileSystemWatcher and calling `/jollimemory/rename-rep` / `/jollimemory/rename-branch` was explicitly flagged as future work if repo rename detection is required.

**📁 FILES**
- `vscode/src/Extension.ts`
- `vscode/src/util/GitRemoteUtils.ts`

</blockquote>
</details>
<details>
<summary><strong>17 · Clarify why IntelliJ plugin passes clientInfo through API call chain</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Developer questioned why `x-jolli-client` version was threaded as a parameter through the call chain rather than resolved internally, and challenged the accuracy of the comment claiming IntelliJ's own version would be rejected by the server.

- **Corrected the misleading code comment rather than preserving it**: The old comment in `SummaryPanel.kt:849-855` described a false constraint; keeping it would have misled future contributors about why the design existed.

**✅ What Was Implemented**

- Identified that the existing code comment was factually wrong: the backend gates each client type (`intellij-plugin/`, `vscode-extension/`, `cli/`) independently, so IntelliJ's own version number would not be rejected
- Confirmed the real (valid) reason for passing version as a parameter was architectural: `JolliApiClient` is a plain `object` without IDE context, so callers with IDE access computed the version and passed it in
- This analysis directly motivated the decision to implement Plan B (build-time resource injection) to fix both the wrong version source and the awkward parameter threading

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/SummaryPanel.kt`

</blockquote>
</details>
<details>
<summary><strong>18 · Debug 426 Upgrade Required error after moving version to header</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

After the plugin moved `pluginVersion` from the request body to the `x-jolli-client` header, clicking "Update to Jolli" showed a "plugin is outdated" error, suggesting the backend was still reading the body field and seeing an empty version.

- **Dual-write over immediate cutover**: Since backend deploys first but old clients still exist in the wild, dual-write prevents false `426` errors during the rollout window without requiring a coordinated simultaneous deploy.
- **Surface err.message in UI**: The current fixed-string error hides whether the server saw version `0.0.0`, `unknown`, or a genuinely old version — making it impossible to distinguish a real outdated client from a contract mismatch.

**✅ What Was Implemented**

- Identified root cause: frontend/backend contract mismatch — client had already migrated to header, backend still expected body field `pluginVersion`
- Recommended transitional dual-write strategy: send version in both header (`x-jolli-client`) and body (`pluginVersion`) until backend is fully deployed
- Noted that the fixed UI error message swallows `err.message`, masking the actual server response; suggested surfacing `err.message` in the modal for easier diagnosis

**📋 Future Enhancements**

Remove the body `pluginVersion` field from the client once all backend instances are confirmed to be reading from the `x-jolli-client` header.

**📁 FILES**
- `vscode/src/services/JolliMemoryApiService.ts`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 1, 2026 at 8:06 PM*


flyerli2025:
Review findings:

1. `vscode/src/views/SummaryWebviewPanel.ts:958`
Using `err.repoUrl || repoUrl` here lets the server dictate the repo key we bind. If the echoed value is stale or normalized differently from the local canonical URL, we can create a binding for one key and then retry the push with another, so the single retry still comes back `412`. We should always bind using the locally normalized `repoUrl`.

2. `vscode/src/views/BindingChooserWebviewPanel.ts:107`
`currentInstance` is global to the whole extension host, so any second binding flow resolves `undefined` as soon as another chooser is open, even if it came from a different repo/workspace. The caller treats `undefined` as cancellation, which means repo B can fail spuriously while repo A's chooser is open. This singleton likely needs to be keyed by repo/workspace instead of process-wide.

3. `intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliApiClient.kt:280`
The PR description says client identity moved into the `x-jolli-client` header on backend requests, and the class comment now says it is sent on every request. `pushToJolli` and `deleteFromJolli` were updated, but `callLlmProxy` still omits the header here, so IntelliJ recap/translation requests remain unversioned on the wire.

foster-han:
> Review findings:
> 
> 1. `vscode/src/views/SummaryWebviewPanel.ts:958`
>    Using `err.repoUrl || repoUrl` here lets the server dictate the repo key we bind. If the echoed value is stale or normalized differently from the local canonical URL, we can create a binding for one key and then retry the push with another, so the single retry still comes back `412`. We should always bind using the locally normalized `repoUrl`.
> 2. `vscode/src/views/BindingChooserWebviewPanel.ts:107`
>    `currentInstance` is global to the whole extension host, so any second binding flow resolves `undefined` as soon as another chooser is open, even if it came from a different repo/workspace. The caller treats `undefined` as cancellation, which means repo B can fail spuriously while repo A's chooser is open. This singleton likely needs to be keyed by repo/workspace instead of process-wide.
> 3. `intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliApiClient.kt:280`
>    The PR description says client identity moved into the `x-jolli-client` header on backend requests, and the class comment now says it is sent on every request. `pushToJolli` and `deleteFromJolli` were updated, but `callLlmProxy` still omits the header here, so IntelliJ recap/translation requests remain unversioned on the wire.

Done

flyerli2025:
Findings

vscode/src/util/GitRemoteUtils.ts:12-18,41-58
The new repo canonicalization still preserves path casing, so the same GitHub repo can normalize to different repoUrl keys across clones, e.g. git@github.com:JolliAI/Jolli.git becomes https://github.com/JolliAI/Jolli while https://github.com/jolliai/jolli.git becomes https://github.com/jolliai/jolli. Since bindings are keyed by exact repoUrl, that means one teammate can bind the repo and another teammate can still hit 412 binding_required from a differently-cased clone, or even create a duplicate binding. That breaks the stated “same physical repo must always yield the same string” invariant.

vscode/src/views/BindingChooserWebviewPanel.ts:98-116, vscode/src/views/SummaryWebviewPanel.ts:950-962
The per-repo singleton avoids cross-repo interference, but same-repo concurrent pushes still fail incorrectly. If a chooser is already open for repo A, openAndAwait() returns undefined to the second caller instead of waiting on the existing promise; runJolliPush() interprets that the same way as an actual cancel and shows “Push cancelled”. Because pushInProgress is only per summary panel, two open commit tabs from the same repo can still hit this path and the second one gets a spurious failure even though the user never cancelled anything.

vscode/src/views/BindingChooserScriptBuilder.ts:117-123
The chooser auto-selects the first returned space, and the API client does not sort or otherwise guarantee that order. With this PR, binding is a persistent repo-level action and the plugin no longer offers any in-plugin repair flow, so a user can permanently bind the repo to an arbitrary space just by clicking “Bind and push” without making an explicit choice. This feels especially risky because the empty-state/product copy frames this as a one-time repo association, not a lightweight preference.

flyerli2025 (approved):

PR #63, opened 2026-05-03
Add CSRF state validation (RFC 6749) to OAuth login flow

flyerli2025:
## Quick recap

The developer added CSRF state validation to both the CLI and VSCode OAuth login flows, implementing RFC 6749 section 10.12. A 256-bit random nonce is generated when the user initiates sign-in, included on the login URL, and must be echoed unchanged by the server on the callback. Mismatched or missing state causes the callback to be rejected with a 400 error, preventing the attacker's code from being exchanged for credentials.

The implementation uses constant-time string comparison to guard against timing attacks and clears the pending nonce after first use, blocking replay attempts. Backward compatibility is preserved for servers older than version 1270 by skipping state validation on the legacy token-in-URL fallback path; this temporary gap closes when that fallback is eventually removed. The test suite covers the happy path, missing state, mismatched state, length-matched content mismatches, and the one-shot consumption guarantee.

---

## E2E Test (4)
<details>
<summary><strong>1. OAuth login succeeds with valid CSRF state parameter</strong></summary>
<br>
<blockquote>

**Steps:**
1. Open the Jolli application and click the "Sign In" button
2. Wait for the browser to open with a login page
3. Complete the login process on the server (enter credentials, approve access)
4. Verify that the browser redirects back to the application and displays a success message
5. Confirm that you are now logged in and can access your account

**Expected Results:**
- The login completes without errors
- You see a "Sign in successful" or similar confirmation message
- Your account information appears in the application (e.g., your name or profile)

</blockquote>
</details>
<details>
<summary><strong>2. OAuth login rejects callback without state parameter (CSRF attack prevention)</strong></summary>
<br>
<blockquote>

**Steps:**
1. Open the Jolli application and click the "Sign In" button
2. Note the login URL that appears in the browser address bar
3. Manually navigate to the callback URL (the one shown in the browser) but remove the "state" parameter from the URL
4. Press Enter to submit the modified URL
5. Observe the error message that appears

**Expected Results:**
- The page displays an error message containing "state mismatch" or "Invalid sign-in callback"
- You are NOT logged in
- You are prompted to try signing in again

</blockquote>
</details>
<details>
<summary><strong>3. OAuth login rejects callback with mismatched state parameter</strong></summary>
<br>
<blockquote>

**Steps:**
1. Open the Jolli application and click the "Sign In" button
2. Note the login URL that appears in the browser address bar and copy the state parameter value
3. Manually navigate to the callback URL but change the state parameter to a different value (e.g., replace one character)
4. Press Enter to submit the modified URL
5. Observe the error message that appears

**Expected Results:**
- The page displays an error message containing "state mismatch" or "Invalid sign-in callback"
- You are NOT logged in
- You are prompted to try signing in again

</blockquote>
</details>
<details>
<summary><strong>4. Legacy token-based login still works without state parameter (backward compatibility)</strong></summary>
<br>
<blockquote>

**Preconditions:** You have an older Jolli server (version before 1270) that uses the legacy token-in-URL login method

**Steps:**
1. Open the Jolli application and initiate the login flow
2. Complete the login on the older server
3. Verify that the callback succeeds even though the state parameter is not present
4. Confirm that you are logged in with your token

**Expected Results:**
- The login completes successfully despite the missing state parameter
- You see a success message and are logged into your account
- The application does not reject the callback for missing state

</blockquote>
</details>

---

## Topic (1)
<details>
<summary><strong>01 · Add CSRF state validation to OAuth login flow</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The OAuth login flow lacked CSRF protection, leaving it vulnerable to attacks where an attacker could trick a user into completing a login initiated by the attacker. The fix implements RFC 6749 state parameter validation to ensure callbacks originate from the login flow the user actually started.

- **One-shot nonce consumption prevents replay attacks**: The pending state is cleared immediately after validation (before comparing), using synchronous assignment with no `await` gaps in between. This ensures an attacker who obtains a valid nonce can only use it once; a second callback with the same state fails because `pendingState` is null. The design was verified to be safe against VSCode's URI handler dispatch model, which is single-threaded and sequential.
- **Constant-time comparison is necessary despite the nonce's strength**: Although a 256-bit random nonce makes timing leaks infeasible in practice, using `timingSafeEqual` costs nothing and keeps the comparison correct-by-construction. The helper function is duplicated in CLI and VSCode rather than shared because the nonce format has no cross-client synchronization requirement (unlike API key parsing, which must match between CLI and IntelliJ). Sharing would create hidden coupling that constrains future refactoring.
- **State validation is scoped to the code-exchange path only**: The legacy token-in-URL path bypasses state checking because pre-1270 servers don't echo state and enforcing it would immediately break sign-in for users on those versions. This creates a temporary security gap, but it closes naturally when the fallback is removed during a future server upgrade cycle.

**✅ What Was Implemented**

- Added 256-bit random nonce generation (`randomBytes(32).toString("hex")`) in both CLI and VSCode login flows. The nonce is generated before opening the browser and passed to the login server as `expectedState`.
- Implemented state parameter validation in the callback handler: the server's response must echo the nonce unchanged, verified using constant-time comparison (`timingSafeEqual`) to prevent timing attacks. Mismatched or missing state on code-exchange callbacks returns a 400 error and prevents credential exchange.
- Preserved backward compatibility with pre-1270 servers by skipping state validation on the legacy `?token=` fallback path, which predates state support. This prevents locking out users on older server versions while the upgrade window remains open.

**📁 FILES**
- `cli/src/auth/Login.ts`
- `cli/src/auth/Login.test.ts`
- `vscode/src/services/AuthService.ts`
- `vscode/src/services/AuthService.test.ts`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 3, 2026 at 8:31 PM*


flyerli2025:
### Code review

Found 1 issue:

1. Typo: "racially" should be "rapidly" in the one-shot rationale comment. The intended meaning is "rapidly fires two callbacks" (i.e., in quick succession). The current wording is a loaded word that's worth fixing before this code is open-sourced.

https://github.com/jolliai/jolliai/blob/23a93751415d0896945245775dfa8464d3d9d65f/vscode/src/services/AuthService.ts#L111-L116

sanshizhang-jolli:
### 🐛 Bug: `constantTimeStringEqual` length check is byte-vs-char mismatched

Both copies — [`cli/src/auth/Login.ts`](cli/src/auth/Login.ts) and [`vscode/src/services/AuthService.ts`](vscode/src/services/AuthService.ts) — have:

```ts
function constantTimeStringEqual(a: string, b: string): boolean {
    if (a.length !== b.length) return false;
    return timingSafeEqual(Buffer.from(a), Buffer.from(b));
}
```

`String.prototype.length` returns **UTF-16 code units**, but `Buffer.from(a)` defaults to **UTF-8 bytes**. They only agree for ASCII input.

**Trigger:** an attacker submits a `state=…` whose JS char length equals the expected nonce length (64) but contains a non-ASCII character — e.g. 63 hex chars + one `é`. `.length === 64` passes the guard, but the resulting Buffers have different byte lengths, so `timingSafeEqual` throws:

```
RangeError [ERR_CRYPTO_TIMING_SAFE_EQUAL_LENGTH]: Input buffers must have the same byte length
```

**Impact:**

- **CLI** (`Login.ts`, in the `createServer` async handler, *before* the `try { … }` block): the throw becomes an unhandled rejection. No HTTP response is sent, `closeServer` is not called, and `browserLogin`'s Promise never resolves — the user is stuck and has to Ctrl-C.
- **VSCode** (`AuthService.handleAuthCallback`): the returned Promise rejects with `RangeError` instead of resolving to `{ success: false, error: "Invalid sign-in callback (state mismatch)…" }`. Whether that surfaces cleanly depends on the URI-handler caller.

This is **not** an auth bypass — the comparison still fails closed in the security sense — but it's a usable DoS against an in-flight sign-in attempt and breaks the documented error contract.

**Suggested fix** — convert to Buffer first, then compare byte length:

```ts
function constantTimeStringEqual(a: string, b: string): boolean {
    const ba = Buffer.from(a);
    const bb = Buffer.from(b);
    if (ba.length !== bb.length) return false;
    return timingSafeEqual(ba, bb);
}
```

**Suggested test** — add a case in both `Login.test.ts` and `AuthService.test.ts` that sends a non-ASCII `state` whose JS string length matches the expected nonce length, and asserts the call resolves to a clean `state mismatch` rejection (no unhandled error, no hang).

Since the helper is duplicated, the fix needs to land in both files — which is also a small argument for sharing it (the CLI is already canonical for `parseJolliApiKey` / `assertJolliOriginAllowed` and the VSCode bundle inlines `cli/src/**` at build time).

sanshizhang-jolli (approved):

PR #67, opened 2026-05-05
feat(openapi): replace swagger-ui-react with a per-endpoint MDX pipeline

summerfang-jolli:
Ports the rich OpenAPI rendering surface into the jolliai CLI. Replaces the old `OpenApiRenderer.ts` (a 4-line swagger-ui-react MDX shim) with a parser → IR → emitter pipeline that produces one MDX page, one JSON sidecar, and a custom React component per endpoint.

Architecture
------------

The OpenAPI surface is split in two so a future Fumadocs / Docusaurus emitter doesn't have to reimplement parsing:

  cli/src/site/openapi/                            framework-agnostic IR
    SpecLoader, SpecParser, RefResolver, CodeSampleGenerator,
    SchemaExample, OpenApiPipeline, Slug, ReservedWords, Escape,
    SpecName, Types

  cli/src/site/renderer/nextra/                    Nextra emitter
    Components, EndpointPageEmitter, EndpointDataEmitter,
    OverviewPageEmitter, SidebarMetaEmitter, ApiCss, Paths

`SiteRenderer.renderOpenApiSpecs(contentDir, publicDir, specs)` is the contract between them. Adding a new docs framework is a sibling directory under renderer/ — the IR layer stays untouched.

What's in the pipeline
----------------------

* SpecLoader.tryParseOpenApi — real `yaml`/`JSON.parse` returning the validated AST. Powers content-based discovery in ContentMirror; any .json / .yaml / .yml file structurally valid as OpenAPI is picked up regardless of filename. ContentMirror caches the parsed AST in MirrorResult.openapiDocs so the rich-renderer pipeline never re-reads or re-parses source files.
* SpecParser.parseFullSpec — walks paths × HTTP_METHODS in declaration order, follows $ref with RFC 6901 ~1/~0 escapes, throws on (tag, operationId) collisions (silently dropping an endpoint would be worse), merges path-level + operation-level parameters with operation-level overriding on (name, in).
* CodeSampleGenerator — five hand-rolled samples (cURL, JS, TS, Python, Go) with `toPythonLiteral` / `goStringLiteral` helpers that avoid the regex-based replacement regression that corrupted strings containing literal true/false/null or backticks.
* SchemaExample.exampleFromSchema — depth-limited synthesis. $ref / oneOf / anyOf / allOf / enum / default / nullable not honoured; documented in the test file so a future enhancement doesn't quietly regress.
* OpenApiPipeline.buildPipeline — single entry point that emitters consume verbatim (`ParsedSpec` + per-operation `EndpointDossier`s with pre-rendered code samples).

What the Nextra emitter writes
------------------------------

For each spec under `<buildDir>/`:

  content/api-{spec}/index.mdx                   per-spec overview
  content/api-{spec}/_refs.ts                    shared schema map
  content/api-{spec}/_meta.ts                    top-level sidebar
  content/api-{spec}/_data/{operationId}.json    per-op JSON sidecar
  content/api-{spec}/{tag}/_meta.ts              per-tag sidebar
  content/api-{spec}/{tag}/{operationId}.mdx     per-endpoint MDX shim

Once per build (in initProject, alongside Nextra scaffolding):

  components/api/Endpoint.tsx + 8 sibling components
  styles/api.css                                 method/status pills,
                                                 2-column grid, dark
                                                 mode aligned with
                                                 Nextra tokens

`tsconfig.json` declares `paths: { "@/*": ["./*"] }`, so each MDX shim imports via `@/components/api/Endpoint` — robust to any future change in MDX directory depth.

Deletions / breaking changes
----------------------------

* `swagger-ui-react` removed from NEXTRA_DEPENDENCIES (~600 KB shaved off every generated user-site bundle) and from ContentRules.safeImportPrefixes. Existing EngineManager hash check detects the dependency change and triggers a one-time engine reinstall on next run.
* `cli/src/site/OpenApiRenderer.{ts,test.ts}` deleted — its 144-line swagger-ui-react shim has no caller after Phase 4.
* `SiteRenderer.renderOpenApiFiles(sourceRoot, contentDir, files, publicDir)` replaced by `renderOpenApiSpecs(contentDir, publicDir, specs[])` — each input carries a pre-built OpenApiPipelineResult, so emitters only emit; they never re-parse the source file.

Documentation
-------------

* `cli/DEVELOPMENT.md` gains a "Site Generation: OpenAPI Reference Pipeline" section with a file-flow diagram, per-module tables, and a four-step recipe for adding a new docs framework.
* `CLAUDE.md` gets a short subsection under "Architecture you can't infer from one file" pointing to the cli doc for detail.

Tests
-----

2671 passed, 0 failed. New tests cover SpecParser collision detection and RFC 6901 escape handling; CodeSampleGenerator language emission edges (Python literal encoding for true/false/null inside strings, Go raw-string fallback for backtick-containing payloads); SchemaExample intentional gaps (locked in so a refactor doesn't quietly regress); the Nextra emitter end-to-end against a real temp directory; and the api.css selector / hue-token shape.

Coverage 98.9 / 96.18 / 99.4 / 99.01 (thresholds 97/96/97/97).

## Changes

-
-

## Testing

- [ ] Added/updated unit tests
- [ ] Ran `./gradlew test` locally
- [ ] Tested in sandbox IDE via `./gradlew runIde`
- [ ] Tested on: (e.g. macOS + IntelliJ 2025.1)

## Checklist

- [ ] Code follows the existing style
- [ ] Documentation updated (README / CHANGELOG) if user-facing
- [ ] No new warnings from Plugin Verifier

## Jolli Memory

https://jolli.ai/very/articles?doc=2076

## E2E Test (4)
<details>
<summary><strong>1. API overview page loads without crashing when endpoints have path parameters</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a documentation project with an OpenAPI spec that includes at least one endpoint whose summary or path contains curly-brace parameters, such as `/users/{username}` or `/items/{id}`.

**Steps:**
1. Open the documentation site in a browser.
2. Navigate to the API overview page (usually listed as the top-level API section in the sidebar).
3. Check that the page fully loads and displays a table of endpoints.
4. Look at the table rows and confirm that path parameters like `{username}` or `{id}` appear as plain text inside the table cells.
5. Scroll through the entire table to confirm no rows are missing or broken.

**Expected Results:**
- The API overview page loads completely without any error message or blank screen.
- Curly-brace path parameters appear as readable text in the table, not as broken symbols or empty spaces.
- All endpoints from the spec are listed in the table.

</blockquote>
</details>
<details>
<summary><strong>2. Documentation site builds successfully without a missing page error</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a documentation project set up and ready to build using the `jolli build` command.

**Steps:**
1. Run the build command for your documentation project.
2. Wait for the build process to complete.
3. Check the terminal output for any error messages.
4. If the build succeeds, open the built site in a browser.
5. Navigate to a URL that does not exist on the site (for example, add `/this-page-does-not-exist` to the end of the site address).

**Expected Results:**
- The build completes without any crash or error about a missing page.
- The terminal shows a successful build message.
- Visiting a non-existent URL on the built site shows a "page not found" message rather than a browser-level error.

</blockquote>
</details>
<details>
<summary><strong>3. API overview table cells with backslashes or pipe characters display correctly</strong></summary>
<br>
<blockquote>

**Preconditions:** Have an OpenAPI spec where at least one endpoint summary contains a backslash (`\`) or a pipe character (`|`).

**Steps:**
1. Open the documentation site and navigate to the API overview page.
2. Find the table row for the endpoint whose summary contains a backslash or pipe character.
3. Read the text in the summary column of that row.
4. Check whether the backslash or pipe appears as a single character, not doubled or broken.

**Expected Results:**
- The summary text in the table cell shows exactly one backslash where one was written, not two.
- Pipe characters appear as the `|` symbol, not as a broken table column separator.
- The table layout remains intact with no extra or misaligned columns.

</blockquote>
</details>
<details>
<summary><strong>4. Per-endpoint API pages load and display code samples correctly</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a documentation project with an OpenAPI spec containing at least one endpoint that requires authentication and has a request body.

**Steps:**
1. Open the documentation site and navigate to the API section in the sidebar.
2. Click on any individual endpoint page listed under a tag group.
3. Check that the page displays the endpoint details, including the HTTP method, path, and description.
4. Look for the code samples section and click through the available language tabs (such as cURL, JavaScript, Python).
5. For the cURL sample, check that the full URL shown is wrapped in single quotes.

**Expected Results:**
- The endpoint page loads without any error.
- Code samples are shown for multiple languages and can be switched between tabs.
- The cURL sample displays the complete URL inside single quotes, making it safe to copy and run directly in a terminal.
- The page does not show any blank sections or broken components.

</blockquote>
</details>

---

## Topics (3)
<details>
<summary><strong>01 · Fix PR review findings in OpenAPI renderer pipeline</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A PR review identified several bugs and code quality issues in the new OpenAPI rendering pipeline that needed to be addressed before merge. The reviewer flagged one high-priority bug (unsafe CSS selector construction) and several medium/low issues around defensive coding and dead parameters.

- **CSS selector safety over simplicity**: Used `CSS.escape()` rather than switching to a DOM traversal approach, because it is the minimal targeted fix with no structural change to the component.
- **Fix ordering for double-escape bug**: MDX escaping was moved before pipe escaping because `escapeMdxText` also escapes backslashes — running it after pipe-escaping caused the inserted backslashes to be double-escaped. Correct order eliminates the bug without changing either function's logic.
- **Remove dead parameter rather than keep it**: The unused `_parsed` parameter was fully removed (including from tests and docs) rather than left with an underscore prefix, because it added confusion about the function's actual dependencies with no benefit.
- **Scope of fix**: The pre-existing `OverviewPageEmitter` double-escape bug was fixed in the same commit even though it wasn't in the original review findings, because it was discovered during test verification and was directly related to the escaping work.

**✅ What Was Implemented**

- `Components.ts`: Wrapped the `active` value in `CSS.escape()` before interpolating it into a `querySelector` selector string, preventing silent copy-button failures when pane identifiers contain CSS-special characters.
- `CodeSampleGenerator.ts`: Applied single-quote shell escaping to the full URL string in cURL output (not just the body), and changed `operation.security[0]` to use optional chaining for defensive access.
- `SpecParser.ts`: Replaced an unsafe `as string | undefined` cast on `operationId` with an explicit `typeof` check, preventing malformed specs from silently producing bad slugs.
- `EndpointPageEmitter.ts` + `index.ts` + `EndpointPageEmitter.test.ts`: Removed the unused `_parsed` parameter from `emitEndpointPage` and updated all callers and tests accordingly.
- `OverviewPageEmitter.ts`: Fixed a pre-existing double-escape bug where pipe-escaping ran before MDX escaping, causing backslashes to be double-escaped in table cells. Reordered to MDX-escape first, then pipe-escape.
- `DEVELOPMENT.md`: Updated the function signature reference to match the removed parameter.

**📁 FILES**
- `cli/src/site/renderer/nextra/Components.ts`
- `cli/src/site/openapi/CodeSampleGenerator.ts`
- `cli/src/site/openapi/SpecParser.ts`
- `cli/src/site/renderer/nextra/EndpointPageEmitter.ts`
- `cli/src/site/renderer/nextra/OverviewPageEmitter.ts`

</blockquote>
</details>
<details>
<summary><strong>02 · Fix build crash caused by curly braces in API summary table cells</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The Next.js build was failing with an "acorn parse error" on a generated API overview page. Path parameters like `{username}` in endpoint summaries were causing the MDX parser to crash, even though the code was already attempting to escape those characters.

- **HTML entities over backslash escapes**: Backslash-escaping `{` and `}` works in normal MDX body text, but inside markdown table cells the parser switches context and hands expression parsing to acorn — which still sees the braces and crashes. HTML entities are opaque to the JSX expression parser, so they eliminate the crash entirely.
- **Ampersand escaped first**: Replacing `&` before the other characters prevents double-encoding (e.g. `&lt;` becoming `&amp;lt;`) when the input already contains HTML entities.
- **Inline chain instead of shared helper**: The existing shared escape helper only handled backslash escapes, which were the wrong tool for table cells. Inlining the replacement chain in the table-cell code path makes the table-specific behavior explicit and avoids silently breaking other callers of the helper.

**✅ What Was Implemented**

- Replaced the `escapeMdxText` helper call in `OverviewPageEmitter.ts` with a direct inline replacement chain that uses HTML entities (`&amp;`, `&lt;`, `&#123;`, `&#125;`) instead of backslash escapes for special characters.
- Markdown pipe (`|`) and backslash (`\`) are still escaped last using the existing backslash approach, since those only affect markdown table syntax, not JSX parsing.

**📁 FILES**
- `cli/src/site/renderer/nextra/OverviewPageEmitter.ts`

</blockquote>
</details>
<details>
<summary><strong>03 · Fix static export build failure caused by missing 404 page</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

After fixing an earlier MDX parsing error, the documentation site build was still failing during static page generation. Next.js 15 requires a custom not-found page to exist in the project scaffold, and the scaffold generator wasn't producing one, causing every build to crash at the final export step.

- **Re-export from theme package rather than custom implementation**: Using the themed 404 component from `nextra-theme-docs` keeps the not-found page visually consistent with the rest of the docs site at zero maintenance cost, matching the pattern from Nextra v4's own example projects.
- **Fix in the scaffold generator rather than at runtime**: The missing file was a gap in the project template, so the right fix was to add it to the generator — every future build gets the correct scaffold automatically without any user action.

**✅ What Was Implemented**

- Added `generateNotFound()` to `cli/src/site/NextraProjectWriter.ts`, which returns the contents of an `app/not-found.tsx` file that re-exports `NotFoundPage` from `nextra-theme-docs`.
- Updated `initNextraProject()` in the same file to write `app/not-found.tsx` to the build directory alongside the other scaffolded files.

**📁 FILES**
- `cli/src/site/NextraProjectWriter.ts`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 5, 2026 at 11:01 AM*


cli/src/site/openapi/Escape.ts
```diff
@@ -0,0 +1,92 @@
...
+export function escapeYaml(str: string): string {
+	const hasSpecialChars = /[:#[\]{}|>`]/.test(str) || str.includes('"') || str.includes("'");
+	const startsWithDigit = /^[0-9]/.test(str);
+	const isSpecialValue = YAML_SPECIAL_VALUES.has(str.toLowerCase());
+
+	if (hasSpecialChars || startsWithDigit || isSpecialValue) {
+		return `"${str.replace(/\\/g, "\\\\").replace(/"/g, '\\"')}"`;
+	}
+	return str;
+}
```
aidancrosbie, line 29: **Show-stopper:** `escapeYaml` does not handle newlines, so a multi-line OpenAPI `summary` propagates verbatim into the per-endpoint MDX YAML frontmatter and crashes the Next.js build with `YAMLParseError: Implicit keys need to be on a single line`.

**Why it matters:** OpenAPI does not forbid multi-line summaries and real-world specs ship them. Reproduced locally with a single `summary: |` block in `edge-cases.yaml` — the entire site fails to build, not just the affected page. Any spec field that flows through `escapeYaml` is vulnerable (e.g. `info.title` in `OverviewPageEmitter`).

**Suggested fix:** treat `\n` / `\r` as special characters that trigger quoting, then JSON-style-escape them in the quoted output. Drop-in replacement for the function:

```suggestion
export function escapeYaml(str: string): string {
	const hasSpecialChars = /[:#[\]{}|>`\n\r]/.test(str) || str.includes('"') || str.includes("'");
	const startsWithDigit = /^[0-9]/.test(str);
	const isSpecialValue = YAML_SPECIAL_VALUES.has(str.toLowerCase());

	if (hasSpecialChars || startsWithDigit || isSpecialValue) {
		return `"${str
			.replace(/\\/g, "\\\\")
			.replace(/"/g, '\\"')
			.replace(/\n/g, "\\n")
			.replace(/\r/g, "\\r")}"`;
	}
	return str;
}
```

Note: this resolves the build-crash path. The summary will then render with a literal `\n` in the page `<title>`, which is ugly but survivable. The related table-cell case in `OverviewPageEmitter.ts:65-70` is distinct — markdown tables cannot host raw newlines even when quoted, so that path needs the summary collapsed (e.g. `.replace(/\s*\n\s*/g, " ")`) before the entity-encoding chain. Reasonable as a follow-up.

aidancrosbie (approved):

PR #68, opened 2026-05-05
Add GitHub Copilot CLI and VS Code Copilot Chat as transcript sources

flyerli2025:
## Commits in this PR (2)

1. Add GitHub Copilot CLI as a transcript source (`b7d3971`) 
2. Add VS Code Copilot Chat as a transcript source (`1de42e7`)

## Plans & Notes (6)

- GitHub Copilot CLI Support — Implementation Plan
- GitHub Copilot CLI as a Transcript Source — Design
- VS Code Copilot Chat as a Transcript Source — Design
- VS Code Copilot Chat Redesign — Read `events.jsonl`, Not OpenTelemetry Traces
- VS Code Copilot Chat Support Implementation Plan
- Copilot Chat events.jsonl Redesign — Implementation Plan

### Commit 1 of 2: Add GitHub Copilot CLI as a transcript source (`b7d3971`)

The developer added GitHub Copilot CLI as a new transcript source, enabling Jolli Memory to automatically capture and summarize AI-assisted development work from the Copilot command-line interface. Copilot stores conversations in a local SQLite database at ~/.copilot/session-store.db; the integration discovers sessions matching the current project directory and extracts conversation turns, then feeds them into the post-commit pipeline alongside transcripts from other AI agents like OpenCode and Codex. Users can enable or disable Copilot session discovery via the configure command, and the status command displays session counts and any database scan errors. The VS Code extension surfaces Copilot in its status tree, summary details view, and settings panel, bringing it to feature parity with other supported integrations.

To support multiple database-backed sources without duplication, the developer extracted shared SQLite helpers into a dedicated module that provides database access primitives, error classification, and version checking. This module is now used by both OpenCode and Copilot CLI, with the planned Cursor integration adopting the same pattern. The extraction uses lazy imports to avoid emitting Node's ExperimentalWarning during unrelated operations, and maintains backward compatibility through deprecated re-exports in the OpenCode module.

Error handling was aligned across all SQLite-backed sources in the post-commit pipeline. Previously, Copilot failures were isolated to individual sessions while OpenCode and Cursor failures could abandon entire batches. Now all three sources use per-session error isolation, preventing a single corrupted or locked database from preventing the rest of the batch from being summarized. This improves resilience in production environments where transient database locks are common. The implementation includes proper error classification for corrupted or locked databases, defensive handling of schema drift, and comprehensive test coverage matching the OpenCode pattern. Cross-platform path handling was fixed to ensure tests pass on Windows, macOS, and Linux by normalizing fixture paths using the same platform-aware functions as the production code. All changes maintain architectural symmetry with the existing OpenCode integration, reducing cognitive load for future maintainers. The system automatically enables Copilot support during installation when it detects the tool is present, and users can enable or disable it via the configure command. Developers who work primarily in the terminal now have their AI-assisted work automatically tracked and summarized alongside their git commits, without requiring manual export or tool switching.

### Commit 2 of 2: Add VS Code Copilot Chat as a transcript source (`1de42e7`)

The developer added VS Code Copilot Chat as a new transcript source, enabling the system to capture conversations from the Chat extension alongside the existing Copilot CLI integration. Both forms now share a single `copilotEnabled` configuration toggle and are discovered, read, and displayed together in the UI, reflecting how users perceive GitHub Copilot as a unified product.

The implementation spans four core modules: a shared workspace locator for finding VS Code storage directories (also refactored into the Cursor integration to eliminate duplication), a detector that probes both the VS Code extension's global storage and the Copilot CLI agent's `~/.copilot/session-state` directory, a session discoverer with a two-scan architecture supporting both storage locations, and a transcript reader with format-specific sub-readers for patch-log and event-stream formats. The reader uses semantic-level cursors based on request indices and line counts rather than byte offsets, making it robust against variable line lengths and streaming token accumulation. All new code achieved 100% test coverage.

The Copilot Chat integration is wired into the existing discovery and processing pipelines: the installer auto-enables the shared `copilotEnabled` flag when either CLI or Chat is detected, the queue worker discovers and reads chat sessions alongside other AI assistant sessions with time-bounded filtering support, and status reporting components display detection status for both forms separately. Users can see at a glance which Copilot variants are available through sub-line breakdowns in the status command and status tree, and chat sessions appear in the Summary panel with the same configuration control as the base Copilot source.

## E2E Test (10)
<details>
<summary><strong>1. [b7d3971] GitHub Copilot CLI integration appears in status command</strong></summary>
<br>
<blockquote>

**Steps:**
1. Open a terminal and run `jolli status`
2. Look for a "Copilot:" row in the integration breakdown section
3. If Copilot CLI is installed on your system, verify the row shows a session count (e.g., "3 sessions")
4. If Copilot CLI is not installed, verify the Copilot row does not appear at all
5. If Copilot is installed but you have disabled it in config, verify the row shows "detected but disabled"

**Expected Results:**
- When Copilot is installed and enabled: the status output includes a "Copilot:" row with the number of discovered sessions
- When Copilot is not installed: no Copilot row appears in the output
- When Copilot is detected but disabled: the row appears with a "detected but disabled" message

</blockquote>
</details>
<details>
<summary><strong>2. [b7d3971] Copilot CLI sessions are discovered and included in commit summaries</strong></summary>
<br>
<blockquote>

**Preconditions:** Have GitHub Copilot CLI installed and have used it to have a conversation in the current project directory within the last 48 hours

**Steps:**
1. Open a terminal in your project directory
2. Run `jolli status` and confirm the Copilot row shows at least 1 session
3. Make a small change to a file and commit it with `git commit -m "test"`
4. Wait for the post-commit hook to complete (check the git output for any messages)
5. Run `jolli status` again and verify the session count has been processed

**Expected Results:**
- The Copilot row in the status output shows the number of recent sessions from your project
- The commit hook completes without errors
- Sessions older than 48 hours are not included in the count

</blockquote>
</details>
<details>
<summary><strong>3. [b7d3971] Copilot CLI is auto-detected and enabled on first install</strong></summary>
<br>
<blockquote>

**Preconditions:** Have GitHub Copilot CLI installed on your system

**Steps:**
1. Run `jolli configure --list-keys` and look for "copilotEnabled" in the output
2. Run `jolli configure --set copilotEnabled=true` to explicitly enable it
3. Run `jolli status` and verify the Copilot row appears
4. Run `jolli configure --set copilotEnabled=false` to disable it
5. Run `jolli status` again and verify the Copilot row shows "detected but disabled"

**Expected Results:**
- The "copilotEnabled" key appears in the configure command's list of valid keys
- You can set copilotEnabled to true or false
- The status command reflects your configuration choice immediately

</blockquote>
</details>
<details>
<summary><strong>4. [b7d3971] Copilot database errors are reported without breaking the commit hook</strong></summary>
<br>
<blockquote>

**Preconditions:** Have GitHub Copilot CLI installed with at least one session in the current project

**Steps:**
1. Run `jolli status` and note the current Copilot session count
2. Make a small file change and commit it with `git commit -m "test"`
3. Wait for the post-commit hook to complete
4. Run `jolli status` again

**Expected Results:**
- If the Copilot database is inaccessible (locked, corrupted, or permission denied), the status output shows "unavailable" next to the Copilot row with an error description
- The commit hook completes successfully even if Copilot data cannot be read
- Other integrations (Claude, OpenCode, Cursor) continue to work normally

</blockquote>
</details>
<details>
<summary><strong>5. [b7d3971] Old Copilot sessions are filtered out automatically</strong></summary>
<br>
<blockquote>

**Preconditions:** Have GitHub Copilot CLI installed

**Steps:**
1. Run `jolli status` and note the Copilot session count
2. Verify that only sessions from the last 48 hours are included (you can check the timestamps in your Copilot CLI history if available)
3. If you have sessions older than 48 hours, they should not appear in the count

**Expected Results:**
- Sessions created more than 48 hours ago do not appear in the Copilot session count
- Only recent sessions (within the last 48 hours) are discovered and included in commit summaries
- This prevents pulling months of historical conversations when Copilot is first enabled

</blockquote>
</details>
<details>
<summary><strong>6. [1de42e7] Copilot Chat sessions appear in status report alongside Copilot CLI</strong></summary>
<br>
<blockquote>

**Preconditions:** Have VS Code installed with the Copilot Chat extension enabled, or have the Copilot CLI agent backend installed at ~/.copilot/session-state with at least one chat session file.

**Steps:**
1. Open a terminal and run the status command to view the current integration status
2. Look for the "Copilot:" row in the output
3. Check that the row shows a breakdown line below it displaying "CLI: ✓" or "✗" and "Chat: ✓" or "✗" to indicate which Copilot variants are detected
4. Verify that the session count on the Copilot row includes sessions from both CLI and Chat sources combined
5. If Copilot Chat is detected but the chat scan failed, verify that an additional indented line appears below the CLI/Chat breakdown showing "Chat scan failed" with the error type and message

**Expected Results:**
- The Copilot row should show as detected if either CLI or Chat (or both) are present
- The CLI/Chat breakdown should clearly show which variant(s) are available with checkmarks or X marks
- Session counts should reflect the total from both sources
- Any chat-specific scan errors should appear as a separate sub-line, not mixed with the main Copilot row

</blockquote>
</details>
<details>
<summary><strong>7. [1de42e7] Copilot Chat sessions are discovered and included in commit context</strong></summary>
<br>
<blockquote>

**Preconditions:** Have VS Code with Copilot Chat extension installed and at least one active chat session in the current project, or have ~/.copilot/session-state with chat session files matching the current project folder.

**Steps:**
1. Open the application and navigate to a project that has Copilot Chat sessions
2. Trigger a commit or build operation that would normally scan for AI assistant sessions
3. Check the summary or session list to verify that Copilot Chat sessions appear alongside other AI assistant sessions (Claude, Copilot CLI, etc.)
4. Click on or inspect a Copilot Chat session to confirm its transcript can be read and displayed

**Expected Results:**
- Copilot Chat sessions should appear in the session list with source labeled as "Copilot Chat"
- Sessions should be ordered consistently (Copilot Chat should appear immediately after Copilot CLI in any sorted lists)
- Transcripts from Copilot Chat should be readable and display conversation history

</blockquote>
</details>
<details>
<summary><strong>8. [1de42e7] Copilot Chat is auto-enabled when detected during setup</strong></summary>
<br>
<blockquote>

**Preconditions:** Have VS Code with Copilot Chat extension installed (or Copilot CLI backend at ~/.copilot/session-state), and the Copilot feature toggle is not yet configured in settings.

**Steps:**
1. Run the installer or initialization process
2. Allow it to auto-detect installed integrations
3. Check the configuration settings to verify that the Copilot feature toggle is now enabled
4. Open the status report and confirm that Copilot Chat is listed as detected

**Expected Results:**
- The Copilot feature should be automatically enabled if either Copilot CLI or Copilot Chat is detected
- The installer log should indicate which Copilot variant(s) were found
- The status report should show Copilot as enabled with the appropriate CLI/Chat breakdown

</blockquote>
</details>
<details>
<summary><strong>9. [1de42e7] Copilot Chat detection status appears in the status tree view</strong></summary>
<br>
<blockquote>

**Preconditions:** Have VS Code with Copilot Chat extension installed, or Copilot CLI backend at ~/.copilot/session-state.

**Steps:**
1. Open the application and navigate to the status or integrations tree view
2. Look for the Copilot integration row
3. Check the tooltip or description text for inline indicators showing CLI and Chat detection status separately
4. If only Chat is detected (no CLI), verify the row still shows as detected and displays "Chat: ✓"
5. If a chat scan error occurred, verify it appears as a separate warning row labeled "Copilot Chat scan failed" (distinct from any CLI scan error)

**Expected Results:**
- The Copilot integration row should show detection status for both CLI and Chat forms
- Each form should have a visible indicator (✓ or ✗) in the tooltip or description
- Chat-specific errors should surface as a separate warning row, not combined with CLI errors
- The integration is treated as detected if either form is present

</blockquote>
</details>
<details>
<summary><strong>10. [1de42e7] Summary panel displays Copilot Chat sessions with correct labeling</strong></summary>
<br>
<blockquote>

**Preconditions:** Have Copilot Chat sessions discovered in the current project (see Scenario 2 preconditions).

**Steps:**
1. Open the Summary panel or session statistics view
2. Look for the source breakdown showing counts by AI assistant type
3. Verify that "Copilot Chat" appears as a distinct label in the source list
4. Check that Copilot Chat sessions are ordered immediately after Copilot CLI sessions in any sorted displays
5. Confirm that the Copilot Chat source is included in the enabled sources filter when the Copilot feature is enabled

**Expected Results:**
- "Copilot Chat" should appear as a labeled source in the summary
- Session counts should correctly attribute sessions to the Copilot Chat source
- Ordering should be consistent with other AI assistant sources
- Toggling the Copilot feature off should exclude both Copilot CLI and Copilot Chat sessions from the summary

</blockquote>
</details>

## Topics (25)
<details>
<summary><strong>01 · [b7d3971] Add GitHub Copilot CLI as a transcript source with full integration</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The system supported multiple AI coding agents but not GitHub Copilot CLI. Copilot stores conversations in a local SQLite database and needed to be integrated as a new source for session discovery and transcript reading.

- **Node SQLite requirement over pure-JavaScript libraries**: Copilot's database uses write-ahead logging (WAL) mode, which only Node's built-in node:sqlite module can read. Requiring Node 22.6+ ensures correctness and avoids silent data loss, with graceful degradation on older versions.
- **Synthetic transcript path format (dbPath#sessionId)**: Copilot stores all sessions in a single database, unlike OpenCode which uses separate files. Encoding both the database location and session ID in the path allows the reader to locate the correct session without additional metadata, keeping the discovery and reading interfaces consistent.
- **Case-insensitive path matching on macOS and Windows**: Copilot stores the working directory as-is, but filesystems on macOS and Windows are case-insensitive. Matching with case-insensitive comparison ensures sessions are discovered even when users invoke the tool with different path casing, improving reliability without filesystem queries.
- **Post-filter staleness in JavaScript rather than SQL**: The updated_at column is TEXT (ISO format), so SQL comparison would be lexicographic rather than temporal. Filtering after Date.parse tolerates any format that JavaScript's date parser accepts and reuses the existing schema-drift tolerance pattern already in the codebase.
- **Error visibility over silent degradation**: The discovery wrapper logs scan errors even though the pipeline continues without sessions. This surfaces real problems (permission issues, database corruption) to operators rather than silently excluding data.

**✅ What Was Implemented**

- Created three core modules (CopilotDetector, CopilotSessionDiscoverer, CopilotTranscriptReader) that detect Copilot installation, discover sessions matching the current project directory, and extract conversation turns from the SQLite database at ~/.copilot/session-store.db. The detector verifies Node 22.6+ support for reading write-ahead-logged databases.
- Extended the type system to include "copilot" in the TranscriptSource union and added configuration fields (copilotEnabled, copilotDetected) and status fields (copilotScanError) to track Copilot state and report database scan failures.
- Integrated Copilot discovery into the post-commit pipeline (QueueWorker.ts) with per-session error isolation, allowing the pipeline to discover and read Copilot sessions alongside Codex and OpenCode without cascading failures.
- Added a 48-hour staleness filter to exclude sessions older than 48 hours, preventing users enabling Copilot for the first time from pulling months of historical sessions into their next commit summary. The filter is applied in JavaScript after timestamp parsing rather than in SQL, since the database stores timestamps as ISO text.
- Added auto-detection during installation: when Copilot is found and the user has not explicitly set a preference, the system enables Copilot support and logs the detection.
- Surfaced Copilot in the CLI status and configure commands, and added toggles in the VS Code extension's status tree, summary details view, and settings panel, bringing Copilot to feature parity with other supported integrations.

**📁 FILES**
- `cli/src/Types.ts`
- `cli/src/core/CopilotDetector.ts`
- `cli/src/core/CopilotSessionDiscoverer.ts`
- `cli/src/core/CopilotTranscriptReader.ts`
- `cli/src/hooks/QueueWorker.ts`

</blockquote>
</details>
<details>
<summary><strong>02 · [b7d3971] Extract shared SQLite helpers for multi-agent database access</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The codebase needed to support reading SQLite databases from multiple external AI agents. Rather than duplicating database access logic across multiple discoverer modules, common SQLite utilities needed to be extracted into a reusable helper module.

- **Lazy module imports preserve startup performance**: The withSqliteDb function dynamically imports node:sqlite only when invoked, not when the module is transitively loaded. This prevents the ExperimentalWarning from appearing in unrelated code paths while keeping the module safely importable by any code without side effects.
- **Structural typing over direct type imports**: The SqliteDbHandle interface is hand-written to match node:sqlite's DatabaseSync shape rather than importing the actual type. This keeps the module loadable in all Node versions without triggering the ExperimentalWarning at import time, with the trade-off of manual maintenance of the interface shape.
- **Backward compatibility through re-exports**: The old OpenCodeSessionDiscoverer module re-exports the new helpers under deprecated names, allowing existing callers to continue working without changes. This staged migration approach lets the codebase transition gradually and avoids breaking changes when Cursor's PR is rebased.
- **Single source of truth for error types**: Importing SqliteScanError ensures that if the error kind union is extended in the future, all locations automatically stay in sync. The alternative—maintaining separate copies—would require manual updates and create silent divergence risk.

**✅ What Was Implemented**

- Created SqliteHelpers.ts containing database access primitives: withSqliteDb() function for opening and closing read-only SQLite connections, SqliteDbHandle interface for structural typing, and version-checking utilities (hasNodeSqliteSupport, NODE_SQLITE_MIN_VERSION). These utilities use lazy imports to avoid emitting Node's ExperimentalWarning during unrelated operations.
- Moved error classification logic into classifyScanError() function, which categorizes database failures (corrupt, locked, permission, schema, unknown) while treating missing files as silent.
- Updated OpenCodeSessionDiscoverer.ts to import from the new helpers module and re-export deprecated aliases to maintain backward compatibility. Updated OpenCodeTranscriptReader.ts to use the new withSqliteDb import directly.
- Replaced inlined error type shapes in StatusInfo with direct imports of SqliteScanError from SqliteHelpers.ts, eliminating structural duplication across three locations.

**📁 FILES**
- `cli/src/core/SqliteHelpers.ts`
- `cli/src/core/OpenCodeSessionDiscoverer.ts`
- `cli/src/Types.ts`

</blockquote>
</details>
<details>
<summary><strong>03 · [b7d3971] Align error handling for SQLite-backed transcript sources in post-commit pipeline</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Asymmetric error handling between OpenCode and Copilot transcript readers: Copilot failures were wrapped in try/catch to skip individual sessions, but OpenCode and Cursor failures could abandon entire batches, causing one corrupted or locked database to prevent the rest of the batch from being summarized.

Per-session isolation is the safer default for any SQLite-backed source: a single corrupted or transiently locked database should never prevent the rest of the batch from being summarized. This principle was already applied to Copilot in the original implementation; extending it uniformly to OpenCode and Cursor eliminates a subtle inconsistency that would only surface in production when a user hit a locked database. JSONL sources are excluded from this change because they handle per-line failures internally and have different error semantics.

**✅ What Was Implemented**

- Wrapped readOpenCodeTranscript() in try/catch with per-session error logging and continue, matching the isolation pattern already applied to Copilot.
- Wrapped readCursorTranscript() in the same try/catch pattern for consistency, since Cursor also uses SQLite and shares identical failure modes (lock, corruption, schema drift, TOCTOU disappearance).
- Wrapped readCopilotTranscript() in try/catch to prevent pipeline crashes if the session database becomes inaccessible during the read phase, matching the resilience pattern already in place for other transcript sources.
- Added a clarifying comment explaining that all three SQLite-backed sources (OpenCode, Cursor, Copilot) share the same failure modes and should all use per-session isolation, while JSONL readers (Gemini, Claude) handle failures internally.

**📁 FILES**
- `cli/src/hooks/QueueWorker.ts`

</blockquote>
</details>
<details>
<summary><strong>04 · [b7d3971] Comprehensive test coverage for Copilot discovery, reading, and integration</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The initial test suite had gaps in edge case coverage and asymmetries with the parallel OpenCode test patterns, creating maintenance risk and potential for silent regressions in error paths.

- **Parity with OpenCode test patterns**: Rather than inventing new test structures, the Copilot tests now mirror the exact same edge cases and combinations already proven in OpenCode. This reduces cognitive load for future maintainers and ensures both integrations have equivalent regression protection.
- **Spy-based mocking over module mocking**: The spy-based approach is more reliable than module mocking for testing Node version checks because it directly targets the runtime condition being tested (Node version) without requiring module cache manipulation. Module mocking introduces timing and isolation risks that are unnecessary when the actual condition can be mocked at a lower level.
- **Mocks in test infrastructure over production code changes**: Mocks were added as a targeted fix to the test infrastructure rather than modifying the production code. This approach isolates the test environment from external dependencies and I/O operations, allowing the test suite to run reliably without network or filesystem access.
- **Platform-aware path normalization in fixtures**: Both sides of the test contract must use the same normalization: path.resolve() is platform-dependent (returns /p on POSIX but <currentDrive>:\p on Windows). Since the production code calls resolve() to normalize paths before SQL matching, test fixtures must also call resolve() on the same input strings so both sides produce identical normalized values. Using a single platformPath helper makes the intent explicit and centralizes the normalization logic.
- **Time-relative test fixtures**: Hardcoded ISO timestamps in test seeds would silently cross the 48-hour staleness threshold once the test suite is more than 2 days old, causing tests to fail without any code change. Using Date.now() relative offsets keeps fixtures fresh indefinitely and matches the pattern already established in OpenCode tests.

**✅ What Was Implemented**

- Added missing edge case tests for parseSyntheticPath in CopilotTranscriptReader.test.ts to cover empty dbPath and empty sessionId branches, ensuring both error conditions are explicitly verified.
- Added combined cursor and beforeTimestamp test in CopilotTranscriptReader.test.ts to verify the two-phase filtering pattern (resume from cursor, then apply timestamp cutoff) works correctly in sequence, matching the equivalent test in OpenCodeTranscriptReader.test.ts.
- Added Linux platform-specific test in CopilotSessionDiscoverer.test.ts to verify case-sensitive matching behavior, confirming the exact-match SQL path is active on Linux rather than relying on darwin test coverage alone.
- Added test case to verify sessions are returned in descending order by updated_at timestamp when multiple sessions exist.
- Added test coverage for the status command when Copilot is not detected, verifying the Copilot integration row is not rendered when the copilotDetected field is undefined.
- Added test case verifying that when Copilot is detected but disabled (copilotEnabled is false), the status row renders with a "detected but disabled" message.
- Added branch coverage tests for error paths: permission-denied errors and unknown error codes in CopilotDetector.ts, database open failures and NULL timestamp handling in CopilotTranscriptReader.ts, and error handling when reading a Copilot transcript fails in QueueWorker.ts.
- Fixed a test isolation bug in CopilotSessionDiscoverer.test.ts where vi.restoreAllMocks() did not unwind mockRejectedValue calls applied inside module mock factories, causing stat rejection to leak between tests. The fix explicitly resets the stat mock in beforeEach.
- Added mock implementations for three Copilot modules in PostCommitHook.test.ts to prevent timeout failures caused by unmocked Copilot-related modules making real I/O calls during test execution.
- Fixed cross-platform path handling in Copilot detector and session discoverer tests: modified CopilotDetector.test.ts to compute the expected path using path.join() instead of hardcoded strings, and added a platformPath helper function in CopilotSessionDiscoverer.test.ts that wraps path.resolve() to normalize all fixture cwd values and test input paths to the same platform-specific format. Replaced hardcoded /tmp/does-not-exist with join(tmpdir(), "copilot-disc-missing", "x.db") to use the system temporary directory, making the "DB missing" test portable across Windows, macOS, and Linux.
- Refactored test fixtures in CopilotSessionDiscoverer.test.ts to use a time-relative helper (isoFromNow) instead of hardcoded ISO timestamps, preventing tests from silently failing once they cross the 48-hour boundary. Added a new test case that verifies sessions older than 49 hours are filtered while fresh sessions are retained.

**📁 FILES**
- `cli/src/core/CopilotTranscriptReader.test.ts`
- `cli/src/core/CopilotSessionDiscoverer.test.ts`
- `cli/src/core/CopilotDetector.test.ts`

</blockquote>
</details>
<details>
<summary><strong>05 · [b7d3971] Auto-detect Copilot CLI on install and surface detection status</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The installer needed to automatically discover when GitHub Copilot CLI is present on the user's system and enable session discovery for it, similar to how it already handles other AI tools like Claude and OpenCode.

- **Respect explicit user configuration**: When copilotEnabled is already set (either true or false), the auto-detection does not override it. This preserves user intent and prevents unwanted re-enablement if a user has deliberately disabled Copilot support.
- **Lazy session discovery on status check**: Copilot sessions are scanned on-demand during getStatus() rather than stored persistently in sessions.json like other sources. This mirrors the existing pattern for OpenCode and avoids coupling the session store to Copilot's database format.
- **Surface scan errors transparently**: When Copilot session scanning fails (e.g., database locked), the error is captured and returned in the status object rather than silently swallowed. This allows the CLI to report integration issues to the user without crashing the status command.

**✅ What Was Implemented**

- Added Copilot detection logic to the install flow in Installer.ts: when installation runs, it checks if Copilot is installed via isCopilotInstalled(), and if the copilotEnabled config is undefined, it automatically enables Copilot session discovery and logs the detection.
- Extended the status reporting in getStatus() to include Copilot detection and enablement state: the function now calls scanCopilotSessions() when Copilot is detected and enabled, collects any scan errors, and surfaces copilotDetected, copilotEnabled, and copilotScanError fields in the returned status object.
- Updated test mocks and test cases in Installer.test.ts to cover Copilot detection: added mocks for CopilotDetector and CopilotSessionDiscoverer, and added test cases verifying auto-enable behavior, config preservation, and error surfacing in status.

**📁 FILES**
- `cli/src/install/Installer.ts`

</blockquote>
</details>
<details>
<summary><strong>06 · [b7d3971] Design specification and implementation plan for Copilot CLI integration</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Before implementing Copilot CLI support, the team needed to document the design, verify the actual on-disk format and runtime behavior of Copilot's data store, and establish clear scope boundaries to prevent future scope creep.

- **Exact cwd matching over time-window heuristics**: Copilot stores the working directory directly on each session row, making workspace attribution precise. The design chose simple path normalization without symlink chasing, avoiding the complexity Cursor needed while still handling trailing-slash edge cases.
- **Read-only database access**: The spec mandates read-only SQLite handles to avoid collision with Copilot's own writer process and the inuse.<pid>.lock semantics, eliminating a class of potential data corruption bugs.
- **Reuse existing SqliteHelpers abstraction**: Rather than introducing a new SQLite wrapper or dependency, the design imports the SqliteHelpers module extracted during the Cursor work, keeping the dependency graph flat and ensuring consistent error classification across all SQLite-backed sources.

**✅ What Was Implemented**

- Created a comprehensive design specification document that details the integration architecture, observed database schema, and implementation roadmap. The spec includes hands-on verification results from testing against a live Copilot CLI process, confirming that the SQLite database with WAL mode is readable via the existing node:sqlite library.
- Established the integration as discovery-based (scanning ~/.copilot/session-store.db for sessions matching the current working directory) rather than hook-based, since Copilot CLI exposes no hook surface. The design reuses the existing Cursor integration template across five layers: type definitions, session discovery, transcript reading, CLI commands, and VSCode UI panels.
- Documented three new core modules with their responsibilities, plus the specific changes needed in SessionTracker, QueueWorker, Installer, and six VSCode panel files. Included an explicit "Intentionally not done" section listing out-of-scope decisions (no use of checkpoint metadata, no Windows CI testing, no hook integration) to prevent future scope creep.
- Created a 1842-line implementation plan document specifying the complete integration strategy, including architecture decisions, file structure, nine implementation phases with explicit commit ordering, and detailed task checklists with test fixtures and expected outcomes.
- Expanded documentation to explain why Copilot checkpoints are excluded: they are user-triggered conversation-compression nodes cut on context-window boundaries rather than git-commit boundaries, making them unsuitable as the primary transcript source. Checkpoints are optional and their schema is more volatile than the stable turns table.

**📁 FILES**
- `docs/superpowers/specs/2026-05-05-copilot-cli-support-design.md`
- `docs/superpowers/plans/2026-05-05-copilot-cli-support.md`

</blockquote>
</details>
<details>
<summary><strong>07 · [b7d3971] Fix error handling and improve code documentation</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Code review identified that scan error types were duplicated across three locations in the type definitions, error handling in the detector was silently swallowing non-ENOENT stat failures that should be logged, and configuration help text and code comments needed refinement to accurately reflect runtime behavior.

Clearer documentation prevents misunderstanding: improved documentation of the scope and intent of shared utilities, and clarified why defensive code branches exist, reducing confusion about coverage exclusion and integration errors.

**✅ What Was Implemented**

- Fixed error handling in CopilotDetector.isCopilotInstalled() to log non-ENOENT stat failures at warn level, preventing silent loss of permission or I/O errors.
- Added a debug log statement in QueueWorker.ts when Copilot sessions are skipped, indicating the session ID and the reason. This provides visibility into the skip path without changing behavior.
- Updated the copilotEnabled configuration key description in ConfigureCommand.ts to clarify that the Node 22.5+ requirement applies at runtime, not at configuration time.
- Updated the scanError field comment in StatusCommand.ts to reflect that the field is used by both OpenCode and Copilot integrations.
- Updated the module docstring in SqliteHelpers.ts to clarify that the helpers are currently used by OpenCode and Copilot CLI, and that the planned Cursor integration will adopt the same pattern.
- Improved an inline comment on the regex validation in hasNodeSqliteSupport() to explicitly state that the branch is unreachable in practice.
- Updated the design specification document to clarify that CopilotDetector exports two separate functions (isCopilotInstalled and getCopilotDbPath) rather than an aggregate return shape.
- Rewrote CopilotTranscriptReader.ts header comment to clarify that cursor tracks row count (zero-based index into sorted results), not turn_index values, and documented the assumption that Copilot's UNIQUE constraint prevents gaps.

**📁 FILES**
- `cli/src/core/CopilotDetector.ts`
- `cli/src/core/SqliteHelpers.ts`
- `cli/src/core/CopilotTranscriptReader.ts`

</blockquote>
</details>
<details>
<summary><strong>08 · [1de42e7] Complete VS Code Copilot Chat integration with dual storage format support</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Users needed the ability to capture conversations from VS Code's Copilot Chat extension alongside the existing Copilot CLI integration, with both forms discoverable and controllable through a single user-facing toggle.

- **Dual-scan architecture over unified path**: Rather than trying to unify all Copilot session storage under one reader, the design accepts that two different chat entry points use two different storage formats and implements separate sub-readers with a dispatcher. This trades some code duplication for clarity, testability, and resilience — a bad line in one format does not block the other, and each reader can be optimized for its specific format's semantics.
- **Sequential concatenation over preference hierarchy**: Both scans run and results are concatenated without deduplication, allowing sessions from both storage locations to coexist in the output. This is necessary because the two scans represent different entry points (copilotcli-backend models vs. non-copilotcli-backend models in the chat panel).
- **Metadata-based gating over directory structure inference**: Scan A relies on reading and parsing `vscode.metadata.json` to extract the workspace folder path, rather than inferring it from directory structure. This is more explicit and maintainable, though it adds a JSON parse step and requires handling malformed metadata gracefully.
- **Semantic-level cursor using request/line indices**: The readers track progress through transcripts using request array indices or line counts rather than byte offsets. This is more robust against variable line lengths and streaming token accumulation, and aligns with the document-oriented model VS Code uses internally.
- **Lexicographic comparison for ISO 8601 timestamps**: Events.jsonl timestamps are compared as strings rather than parsed to Date objects. ISO 8601 format is lexicographically sortable, so string comparison is safe and avoids parsing overhead while remaining correct across all valid ISO 8601 variants.

**✅ What Was Implemented**

The Copilot Chat integration was implemented as a complete transcript source with support for two distinct storage formats. The detector probes both the VS Code extension's global storage directory and the Copilot CLI agent's `~/.copilot/session-state` directory, returning true if either location exists. The session discoverer uses a two-scan architecture: Scan A discovers sessions from `~/.copilot/session-state/<sid>/events.jsonl` by reading workspace metadata and matching folder paths, while Scan B discovers sessions from the traditional `<wsHash>/chatSessions/<sid>.jsonl` location, with results concatenated to support both entry points. The transcript reader dispatches to format-specific sub-readers based on file path patterns: `readPatchLog()` handles the patch-log format (JSON patches replayed to reconstruct conversations), and `readEventsJsonl()` handles the event stream format (line-delimited JSON with user and assistant message filtering). Both readers support optional timestamp filtering via a `beforeTimestamp` parameter for time-bounded incremental reads, with cursor tracking using request/line indices for resumable consumption. All new code achieved 99%+ test coverage with comprehensive edge-case handling.

**📁 FILES**
- `cli/src/core/CopilotChatDetector.ts`
- `cli/src/core/CopilotChatSessionDiscoverer.ts`
- `cli/src/core/CopilotChatTranscriptReader.ts`

</blockquote>
</details>
<details>
<summary><strong>09 · [1de42e7] Refactor Cursor integration to use shared VS Code workspace locator</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The Cursor integration already solved the problem of locating VS Code workspace directories by project path; duplicating this logic for the new Copilot Chat integration would create maintenance burden and risk code drift.

- **Fail-soft on workspace metadata errors**: When a workspace.json is missing, unparseable, or contains an invalid URI, the scanner logs a debug message and continues to the next entry rather than crashing. This allows the integration to gracefully handle partially corrupted or incomplete workspace storage without blocking discovery of other valid sessions.
- **Parameterized flavor over separate implementations**: The locator accepts an IDE flavor parameter rather than duplicating the logic across both Copilot Chat and Cursor integrations. This eliminates code drift, makes the logic testable in isolation, and makes it trivial to add future VS Code-family forks by passing only a flavor string.

**✅ What Was Implemented**

A shared `VscodeWorkspaceLocator` module was created with platform-aware path resolution functions parameterized by IDE flavor ("Code" or "Cursor"). The module provides `getVscodeUserDataDir()`, `getVscodeWorkspaceStorageDir()`, `findVscodeWorkspaceHash()`, and `normalizePathForMatch()` functions. The `findVscodeWorkspaceHash()` function scans the workspaceStorage directory, parses each workspace's workspace.json, extracts the folder URI, and matches it against the project directory with case-insensitive matching on macOS and Windows. The Cursor integration was refactored to delegate path resolution to this shared locator, eliminating approximately 50 lines of duplicated path-matching and JSON-parsing code.

**📁 FILES**
- `cli/src/core/VscodeWorkspaceLocator.ts`
- `cli/src/core/CursorDetector.ts`

</blockquote>
</details>
<details>
<summary><strong>10 · [1de42e7] Wire Copilot Chat discovery and reading into QueueWorker pipeline</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The QueueWorker needs to discover and include GitHub Copilot Chat sessions alongside other AI assistant sessions when building commit context, and then read transcripts from discovered chat sessions.

- **Shared configuration toggle**: Copilot Chat discovery shares the existing `copilotEnabled` configuration toggle rather than introducing a separate feature flag. This keeps the user-facing settings simple — one toggle controls all GitHub Copilot integrations — while allowing the implementation to distinguish between the CLI source and the VS Code workspaceStorage source internally.
- **Graceful error handling**: The try-catch wrapper with error logging ensures that a single malformed or inaccessible Copilot Chat transcript does not halt the entire queue processing run. This maintains consistency with how other session readers are handled and allows the worker to continue processing remaining sessions even when one fails.

**✅ What Was Implemented**

Conditional discovery logic was added to the `loadSessionTranscripts` function in QueueWorker that checks if Copilot Chat is installed and enabled, then discovers and appends any found chat sessions to the all-sessions list with logging. The copilot-chat branch in the session reading dispatch was wired with a try-catch block that calls `readCopilotChatTranscript`, passing the session's transcript path, optional cursor parameter, and `beforeTimestamp` for time-bounded filtering. Error handling logs failures and skips the session gracefully if reading fails, maintaining consistency with how other session readers are handled.

**📁 FILES**
- `cli/src/hooks/QueueWorker.ts`

</blockquote>
</details>
<details>
<summary><strong>11 · [1de42e7] Extend Copilot auto-enable to detect Chat variant alongside CLI</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The installer previously only auto-enabled Copilot support when the terminal CLI was detected. The system now needs to recognize and enable support for GitHub Copilot Chat as well, since both forms share a single configuration toggle.

**OR semantics for dual detection**: Both Copilot variants (CLI and Chat) share a single `copilotEnabled` toggle. The detection logic uses OR semantics: if either form exists and the user has not explicitly disabled Copilot, the feature turns on. This approach respects explicit user choice while defaulting to enabled for convenience.

**✅ What Was Implemented**

The auto-enable logic in Installer.ts was modified to check both `isCopilotInstalled()` (CLI) and `isCopilotChatInstalled()` (Chat), enabling the shared `copilotEnabled` flag when either form is present and the setting is undefined. The log message was updated to report detection status for both Copilot forms, replacing the CLI-only message with parameterized output showing which variant(s) were found.

**📁 FILES**
- `cli/src/install/Installer.ts`

</blockquote>
</details>
<details>
<summary><strong>12 · [1de42e7] Surface Copilot Chat detection and session discovery in status reporting</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The status reporting system needed to expose whether VS Code's Copilot Chat extension is installed and to discover active chat sessions, similar to how it already handles Copilot and other coding agents.

**On-demand discovery during status checks**: Copilot Chat sessions are discovered on-demand during status checks rather than stored persistently in sessions.json. This approach mirrors the existing pattern for Copilot sessions and avoids maintaining a separate persistent store for a third-party extension's data. The discovery is gated by the `copilotEnabled` configuration flag, ensuring that users who have disabled Copilot integration do not incur the cost of scanning the chat directory.

**✅ What Was Implemented**

The `StatusInfo` type was extended with `copilotChatDetected` and `copilotChatScanError` fields, mirroring the existing pattern for Copilot detection and error reporting. The `Installer.getStatus()` method was modified to call `isCopilotChatInstalled()` and conditionally invoke `scanCopilotChatSessions()` when Copilot Chat is detected and the `copilotEnabled` config flag is true, merging discovered sessions into the active session count and surfacing any scan errors.

**📁 FILES**
- `cli/src/Types.ts`
- `cli/src/install/Installer.ts`

</blockquote>
</details>
<details>
<summary><strong>13 · [1de42e7] Display Copilot CLI and Chat detection breakdown in status output</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The status command was showing aggregate Copilot detection but not distinguishing between CLI and Chat variants, making it unclear which integration method was actually available when only one was detected.

**Sub-line breakdown for clarity**: Displaying the breakdown as indented sub-lines (using ↳ prefix) rather than inline keeps the main Copilot row clean while making the distinction visible without requiring users to parse multiple columns. Aggregating session counts and detection state in the main row ensures the overall integration status is immediately clear, while the sub-line provides diagnostic detail for users troubleshooting which variant is working.

**✅ What Was Implemented**

The StatusCommand was modified to display a sub-line breakdown showing CLI and Chat detection status separately (✓ or ✗) whenever any Copilot variant is detected. The main Copilot row now aggregates both detection states and session counts from both sources. Chat-specific scan errors are surfaced in a second sub-line below the CLI/Chat breakdown when the CLI scan succeeds but the chat scan fails.

**📁 FILES**
- `cli/src/commands/StatusCommand.ts`

</blockquote>
</details>
<details>
<summary><strong>14 · [1de42e7] Show CLI and Chat form breakdown in Copilot status tree</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The Copilot integration status row previously showed only a single detection state, but Copilot now has two independent forms: the terminal CLI and VS Code Chat. Users need visibility into which forms are detected and working.

- **Unified detection logic with per-form visibility**: Rather than showing separate integration rows for CLI and Chat, a single integration row displays both forms' status inline. This keeps the status tree compact while providing the detail users need.
- **Independent error rows for each form**: When a scan fails, a dedicated warning row appears for that form. This allows users to understand which form is broken without losing the overall integration status.

**✅ What Was Implemented**

The StatusTreeProvider was modified to track and display detection status for both Copilot CLI and Chat forms separately. The integration row now shows inline indicators (✓ or ✗) for each form in the tooltip and description text, allowing users to see at a glance which forms are available. The integration is treated as "detected" if either CLI or Chat is present. Scan error handling was separated so that CLI and Chat scan failures surface as independent warning rows, each labeled distinctly ("Copilot CLI database scan failed" vs. "Copilot Chat scan failed").

**📁 FILES**
- `vscode/src/providers/StatusTreeProvider.ts`

</blockquote>
</details>
<details>
<summary><strong>15 · [1de42e7] Add Copilot Chat source support to Summary panel rendering</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The Summary panel needed to recognize and display sessions from the Copilot Chat source alongside existing AI assistant sources like Claude, Copilot, and others.

**Dependent feature of copilot toggle**: The copilot-chat source is treated as a dependent feature of the copilot configuration flag rather than a separate toggle. This keeps the configuration surface simple and prevents users from accidentally enabling one without the other. Placing copilot-chat immediately after copilot in the source ordering maintains logical grouping in the UI, making it clear to users that these are related features.

**✅ What Was Implemented**

A new conditional branch was added to the source-to-label mapping function within SummaryScriptBuilder to return the display label "Copilot Chat" when the source is "copilot-chat". The source ordering array in the session statistics rendering logic was extended to include "copilot-chat" immediately after "copilot", ensuring consistent ordering in the Summary panel UI. The enabled sources filter in SummaryWebviewPanel was updated to include "copilot-chat" whenever the copilot feature is enabled.

**📁 FILES**
- `vscode/src/views/SummaryScriptBuilder.ts`

</blockquote>
</details>
<details>
<summary><strong>16 · [1de42e7] Add copilot-chat as a recognized AI agent source</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The system needed to support a new variant of GitHub Copilot (copilot-chat) alongside the existing copilot integration, ensuring transcripts from this agent are properly recognized and filtered.

**Unified filtering for copilot variants**: Copilot and copilot-chat are filtered together as a single feature family rather than as separate integrations. This approach keeps the configuration surface simple — users do not need separate toggles for each copilot variant — while allowing the system to recognize and track both sources independently.

**✅ What Was Implemented**

The `TranscriptSource` type was updated to include "copilot-chat" as a valid source value alongside existing agents like claude, codex, gemini, opencode, cursor, and copilot. The `filterSessionsByEnabledIntegrations` function was modified to exclude both "copilot" and "copilot-chat" sessions when the `copilotEnabled` configuration is explicitly set to false, treating them as a unified feature family. Test cases were added verifying that copilot-chat sessions are excluded when copilotEnabled is false and confirming that copilot-chat sessions are included during auto-detection.

**📁 FILES**
- `cli/src/Types.ts`
- `cli/src/core/SessionTracker.ts`

</blockquote>
</details>
<details>
<summary><strong>17 · [1de42e7] Boost Copilot Chat test coverage to 98%+ with error-path tests</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The developer added support for reading Copilot Chat transcripts as a new data source, but the new code had gaps in branch coverage. The goal was to raise coverage above 98% for all new code introduced in this feature.

- **Isolated mocking for unreachable branches**: Rather than modify existing test setup to artificially trigger code-less errors (which would break tests relying on real filesystem behavior), separate describe blocks with vi.doMock / vi.doUnmock were created to test the nullish-coalescing fallback paths. This kept the test suite maintainable and prevented cross-test pollution while still achieving full branch coverage.
- **Selective mocking for error precedence**: The error precedence tests use selective mocking of the file system module to inject permission errors at specific paths while allowing other operations to proceed normally. This approach isolates error handling logic without requiring actual file system state manipulation, making tests deterministic and fast.
- **Reuse existing mocking patterns**: The codebase already used `vi.doMock("node:fs/promises", { readdir: path-selective throw, delegate to actual })` to test error paths while preserving real file system behavior for unrelated operations. New tests adopted this pattern rather than introducing a different mocking strategy, reducing cognitive load for future maintainers and keeping test code consistent with established conventions.

**✅ What Was Implemented**

Eleven new test cases were added across six test files to cover previously untested error paths and edge cases: code-less stat errors in CopilotChatDetector, wrapper error handling and symlink-induced stat failures in CopilotChatSessionDiscoverer, null/undefined intermediate paths and non-string response chunks in CopilotChatTranscriptReader, the getCursorWorkspaceStorageDir function in CursorDetector, Chat scan-failed status output in StatusCommand, and getWorkerBusy / dispose lifecycle methods in StatusTreeProvider. Three additional test cases were added to verify error precedence behavior when permission errors occur during directory scanning. Isolated vi.doMock blocks were used to test nullish-coalescing operator branches that real filesystem operations cannot naturally trigger, avoiding pollution of other tests that rely on actual file I/O. Final coverage reached 100% across all four metrics (statements, branches, functions, lines) for all new Copilot Chat modules.

**📁 FILES**
- `cli/src/core/CopilotChatDetector.test.ts`
- `cli/src/core/CopilotChatSessionDiscoverer.test.ts`
- `cli/src/core/CopilotChatTranscriptReader.test.ts`

</blockquote>
</details>
<details>
<summary><strong>18 · [1de42e7] Mock CopilotChat modules in PostCommitHook tests</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Test suite was timing out with 11 failures in PostCommitHook tests. The root cause was that newly added CopilotChat discovery functions were not mocked, causing the QueueWorker's chat discovery logic to attempt real filesystem operations that hung indefinitely.

**Mocks in test files rather than shared utility**: The mocks were added directly to the test files rather than extracted to a shared test utility because the existing mock pattern in these files already groups all module mocks together at the top of each test file. This maintains consistency with the established test structure and makes the mocks immediately visible to anyone reading the test.

**✅ What Was Implemented**

Mock definitions were added for five CopilotChat-related modules in PostCommitHook.helpers.test.ts and PostCommitHook.test.ts: CopilotChatDetector, CopilotChatSessionDiscoverer, and CopilotChatTranscriptReader, each returning empty or false values to prevent filesystem access during test execution. Additional mocks for CopilotDetector and CopilotSessionDiscoverer were added to PostCommitHook.helpers.test.ts to ensure complete coverage of all Copilot-related discovery paths.

**📁 FILES**
- `cli/src/hooks/PostCommitHook.helpers.test.ts`
- `cli/src/hooks/PostCommitHook.test.ts`

</blockquote>
</details>
<details>
<summary><strong>19 · [1de42e7] Clarify runtime compatibility gating and improve logging context</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The existing comment explaining why Cursor detection is gated on Node version support was vague about the user-facing consequence and the logging rationale.

**Info-level logging for expected incompatibility**: Logging at info level rather than warning or error is appropriate because the condition is not a failure — it is a runtime incompatibility that is expected and normal on certain hosts (e.g., VS Code extension Node 18 environments). The expanded comments make this intent explicit so future maintainers understand that the gating is a feature, not a workaround.

**✅ What Was Implemented**

The docstring for `isCursorInstalled()` was expanded to explain that the function reports "not installed" on incompatible runtimes (Node <22.5) rather than "detected but 0 sessions", which would be misleading since a scan would always fail on those runtimes. An inline comment was added before the log statement clarifying that the info-level log is intentional so operators can correlate the absence of Cursor from status output with the actual runtime version constraint.

**📁 FILES**
- `cli/src/core/CursorDetector.ts`

</blockquote>
</details>
<details>
<summary><strong>20 · [1de42e7] Log skipped chat session files when stat fails</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A code review identified that file stat errors during chat session discovery were silently swallowed, making it difficult to diagnose why sessions were unexpectedly missing from the scan results.

**Debug-level logging for observability**: Debug-level logging was chosen over silent failure to preserve observability without cluttering normal operation. Users running the CLI in standard mode will not see these messages, but developers troubleshooting session discovery issues can enable debug output to understand what went wrong.

**✅ What Was Implemented**

Debug logging was added to the catch block in CopilotChatSessionDiscoverer.ts when the stat() call fails on a discovered file. The log message includes the file entry name and the underlying error message, allowing developers to see which files were skipped and why.

**📁 FILES**
- `cli/src/core/CopilotChatSessionDiscoverer.ts`

</blockquote>
</details>
<details>
<summary><strong>21 · [1de42e7] Clarify status display comment for Copilot scan errors</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A code review noted that the inline comment describing how Copilot scan errors are displayed in the status output was ambiguous and could mislead future maintainers about the rendering behavior.

**Explicit rendering language**: Replacing "takes precedence" with "renders on" removes the implication of priority or suppression, clarifying that the behavior is purely about visual layout rather than error filtering. This prevents future developers from misinterpreting the code and potentially introducing bugs when modifying error handling logic.

**✅ What Was Implemented**

The comment in StatusCommand.ts was updated from "CLI scan error takes precedence; chat scan error surfaced via sub-line below" to "CLI scan error renders on the main row; Chat scan error renders as a sub-line below." The change makes explicit that both errors are displayed, just in different visual locations.

**📁 FILES**
- `cli/src/commands/StatusCommand.ts`

</blockquote>
</details>
<details>
<summary><strong>22 · [1de42e7] Align API naming between spec and implementation for session state directory</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The design specification and implementation had inconsistent function names for retrieving the Copilot CLI session state directory, creating a mismatch between documented and actual APIs.

**Naming consistency between spec and code**: Naming consistency between spec and code prevents future confusion and makes the API contract explicit. Distinguishing the two directory functions by their purpose (CLI backend vs. extension storage) in the documentation clarifies which function should be used in different contexts and prevents accidental mixing of concerns.

**✅ What Was Implemented**

The design specification was updated to rename `getCopilotChatSessionStateDir` to `getCopilotCliSessionStateDir` and added clarifying comments distinguishing it from the existing `getCopilotChatStorageDir` function. The spec now documents that `getCopilotCliSessionStateDir` points to `~/.copilot/session-state` (the new Copilot CLI agent backend root) while `getCopilotChatStorageDir` points to the VS Code extension's global storage directory.

**📁 FILES**
- `docs/superpowers/specs/2026-05-07-copilot-chat-events-jsonl-redesign-design.md`

</blockquote>
</details>
<details>
<summary><strong>23 · [1de42e7] Remove stale task-reference comments from reader and design documentation</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Code review identified outdated comments and documentation that referenced future work items no longer applicable after task consolidation.

**Stale comments create confusion**: Stale comments create confusion for future maintainers who may spend time looking for work that was already completed or merged. Removing task-specific qualifiers and updating documentation to reflect the current state of the codebase improves clarity without changing any behavior.

**✅ What Was Implemented**

A comment in CopilotChatTranscriptReader.ts that mentioned `readEventsJsonl` would be "added in a later task" was removed since the function is now present. The format detection table in the design specification was revised to use current terminology and remove references to legacy format detection logic, replacing outdated scenario descriptions with accurate descriptions of the two probe paths (Copilot CLI agent backend and Copilot Chat extension).

**📁 FILES**
- `cli/src/core/CopilotChatTranscriptReader.ts`
- `docs/superpowers/specs/2026-05-07-copilot-chat-events-jsonl-redesign-design.md`

</blockquote>
</details>

> ⚠️ 2 more topics omitted due to GitHub PR body size limit.

---

*Generated by Jolli Memory · May 7, 2026 at 1:51 PM*

### DIFF
--- a/cli/src/Types.ts
+++ b/cli/src/Types.ts
@@ -4,10 +4,11 @@
  * Central type definitions for all modules in the Jolli Memory tool.
  */
 
+import type { CopilotChatScanError } from "./core/CopilotChatTranscriptReader.js";
 import type { SqliteScanError } from "./core/SqliteHelpers.js";
 
 /** Which AI coding agent produced the transcript */
-export type TranscriptSource = "claude" | "codex" | "gemini" | "opencode" | "cursor" | "copilot";
+export type TranscriptSource = "claude" | "codex" | "gemini" | "opencode" | "cursor" | "copilot" | "copilot-chat";
 
 /** Metadata about an AI coding session, saved by the Stop hook (Claude) or discovered on-demand (Codex) */
 export interface SessionInfo {
@@ -614,6 +615,10 @@ export interface StatusInfo {
 	readonly openCodeScanError?: SqliteScanError;
 	/** Copilot DB scan failed with a real (non-ENOENT) error. Same UI semantics as openCodeScanError. */
 	readonly copilotScanError?: SqliteScanError;
+	/** Whether vscode's Copilot Chat globalStorage dir was detected */
+	readonly copilotChatDetected?: boolean;
+	/** Copilot Chat scan failed with a real (non-ENOENT) error: parse / fs / schema. */
+	readonly copilotChatScanError?: CopilotChatScanError;
 }
 
 /**
--- a/cli/src/Types.ts
+++ b/cli/src/Types.ts
@@ -4,8 +4,10 @@
  * Central type definitions for all modules in the Jolli Memory tool.
  */
 
+import type { SqliteScanError } from "./core/SqliteHelpers.js";
+
 /** Which AI coding agent produced the transcript */
-export type TranscriptSource = "claude" | "codex" | "gemini" | "opencode" | "cursor";
+export type TranscriptSource = "claude" | "codex" | "gemini" | "opencode" | "cursor" | "copilot";
 
 /** Metadata about an AI coding session, saved by the Stop hook (Claude) or discovered on-demand (Codex) */
 export interface SessionInfo {
@@ -488,6 +490,8 @@ export interface JolliMemoryConfig {
 	readonly openCodeEnabled?: boolean;
 	/** Enable Cursor Composer session discovery at post-commit time (default: auto-detect) */
 	readonly cursorEnabled?: boolean;
+	/** Enable GitHub Copilot CLI session discovery at post-commit time (default: auto-detect) */
+	readonly copilotEnabled?: boolean;
 	/** Global minimum log level written to debug.log (default: "info") */
 	readonly logLevel?: LogLevel;
 	/** Per-module log level overrides (e.g. { "GitOps": "debug" }) */
@@ -569,10 +573,11 @@ export interface StatusInfo {
 	 * schema drift, or permission denied. UI surfaces this adjacent to the Cursor
 	 * row instead of silently rendering "0 sessions".
 	 */
-	readonly cursorScanError?: {
-		readonly kind: "corrupt" | "locked" | "permission" | "schema" | "unknown";
-		readonly message: string;
-	};
+	readonly cursorScanError?: SqliteScanError;
+	/** Whether Copilot CLI's session DB (~/.copilot/session-store.db) was detected */
+	readonly copilotDetected?: boolean;
+	/** Whether Copilot CLI session discovery is enabled in config (undefined = auto-detect) */
+	readonly copilotEnabled?: boolean;
 	/** Directory path for global config (~/.jolli/jollimemory) */
 	readonly globalConfigDir?: string;
 	/** Path to the worktree state directory */
@@ -606,10 +611,9 @@ export interface StatusInfo {
 	 * a warning adjacent to the OpenCode integration row rather than rendering
 	 * "0 sessions" (which is indistinguishable from "no OpenCode activity").
 	 */
-	readonly openCodeScanError?: {
-		readonly kind: "corrupt" | "locked" | "permission" | "schema" | "unknown";
-		readonly message: string;
-	};
+	readonly openCodeScanError?: SqliteScanError;
+	/** Copilot DB scan failed with a real (non-ENOENT) error. Same UI semantics as openCodeScanError. */
+	readonly copilotScanError?: SqliteScanError;
 }
 
 /**
--- a/cli/src/commands/ConfigureCommand.test.ts
+++ b/cli/src/commands/ConfigureCommand.test.ts
@@ -1,0 +1,103 @@
+/**
+ * Tests for ConfigureCommand — `jolli configure` copilotEnabled key.
+ *
+ * Covers:
+ *   - copilotEnabled accepted as a settable boolean key
+ *   - copilotEnabled appears in --list-keys output
+ */
+
+import { Command } from "commander";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { JolliMemoryConfig } from "../Types.js";
+
+// ─── Hoist mocks ─────────────────────────────────────────────────────────────
+
+const { mockLoadConfig, mockSaveConfig } = vi.hoisted(() => ({
+	mockLoadConfig: vi.fn(),
+	mockSaveConfig: vi.fn(),
+}));
+
+vi.mock("../core/SessionTracker.js", () => ({
+	getGlobalConfigDir: vi.fn().mockReturnValue("/mock/global/config"),
+	loadConfig: mockLoadConfig,
+	saveConfig: mockSaveConfig,
+}));
+
+vi.mock("../core/JolliApiUtils.js", () => ({
+	validateJolliApiKey: vi.fn(),
+}));
+
+// ─── Helpers ─────────────────────────────────────────────────────────────────
+
+/** Captured config built up from saveConfig calls. */
+let savedConfig: Partial<JolliMemoryConfig> = {};
+
+/**
+ * Runs `jolli configure` with the given args and returns stdout output.
+ * Side effects on savedConfig accumulate across calls (reset in beforeEach).
+ */
+async function runConfigure(args: string[]): Promise<string> {
+	const { registerConfigureCommand } = await import("./ConfigureCommand.js");
+	const program = new Command();
+	program.exitOverride();
+	registerConfigureCommand(program);
+
+	const lines: string[] = [];
+	const spy = vi.spyOn(console, "log").mockImplementation((...a: unknown[]) => {
+		lines.push(a.map(String).join(" "));
+	});
+	try {
+		await program.parseAsync(["configure", ...args], { from: "user" });
+	} finally {
+		spy.mockRestore();
+	}
+	return lines.join("\n");
+}
+
+/**
+ * Returns a loadable config object reflecting whatever saveConfig was last
+ * called with.  Mimics the real SessionTracker round-trip.
+ */
+async function loadConfig(): Promise<Partial<JolliMemoryConfig>> {
+	return savedConfig;
+}
+
+/**
+ * Runs `jolli configure --list-keys` and returns the captured output.
+ */
+async function runConfigureHelp(): Promise<string> {
+	return runConfigure(["--list-keys"]);
+}
+
+// ─── Tests ────────────────────────────────────────────────────────────────────
+
+describe("ConfigureCommand — copilotEnabled key", () => {
+	beforeEach(() => {
+		vi.clearAllMocks();
+		savedConfig = {};
+		mockLoadConfig.mockResolvedValue({});
+		// Wire saveConfig to capture what was saved
+		mockSaveConfig.mockImplementation(async (update: Partial<JolliMemoryConfig>) => {
+			savedConfig = { ...savedConfig, ...update };
+		});
+	});
+
+	afterEach(() => {
+		vi.clearAllMocks();
+	});
+
+	it("accepts copilotEnabled as a boolean key", async () => {
+		await runConfigure(["--set", "copilotEnabled=true"]);
+		expect(mockSaveConfig).toHaveBeenCalledWith(expect.objectContaining({ copilotEnabled: true }));
+		expect((await loadConfig()).copilotEnabled).toBe(true);
+
+		await runConfigure(["--set", "copilotEnabled=false"]);
+		expect(mockSaveConfig).toHaveBeenCalledWith(expect.objectContaining({ copilotEnabled: false }));
+		expect((await loadConfig()).copilotEnabled).toBe(false);
+	});
+
+	it("lists copilotEnabled in help/description output", async () => {
+		const help = await runConfigureHelp();
+		expect(help).toContain("copilotEnabled");
+	});
+});
--- a/cli/src/commands/ConfigureCommand.ts
+++ b/cli/src/commands/ConfigureCommand.ts
@@ -32,6 +32,7 @@ const VALID_CONFIG_KEYS = [
 	"claudeEnabled",
 	"openCodeEnabled",
 	"cursorEnabled",
+	"copilotEnabled",
 	"logLevel",
 	"excludePatterns",
 ] as const satisfies ReadonlyArray<keyof JolliMemoryConfig>;
@@ -70,7 +71,8 @@ function coerceConfigValue(key: ConfigKey, raw: string): string | number | boole
 		key === "geminiEnabled" ||
 		key === "claudeEnabled" ||
 		key === "openCodeEnabled" ||
-		key === "cursorEnabled"
+		key === "cursorEnabled" ||
+		key === "copilotEnabled"
 	) {
 		const lower = raw.toLowerCase();
 		if (lower === "true" || lower === "1" || lower === "yes") return true;
@@ -114,6 +116,11 @@ const CONFIG_KEY_INFO: ReadonlyArray<{ key: ConfigKey; type: string; description
 		key: "cursorEnabled",
 		type: "boolean",
 		description: "Enable Cursor Composer session discovery (true/false; requires Node 22.5+ at runtime)",
+	},
+	{
+		key: "copilotEnabled",
+		type: "boolean",
+		description: "Enable Copilot CLI session discovery (true/false; requires Node 22.5+ at runtime)",
 	},
 	{ key: "logLevel", type: "enum", description: "Log level: debug | info | warn | error" },
 	{ key: "excludePatterns", type: "string[]", description: "Glob patterns for file exclusion (comma-separated)" },
--- a/cli/src/commands/StatusCommand.test.ts
+++ b/cli/src/commands/StatusCommand.test.ts
@@ -127,4 +127,32 @@ describe("StatusCommand — Copilot integration row", () => {
 		expect(out).toContain("Copilot");
 		expect(out).toContain("detected but disabled");
 	});
+
+	it("Copilot row shows CLI/Chat breakdown when only chat is detected", async () => {
+		const out = await renderStatus({
+			...baseStatus,
+			copilotDetected: false,
+			copilotChatDetected: true,
+			copilotEnabled: true,
+			sessionsBySource: { "copilot-chat": 2 },
+		});
+		expect(out).toContain("CLI: ✗");
+		expect(out).toContain("Chat: ✓");
+		expect(out).toMatch(/2\s*sessions/);
+	});
+
+	it("emits a Chat scan-failed sub-line when copilotChatScanError is set", async () => {
+		// CLI scan error renders on the main row; Chat error is structurally
+		// different — it surfaces as an indented "↳ Chat scan failed" sub-line
+		// underneath the CLI/Chat breakdown so users can tell which form failed.
+		const out = await renderStatus({
+			...baseStatus,
+			copilotDetected: true,
+			copilotChatDetected: true,
+			copilotEnabled: true,
+			copilotChatScanError: { kind: "parse", message: "bad jsonl event at line 7" },
+		});
+		expect(out).toContain("Chat scan failed (parse)");
+		expect(out).toContain("bad jsonl event at line 7");
+	});
 });
--- a/cli/src/commands/StatusCommand.test.ts
+++ b/cli/src/commands/StatusCommand.test.ts
@@ -1,0 +1,130 @@
+/**
+ * Tests for StatusCommand — `jolli status` integration rows.
+ *
+ * Covers the per-integration breakdown output, focused on the Copilot row.
+ */
+
+import { Command } from "commander";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { StatusInfo } from "../Types.js";
+
+// ─── Hoist mocks ─────────────────────────────────────────────────────────────
+
+const { mockGetStatus, mockLoadConfigFromDir, mockLoadAuthToken } = vi.hoisted(() => ({
+	mockGetStatus: vi.fn(),
+	mockLoadConfigFromDir: vi.fn(),
+	mockLoadAuthToken: vi.fn(),
+}));
+
+vi.mock("../install/Installer.js", () => ({
+	getStatus: mockGetStatus,
+}));
+
+vi.mock("../core/SessionTracker.js", () => ({
+	getGlobalConfigDir: vi.fn().mockReturnValue("/mock/global/config"),
+	loadConfigFromDir: mockLoadConfigFromDir,
+}));
+
+vi.mock("../auth/AuthConfig.js", () => ({
+	loadAuthToken: mockLoadAuthToken,
+}));
+
+vi.mock("../core/JolliApiUtils.js", () => ({
+	parseJolliApiKey: vi.fn().mockReturnValue(null),
+}));
+
+// ─── Helpers ─────────────────────────────────────────────────────────────────
+
+/**
+ * Minimal StatusInfo satisfying the required fields — used as the spread base
+ * in individual test cases.
+ */
+const baseStatus: StatusInfo = {
+	enabled: true,
+	claudeHookInstalled: false,
+	gitHookInstalled: false,
+	geminiHookInstalled: false,
+	activeSessions: 0,
+	mostRecentSession: null,
+	summaryCount: 0,
+	orphanBranch: "jollimemory/summaries/v3",
+	sessionsBySource: {},
+};
+
+/**
+ * Renders `jolli status` output to a string by capturing console.log calls.
+ * Configures mockGetStatus with the given status before parsing.
+ */
+async function renderStatus(status: StatusInfo): Promise<string> {
+	mockGetStatus.mockResolvedValueOnce(status);
+
+	const { registerStatusCommand } = await import("./StatusCommand.js");
+	const program = new Command();
+	program.exitOverride();
+	registerStatusCommand(program);
+
+	const lines: string[] = [];
+	const spy = vi.spyOn(console, "log").mockImplementation((...args: unknown[]) => {
+		lines.push(args.map(String).join(" "));
+	});
+	try {
+		await program.parseAsync(["status"], { from: "user" });
+	} finally {
+		spy.mockRestore();
+	}
+	return lines.join("\n");
+}
+
+// ─── Tests ────────────────────────────────────────────────────────────────────
+
+describe("StatusCommand — Copilot integration row", () => {
+	beforeEach(() => {
+		vi.clearAllMocks();
+		mockLoadConfigFromDir.mockResolvedValue({});
+		mockLoadAuthToken.mockResolvedValue(null);
+	});
+
+	afterEach(() => {
+		vi.clearAllMocks();
+	});
+
+	it("does not render Copilot row when copilot not detected", async () => {
+		const out = await renderStatus({
+			...baseStatus,
+			// copilotDetected unset (undefined) -> default state
+		});
+		expect(out).not.toContain("Copilot:");
+	});
+
+	it("renders Copilot Integration row when detected", async () => {
+		const out = await renderStatus({
+			...baseStatus,
+			copilotDetected: true,
+			copilotEnabled: true,
+			sessionsBySource: { copilot: 3 },
+		});
+		expect(out).toContain("Copilot");
+		expect(out).toContain("3");
+	});
+
+	it("renders Copilot row as unavailable on scan error", async () => {
+		const out = await renderStatus({
+			...baseStatus,
+			copilotDetected: true,
+			copilotScanError: { kind: "locked", message: "database is locked" },
+		});
+		expect(out).toContain("Copilot");
+		expect(out).toContain("unavailable");
+		expect(out).toContain("locked");
+	});
+
+	it("renders Copilot row as disabled when detected but copilotEnabled is false", async () => {
+		const out = await renderStatus({
+			...baseStatus,
+			copilotDetected: true,
+			copilotEnabled: false,
+		});
+		expect(out).toContain("Copilot");
+		expect(out).toContain("detected but disabled");
+	});
+});
--- a/cli/src/commands/StatusCommand.ts
+++ b/cli/src/commands/StatusCommand.ts
@@ -160,11 +160,12 @@ export function registerStatusCommand(program: Command): void {
 				],
 				[
 					"Copilot:",
-					status.copilotDetected,
+					(status.copilotDetected ?? false) || (status.copilotChatDetected ?? false),
 					{
 						enabled: status.copilotEnabled !== false,
 						hookInstalled: undefined,
-						sessionCount: counts.copilot,
+						sessionCount: (counts.copilot ?? 0) + (counts["copilot-chat"] ?? 0),
+						// CLI scan error renders on the main row; Chat scan error renders as a sub-line below.
 						scanError: status.copilotScanError,
 					},
 				],
@@ -172,6 +173,17 @@ export function registerStatusCommand(program: Command): void {
 			for (const [label, detected, inputs] of integrationRows) {
 				if (!detected) continue;
 				console.log(`  ${label.padEnd(18)}${describeIntegrationStatus(inputs)}`);
+			}
+			const anyCopilotDetected = (status.copilotDetected ?? false) || (status.copilotChatDetected ?? false);
+			if (anyCopilotDetected) {
+				const cliMark = status.copilotDetected ? "✓" : "✗";
+				const chatMark = status.copilotChatDetected ? "✓" : "✗";
+				console.log(`  ${"".padEnd(18)}↳ CLI: ${cliMark}, Chat: ${chatMark}`);
+				if (status.copilotChatScanError) {
+					console.log(
+						`  ${"".padEnd(18)}↳ Chat scan failed (${status.copilotChatScanError.kind}): ${status.copilotChatScanError.message}`,
+					);
+				}
 			}
 
 			console.log(`  Stored memories:  ${status.summaryCount}`);
--- a/cli/src/commands/StatusCommand.ts
+++ b/cli/src/commands/StatusCommand.ts
@@ -22,7 +22,7 @@ interface IntegrationStatusInputs {
 	/** undefined = this integration has no hook concept (Codex, OpenCode). */
 	readonly hookInstalled: boolean | undefined;
 	readonly sessionCount: number | undefined;
-	/** OpenCode-only: DB existed but scan failed (corrupt/locked/schema/etc). */
+	/** DB existed but scan failed (corrupt/locked/schema/etc). Used by OpenCode and Copilot integrations. */
 	readonly scanError?: StatusInfo["openCodeScanError"];
 }
 
@@ -156,6 +156,16 @@ export function registerStatusCommand(program: Command): void {
 						hookInstalled: undefined,
 						sessionCount: counts.cursor,
 						scanError: status.cursorScanError,
+					},
+				],
+				[
+					"Copilot:",
+					status.copilotDetected,
+					{
+						enabled: status.copilotEnabled !== false,
+						hookInstalled: undefined,
+						sessionCount: counts.copilot,
+						scanError: status.copilotScanError,
 					},
 				],
 			];
--- a/cli/src/core/CopilotChatDetector.test.ts
+++ b/cli/src/core/CopilotChatDetector.test.ts
@@ -1,0 +1,111 @@
+import { stat } from "node:fs/promises";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("node:fs/promises", async () => {
+	const actual = await vi.importActual<typeof import("node:fs/promises")>("node:fs/promises");
+	return { ...actual, stat: vi.fn() };
+});
+
+const { mockHomedir, mockPlatform } = vi.hoisted(() => ({
+	mockHomedir: vi.fn().mockReturnValue("/Users/test"),
+	mockPlatform: vi.fn().mockReturnValue("darwin"),
+}));
+
+vi.mock("node:os", async (importOriginal) => {
+	const original = await importOriginal<typeof import("node:os")>();
+	return { ...original, homedir: mockHomedir, platform: mockPlatform };
+});
+
+describe("CopilotChatDetector", () => {
+	beforeEach(() => {
+		mockHomedir.mockReturnValue("/Users/test");
+		mockPlatform.mockReturnValue("darwin");
+		vi.mocked(stat).mockReset();
+	});
+
+	afterEach(() => {
+		vi.restoreAllMocks();
+	});
+
+	it("getCopilotChatStorageDir returns vscode globalStorage path on darwin", async () => {
+		const { getCopilotChatStorageDir } = await import("./CopilotChatDetector.js");
+		expect(getCopilotChatStorageDir()).toBe(
+			join(
+				"/Users/test",
+				"Library",
+				"Application Support",
+				"Code",
+				"User",
+				"globalStorage",
+				"github.copilot-chat",
+			),
+		);
+	});
+
+	it("getCopilotCliSessionStateDir returns ~/.copilot/session-state path", async () => {
+		const { getCopilotCliSessionStateDir } = await import("./CopilotChatDetector.js");
+		expect(getCopilotCliSessionStateDir()).toBe(join("/Users/test", ".copilot", "session-state"));
+	});
+
+	it("isCopilotChatInstalled returns true when ONLY globalStorage exists", async () => {
+		// Use a path-segment regex so the includes() check works on both `/` and `\` separators.
+		const globalStorageMarker = /[\\/]globalStorage[\\/]github\.copilot-chat/;
+		vi.mocked(stat).mockImplementation(async (path) => {
+			if (globalStorageMarker.test(String(path))) {
+				return { isDirectory: () => true } as Awaited<ReturnType<typeof stat>>;
+			}
+			throw Object.assign(new Error("no dir"), { code: "ENOENT" });
+		});
+		const { isCopilotChatInstalled } = await import("./CopilotChatDetector.js");
+		await expect(isCopilotChatInstalled()).resolves.toBe(true);
+	});
+
+	it("isCopilotChatInstalled returns true when ONLY ~/.copilot/session-state exists", async () => {
+		const sessionStateMarker = /[\\/]\.copilot[\\/]session-state/;
+		vi.mocked(stat).mockImplementation(async (path) => {
+			if (sessionStateMarker.test(String(path))) {
+				return { isDirectory: () => true } as Awaited<ReturnType<typeof stat>>;
+			}
+			throw Object.assign(new Error("no dir"), { code: "ENOENT" });
+		});
+		const { isCopilotChatInstalled } = await import("./CopilotChatDetector.js");
+		await expect(isCopilotChatInstalled()).resolves.toBe(true);
+	});
+
+	it("isCopilotChatInstalled returns true when BOTH paths exist", async () => {
+		vi.mocked(stat).mockResolvedValue({ isDirectory: () => true } as Awaited<ReturnType<typeof stat>>);
+		const { isCopilotChatInstalled } = await import("./CopilotChatDetector.js");
+		await expect(isCopilotChatInstalled()).resolves.toBe(true);
+	});
+
+	it("isCopilotChatInstalled returns false when both paths missing", async () => {
+		vi.mocked(stat).mockRejectedValue(Object.assign(new Error("no dir"), { code: "ENOENT" }));
+		const { isCopilotChatInstalled } = await import("./CopilotChatDetector.js");
+		await expect(isCopilotChatInstalled()).resolves.toBe(false);
+	});
+
+	it("isCopilotChatInstalled returns false when path exists but is not a directory", async () => {
+		vi.mocked(stat).mockResolvedValue({ isDirectory: () => false } as Awaited<ReturnType<typeof stat>>);
+		const { isCopilotChatInstalled } = await import("./CopilotChatDetector.js");
+		await expect(isCopilotChatInstalled()).resolves.toBe(false);
+	});
+
+	it("isCopilotChatInstalled returns false on unexpected stat error and warn-logs", async () => {
+		vi.mocked(stat).mockRejectedValue(Object.assign(new Error("perm denied"), { code: "EACCES" }));
+		const { isCopilotChatInstalled } = await import("./CopilotChatDetector.js");
+		await expect(isCopilotChatInstalled()).resolves.toBe(false);
+	});
+
+	it("isCopilotChatInstalled warn-logs with 'unknown' when stat error has no code property", async () => {
+		// Exercises the `code ?? "unknown"` nullish-coalescing fallback in existsAsDir.
+		vi.mocked(stat).mockRejectedValue(new Error("weird error without errno code"));
+		const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+		const { isCopilotChatInstalled } = await import("./CopilotChatDetector.js");
+		await expect(isCopilotChatInstalled()).resolves.toBe(false);
+		expect(warnSpy).toHaveBeenCalled();
+		const formatted = warnSpy.mock.calls.map((c) => String(c[0])).join("\n");
+		expect(formatted).toMatch(/unknown/);
+		warnSpy.mockRestore();
+	});
+});
--- a/cli/src/core/CopilotChatDetector.ts
+++ b/cli/src/core/CopilotChatDetector.ts
@@ -1,0 +1,60 @@
+/**
+ * VS Code Copilot Chat detector.
+ *
+ * Returns true when EITHER of the two Copilot Chat data roots exist:
+ *   - <userDataDir>/User/globalStorage/github.copilot-chat (Copilot Chat extension installed)
+ *   - ~/.copilot/session-state                              (Copilot CLI agent backend on disk)
+ *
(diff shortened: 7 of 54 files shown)


sanshizhang-jolli:
### Test suite has 7 Windows-specific failures

The new Copilot test files hardcode POSIX paths in fixtures and assertions, which fail on Windows because `path.resolve` / `path.join` produce backslash-separated, drive-prefixed paths there.

**Failures observed locally on Windows + Node 22.5+:**

- `src/core/CopilotDetector.test.ts:32` asserts:
  ```ts
  expect(getCopilotDbPath()).toBe(\"/Users/test/.copilot/session-store.db\")
  ```
  On Windows, `join(\"/Users/test\", \".copilot\", \"session-store.db\")` returns `\Users\test\.copilot\session-store.db`.

- `src/core/CopilotSessionDiscoverer.test.ts` seeds DB rows with \`cwd = \"/p\"\`, \`cwd = \"/Users/x/project\"\`, etc. The discoverer's \`normalizeCwd(p)\` calls \`path.resolve(p)\`, which on Windows turns \`/p\` into \`E:\p\` (current drive). The SQL \`cwd = :cwd\` then never matches, and 6 tests fail with empty result sets.

**Repro:**
\`\`\`
$ npx vitest run src/core/CopilotDetector.test.ts src/core/CopilotSessionDiscoverer.test.ts
 Tests  7 failed | 7 passed (14)
\`\`\`

The spec explicitly excludes Windows CI testing, so this won't surface in the build pipeline — but any contributor running \`npm test\` on a Windows dev box hits a wall of red. The production code itself is platform-correct (\`path.join\` + \`homedir()\` + \`process.platform\`-gated case comparison); only the tests are non-portable.

**Suggested fixes (test-only, no production change):**

- \`CopilotDetector.test.ts:32\` → assert against \`join(homedir(), \".copilot\", \"session-store.db\")\` directly, or use a regex like \`/[\\/]\.copilot[\\/]session-store\.db\$/\`.
- \`CopilotSessionDiscoverer.test.ts\` fixtures → seed \`cwd\` with \`path.resolve(...)\` of a platform-appropriate string (e.g. \`process.platform === \"win32\" ? \"C:\\p\" : \"/p\"\`), so the seeded value matches what \`normalizeCwd\` produces at query time. Easiest variant: use the test's own \`mkdtemp\` directory as both the fixture cwd and the projectDir argument.

sanshizhang-jolli:
### Asymmetric error handling between OpenCode and Copilot in `QueueWorker.readAllTranscripts`

In `cli/src/hooks/QueueWorker.ts` (~line 1481), the two SQLite-backed sources are wrapped differently:

\`\`\`ts
} else if (source === \"opencode\") {
    result = await readOpenCodeTranscript(session.transcriptPath, cursor, beforeTimestamp);
} else if (source === \"copilot\") {
    try {
        result = await readCopilotTranscript(session.transcriptPath, cursor, beforeTimestamp);
    } catch (error: unknown) {
        log.error(\"Skipping Copilot session %s: %s\", session.sessionId, (error as Error).message);
        continue;
    }
}
\`\`\`

Both sources read SQLite databases produced by external agents and share the same failure modes (locked / corrupt / schema drift / DB disappeared between scan and read). But:

- **OpenCode** read failure → terminates the entire \`readAllTranscripts\` call, abandoning every other session in the batch.
- **Copilot** read failure → isolated to one session, pipeline continues.

This is the kind of subtle inconsistency that surfaces only in production: someone hits a transiently locked OpenCode DB and loses the whole post-commit summary instead of just one session's contribution.

**Two ways to align:**

1. **Preferred** — wrap \`readOpenCodeTranscript\` in the same try/catch + \`continue\` pattern. Per-session isolation is the safer default for any SQLite-backed source, since a single corrupted/locked session shouldn't prevent the rest of the batch from being summarized. This matches the rationale that drove the try/catch addition for Copilot in the first place.
2. Drop the try/catch around \`readCopilotTranscript\` and let both readers terminate the call on failure. Lower effort, but gives up the resilience benefit and contradicts the per-session isolation argument made in the PR description.

Recommend option 1 — it keeps the safer behavior the Copilot integration introduced, and applies it uniformly.

sanshizhang-jolli (dismissed):

sanshizhang-jolli (approved):

PR #92, opened 2026-05-09
Split worker and orphan-write locks into separate directories with PID-ownership checks

sanshizhang-jolli:
## Commits in this PR (3)

1. Split worker and orphan-write locks to fix orphaned queue entries (`0291c58`)
2. Tighten orphan-write lock: full LMW, longer worker timeout, sync VS Code watchers (`d24f1de`)
3. Split worker and orphan-write locks into separate directories with PID-ownership checks (`fd15246`)

### Commit 1 of 3: Split worker and orphan-write locks to fix orphaned queue entries (`0291c58`)

A long-standing race condition in the queue processing pipeline was resolved. Previously, a single shared lock file served two unrelated purposes: signaling that a worker was actively draining the queue, and protecting brief writes to the summary branch. Whenever a background writer held that lock — even for milliseconds — the post-rewrite hook saw it as evidence of a running worker and skipped spawning a new one, leaving the queued session permanently unprocessed.

The fix introduces two separate lock files with distinct acquisition strategies matched to their actual contention patterns. The worker lock is now the sole signal that a drain is in progress, and background writers use an entirely separate lock that the hook never consults. All write operations on the summary branch now acquire the background lock around their full read-modify-write cycle, closing a secondary window where concurrent writers could overwrite each other's index updates. The old unified lock and all its associated functions were removed from the codebase entirely, eliminating the possibility of future callers accidentally reintroducing the same confusion.

### Commit 2 of 3: Tighten orphan-write lock: full LMW, longer worker timeout, sync VS Code watchers (`d24f1de`)

Four related bugs in the orphan-branch write coordination were fixed in this batch of changes. The most significant was a lost-update race: a background scan that resolves commit aliases was reading the shared index before acquiring the write lock, then writing back a stale copy that silently dropped any entries a concurrent worker had added in the meantime. The fix splits the scan into a preflight phase (expensive git lookups, no lock held) and a locked phase that re-reads the freshly persisted index before merging, preserving all concurrent writes.

The write lock timeout was also extended from 5 seconds to 30 seconds, with detailed documentation explaining the reasoning. The previous timeout was too short to survive a large-repository one-time migration running at the same moment as a commit. The failure mode on timeout — discarding the queue entry rather than retrying — was deliberately kept, because the transcript reading cursor advances before the write attempt, making a retry produce a summary over the wrong content.

Two additional gaps were closed: the one-time v1-to-v3 migration path was writing directly to the orphan branch without holding the write lock at all, and the VS Code extension was still watching the old combined lock file path after the lock was split into two separate files. The extension now correctly tracks only the worker lock for button-guard purposes, and the migration path acquires the write lock around both of its raw branch writes.

### Commit 3 of 3: Split worker and orphan-write locks into separate directories with PID-ownership checks (`fd15246`)

Three concurrency bugs in the locking system were fixed in this batch of commits. The most critical was a data-loss window in the migration path: when upgrading from an older storage format, the process previously held its write lock only during the final file-write step. A background worker landing a new summary entry during the earlier read-and-build phase would have its entry silently erased from the index — the summary file would survive on disk but become unreachable. The lock now covers the entire read-build-write cycle, so any concurrent write queues behind the migration and is applied on top of the completed result.

The second fix addressed a race where a slow process's lock could be reclaimed as stale by a second process, only for the original process's cleanup code to then delete the second process's freshly acquired lock — reopening the concurrent-write window it was meant to close. Both lock release functions and the periodic lock-refresh operation now read the stored owner identity before acting, and skip the operation entirely when the lock belongs to a different process.

The third fix corrected a structural problem: the write lock protecting the shared summary branch was stored inside each worktree's own local directory, meaning two worktrees of the same repository could each hold "their own" copy of the lock simultaneously and write to the shared branch concurrently. The lock is now stored inside the repository's shared git directory, which all worktrees resolve to the same physical path. The per-commit worker lock was intentionally left per-worktree, since each worktree processes its own independent queue and running those workers in parallel is the correct behavior.

## Topics (8)
<details>
<summary><strong>01 · [0291c58] Split worker and orphan-write locks to fix orphaned queue entries</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A race condition caused queue entries to be permanently orphaned: when any background writer briefly held the single shared lock file, the post-rewrite hook mistakenly concluded a worker was already running and skipped spawning a new one. The queue entry was never processed.

- **Two locks instead of one**: The core insight from the pre-implementation review was that a single lock file cannot simultaneously mean "a worker is draining the queue" and "someone is writing to the orphan branch." Giving each role its own file makes `isWorkerLockHeld` unambiguous — a brief background writer can no longer produce a false positive that suppresses worker spawning. The trade-off is a slightly more complex locking surface, but the two locks are fully independent and never nest, so deadlock is not a concern.
- **Fail-fast for worker lock, poll-with-timeout for orphan-write lock**: Worker lock contention means two workers are racing — the right answer is immediate rejection, not waiting. Orphan-write lock contention is millisecond-scale (a single file write), so a short poll loop almost always succeeds within one or two intervals and avoids needless "deferred" outcomes that would leave index state inconsistent. Using different strategies for each lock matches the actual contention profile of each use case.
- **Worker path throws on orphan-write timeout; background path defers silently**: The pre-implementation review flagged that a failed write in the worker path must not silently delete the queue entry. The 5-second timeout for the worker path is long enough to outlast any realistic background writer, and throwing causes the queue entry to remain for retry. Background callers (catalog scans, alias resolution) use 1 second and log at debug level on deferral — these are opportunistic writes and a missed update is recoverable on the next run.

**✅ What Was Implemented**

- Created `cli/src/core/Locks.ts` with two independent lock primitives: `worker.lock` (fail-fast, held for the full duration of a queue drain) and `orphan-write.lock` (poll-with-timeout, held only during each read-modify-write critical section on the orphan branch). The worker lock also gains a `refreshWorkerLockMtime` function called on a 60-second interval so long-running LLM calls cannot be mistaken for a crashed worker. The root-cause fix is in `PostRewriteHook.ts`, which now calls `isWorkerLockHeld` — a function that checks only `worker.lock` and ignores `orphan-write.lock`.
- Migrated all orphan-branch write paths in `cli/src/core/SummaryStore.ts` to acquire `orphan-write.lock` around their full read-modify-write cycle: `storeSummary`, `migrateOneToOne`, `mergeManyToOne`, `removeFromIndex`, `scanTreeHashAliases`, `getCatalogWithLazyBuild`, `migrateIndexToV3`, `saveTranscriptsBatch`, `storePlans`, and `storeNotes`. A shared helper `withRequiredOrphanWriteLock` reduces boilerplate for the worker path (5 s timeout, throws on failure) while background callers use a 1 s timeout and defer gracefully.
- Removed the old `acquireLock`, `releaseLock`, `isLockHeld`, `isLockStale`, and `LOCK_FILE` exports from `cli/src/core/SessionTracker.ts` entirely. `cli/src/commands/DoctorCommand.ts` was updated to use `isWorkerLockStale` and `releaseWorkerLock` from the new module, and its UI label was changed from "Lock file" to "Worker lock".

**📋 Future Enhancements**

The `removeFromIndex` failure path still defers with a log warning rather than throwing to the caller, deviating from the original plan. This was preserved to avoid breaking an existing test that asserts defer behavior, but should be revisited if `removeFromIndex` gains a production caller that requires strict delivery guarantees.

**📁 FILES**
- `cli/src/core/Locks.ts`
- `cli/src/core/SummaryStore.ts`
- `cli/src/hooks/PostRewriteHook.ts`
- `cli/src/commands/DoctorCommand.ts`
- `cli/src/core/SessionTracker.ts`

</blockquote>
</details>
<details>
<summary><strong>02 · [d24f1de] Fix lost-update race in tree-hash alias scanning</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A background scan that matches commit aliases could overwrite entries that a concurrent worker had just written, permanently losing those summaries. The scan read the index before acquiring the write lock, then wrote back a stale copy that didn't include the worker's new data.

- **Keep expensive git calls outside the lock**: Tree-hash lookups are O(n) git operations and can take hundreds of milliseconds on large repositories. Running them inside the lock would block every concurrent orphan-branch writer for that entire duration. The preflight result is treated as tentative and re-validated inside the lock at negligible cost, giving safety without the throughput penalty.
- **Merge only the alias field, not the full index**: Writing back the preflight index object (even with updated aliases) would silently drop any entries the worker added after the preflight read. Merging only the alias delta into the freshly loaded index preserves all concurrent writes and keeps the fix minimal in scope.

**✅ What Was Implemented**

- Refactored `scanTreeHashAliases` in `cli/src/core/SummaryStore.ts` into two phases: a preflight phase (outside the lock) that performs the expensive tree-hash lookups, and a critical section (inside the lock) that re-reads the freshly persisted index before merging. The write now uses the fresh index as its base, so any entries the worker added between preflight and lock-acquire are preserved. Candidates that a concurrent writer already resolved are silently dropped.
- Added three regression tests in `cli/src/core/SummaryStore.test.ts` covering the lost-update scenario (worker entry survives scan's write), the already-aliased candidate drop, and the already-promoted-to-entry candidate drop. Existing tree-hash tests were updated to supply two mock reads (preflight + inside-lock re-read) instead of one.

**📁 FILES**
- `cli/src/core/SummaryStore.ts`

</blockquote>
</details>
<details>
<summary><strong>03 · [d24f1de] Extend orphan-write lock timeout to 30 seconds with documented rationale</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The previous 5-second timeout on the write lock was too short: a large-repository one-time migration running concurrently with a commit could hold the lock for several seconds, causing the worker to time out and permanently lose a summary entry.

- **30 seconds, not 5 minutes**: The user initially questioned whether 5 minutes had any meaning. Analysis showed that the real worst-case contention is a one-time migration taking a few seconds, not a stale-lock scenario. A 5-minute timeout would cause workers to block for 5 minutes on a stale lock before the stale-reclaim mechanism fires, degrading UX far more than the rare data loss it would prevent. 30 seconds covers the legitimate worst case with ~10× headroom and keeps blocking time acceptable.
- **Delete the queue entry on timeout, not preserve it**: Preserving the entry for retry sounds safer but is actually worse: the transcript cursor advances before the write attempt, so a retried run reads only the subsequent transcript window and generates a summary for the wrong content. Corrupt output is a worse outcome than a missing summary, so the existing delete-on-failure behavior is kept unchanged.

**✅ What Was Implemented**

Changed `ORPHAN_WRITE_REQUIRED_TIMEOUT_MS` in `cli/src/core/SummaryStore.ts` from 5 000 ms to 30 000 ms. Added an extensive inline comment explaining the sizing rationale, why "lose this entry" is the correct failure mode on timeout (cursor has already advanced past the relevant transcript window, making retry produce a corrupt summary), and why this matches the system-wide no-retry philosophy.

**📁 FILES**
- `cli/src/core/SummaryStore.ts`

</blockquote>
</details>
<details>
<summary><strong>04 · [d24f1de] Lock orphan-branch writes in the one-time migration path</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The v1-to-v3 migration wrote directly to the orphan branch using raw git operations, bypassing the write lock entirely. A concurrent post-commit worker writing a summary at the same moment could corrupt the branch.

Adding the lock directly at the call sites rather than refactoring the migration to go through the shared storage abstraction kept the change minimal. The migration is a one-time activation path; pulling it into the abstraction layer would expand scope significantly for negligible long-term benefit.

**✅ What Was Implemented**

Added `withMigrationOrphanWriteLock` helper in `cli/src/core/SummaryMigration.ts` that acquires the orphan-write lock with a 30-second timeout and releases it on exit. Wrapped both raw-write call sites — the bulk summary migration and the migration-metadata write — with this helper. Added a module-level comment explaining why the lock lives here rather than in a shared abstraction (the migration bypasses the storage provider layer by design, as a one-time path not worth refactoring).

**📁 FILES**
- `cli/src/core/SummaryMigration.ts`

</blockquote>
</details>
<details>
<summary><strong>05 · [d24f1de] Sync VS Code worker-busy indicators to the correct lock file</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

After the lock was split into a worker lock and a short-lived write lock, the VS Code extension was still watching the old combined lock file path. This meant the "worker busy" button state and the post-worker UI refresh never fired, silently breaking the Commit, Squash, and Push button guards.

Only the worker lock is surfaced to the UI, not the short-lived write lock. The write lock is held for milliseconds at a time by background scans; exposing it as "worker busy" would cause the Commit/Squash/Push buttons to flicker constantly during normal use. The worker lock, by contrast, is held for the full duration of the LLM summarization pipeline (tens of seconds), which is exactly the window where blocking user actions is correct.

**✅ What Was Implemented**

- Updated `vscode/src/util/LockUtils.ts` to stat `worker.lock` instead of the old `lock` path. Expanded the module comment to explain the two-lock split and why only the worker lock is relevant to busy state.
- Updated `vscode/src/Extension.ts` to watch `worker.lock` via `createFileSystemWatcher`. Added an inline comment explaining why the short-lived orphan-write lock is intentionally not surfaced (watching it would cause button flicker on every background alias scan).

**📁 FILES**
- `vscode/src/util/LockUtils.ts`
- `vscode/src/Extension.ts`

</blockquote>
</details>
<details>
<summary><strong>06 · [fd15246] Migration lock now covers full read-build-write cycle to prevent index corruption</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A code review found that the migration process only held its write lock during the final file-write step, leaving a window where a background worker could write a new summary entry that the migration would then silently erase from the index.

Holding the lock for the entire read-build-write cycle means any concurrent worker write will queue behind the migration and layer its entry on top of the completed migration index, rather than being silently dropped. The trade-off is that a very large migration could exhaust a worker's 30-second write timeout, losing one summary entry — but that entry will be recovered on the next commit, and index consistency is more important than avoiding that rare single-entry delay.

**✅ What Was Implemented**

- `SummaryMigration.ts`: refactored `migrateV1toV3` to acquire the orphan-write lock before reading the v1 branch, building the file list, and reconstructing the index — not just around the final write call. The body was extracted into a private `migrateV1toV3Locked` helper so the public function's early-exit guard remains outside the lock.

**📁 FILES**
- `cli/src/core/SummaryMigration.ts`

</blockquote>
</details>
<details>
<summary><strong>07 · [fd15246] Lock release now verifies ownership before removing to prevent stale-reclaim race</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A review identified that when a slow process's lock was reclaimed as stale by a second process, the original process's cleanup code would still delete the lock — now owned by the second process — reopening a concurrent-write window.

- **PID over UUID token**: A simple process ID written into the lock file was chosen over a random UUID token. PID reuse within the lifetime of a single lock is theoretically possible but practically negligible for short-lived hook processes, and the simpler approach avoids adding a token-generation and token-passing mechanism across the call stack. A warning log on the skip path makes the race visible in diagnostics without adding complexity.
- **Shared helper for both locks**: Both release functions and the mtime refresh were unified through a single `releaseIfOwned` helper rather than duplicating the ownership check. This keeps the guard logic in one place and ensures future lock types automatically inherit the same protection.

**✅ What Was Implemented**

- `Locks.ts`: added `readLockOwnerPid` and `releaseIfOwned` helpers. Both `releaseWorkerLock` and `releaseOrphanWriteLock` now read the process ID stored in the lock file and skip the delete when it belongs to a different process, logging a warning instead.
- `Locks.ts`: `refreshWorkerLockMtime` gained the same ownership check — it skips the timestamp bump when the lock file records a foreign process ID, preventing the original holder from inadvertently extending a lock it no longer owns.
- `Locks.test.ts`: new PID-ownership tests cover the skip-on-foreign-PID path for both locks and for the mtime refresh, plus the happy-path remove-on-matching-PID case.

**📁 FILES**
- `cli/src/core/Locks.ts`

</blockquote>
</details>
<details>
<summary><strong>08 · [fd15246] Orphan-write lock moved to shared git directory so all worktrees use the same file</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The orphan summary branch is a single shared resource across all git worktrees of a repository, but the write lock was stored inside each worktree's own local directory — meaning two worktrees could each hold "their" lock simultaneously and write to the branch concurrently.

- **`.git/jollimemory/` over a worktree-adjacent `.jolli/` path**: Placing the shared lock inside the git directory (rather than alongside the main worktree's local state folder) was chosen after evaluating three options. The git directory is guaranteed to exist as long as any worktree is alive, avoids a linked worktree writing into the main worktree's directory, and is invisible to users browsing the working tree. A path adjacent to the main worktree's local folder would require the main worktree's root to exist and would cross a conceptual ownership boundary — a linked worktree writing files into a sibling worktree's directory.
- **Worker lock stays per-worktree**: The worker lock was deliberately left in the per-worktree directory. Each worktree has its own independent commit queue, and two worktrees running their own workers in parallel is correct and desirable — serializing them through a shared worker lock would cause one worktree's queue to stall for the entire duration of the other's language-model call, with no mechanism to restart the stalled worker afterward.
- **Lazy subprocess invocation for caching**: The git subprocess call is wrapped lazily rather than at module load time so that test code replacing the underlying function after import actually intercepts the call. The resolved path is then cached per working directory so repeated lock poll iterations pay the subprocess cost only once.

**✅ What Was Implemented**

- `Locks.ts`: added `resolveSharedLockDir`, which runs `git rev-parse --git-common-dir` to locate the repository's shared git directory and places the orphan-write lock under a `jollimemory/` subdirectory there. Results are cached per working directory to avoid spawning a subprocess on every poll iteration. Falls back to the per-worktree path when git is unavailable or the directory is not a git repository.
- `Locks.ts`: `acquireOrphanWriteLock` and `releaseOrphanWriteLock` now route through `resolveSharedLockDir`; the worker lock continues to use the per-worktree path unchanged, since each worktree's queue is independent and parallel worker execution across worktrees is correct behavior.
- `Locks.test.ts`: restructured to `git init` once per file (rather than per test) to avoid memory pressure from spawning ~30 subprocesses; added tests for the git-common-dir path, the fallback path, and the cache-hit behavior.

**📁 FILES**
- `cli/src/core/Locks.ts`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 9, 2026 at 7:44 PM*

### DIFF
--- a/cli/src/Cli.test.ts
+++ b/cli/src/Cli.test.ts
@@ -33,8 +33,6 @@ vi.mock("./core/SessionTracker.js", () => ({
 	loadConfig: vi.fn().mockResolvedValue({}),
 	saveConfig: vi.fn().mockResolvedValue(undefined),
 	// Used by doctor / clean commands
-	isLockStale: vi.fn().mockResolvedValue(false),
-	releaseLock: vi.fn().mockResolvedValue(undefined),
 	loadAllSessions: vi.fn().mockResolvedValue([]),
 	countActiveQueueEntries: vi.fn().mockResolvedValue(0),
 	countStaleSessions: vi.fn().mockResolvedValue(0),
@@ -43,6 +41,11 @@ vi.mock("./core/SessionTracker.js", () => ({
 	pruneStaleQueueEntries: vi.fn().mockResolvedValue(0),
 	checkStaleSquashPending: vi.fn().mockResolvedValue(null),
 	deleteSquashPending: vi.fn().mockResolvedValue(undefined),
+}));
+
+vi.mock("./core/Locks.js", () => ({
+	isWorkerLockStale: vi.fn().mockResolvedValue(false),
+	releaseWorkerLock: vi.fn().mockResolvedValue(undefined),
 }));
 
 vi.mock("./core/JolliApiUtils.js", async () => {
@@ -3444,15 +3447,14 @@ describe("CLI", () => {
 			} = {},
 		): Promise<string[]> {
 			const { getStatus } = await import("./install/Installer.js");
-			const { isLockStale, countActiveQueueEntries, loadConfig, loadAllSessions } = await import(
-				"./core/SessionTracker.js"
-			);
+			const { countActiveQueueEntries, loadConfig, loadAllSessions } = await import("./core/SessionTracker.js");
+			const { isWorkerLockStale } = await import("./core/Locks.js");
 			const { orphanBranchExists } = await import("./core/GitOps.js");
 			const { traverseDistPaths } = await import("./install/DistPathResolver.js");
 
 			// Reset (prevent bleed-over from prior tests' mockResolvedValueOnce)
 			vi.mocked(getStatus).mockReset();
-			vi.mocked(isLockStale).mockReset();
+			vi.mocked(isWorkerLockStale).mockReset();
 			vi.mocked(countActiveQueueEntries).mockReset();
 			vi.mocked(loadConfig).mockReset();
 			vi.mocked(loadAllSessions).mockReset();
@@ -3469,7 +3471,7 @@ describe("CLI", () => {
 				summaryCount: 0,
 				orphanBranch: "jollimemory/summaries/v3",
 			});
-			vi.mocked(isLockStale).mockResolvedValue(overrides.lockStale ?? false);
+			vi.mocked(isWorkerLockStale).mockResolvedValue(overrides.lockStale ?? false);
 			vi.mocked(countActiveQueueEntries).mockResolvedValue(overrides.activeQueue ?? 0);
 			vi.mocked(loadConfig).mockResolvedValue(overrides.apiKey ? { apiKey: overrides.apiKey } : {});
 			vi.mocked(loadAllSessions).mockResolvedValue([]);
@@ -3489,7 +3491,7 @@ describe("CLI", () => {
 
 			expect(output).toContain("Jolli Memory Doctor");
 			expect(output).toContain("✓ Git hooks");
-			expect(output).toContain("✓ Lock file");
+			expect(output).toContain("✓ Worker lock");
 			expect(output).toContain("✓ Config");
 			expect(output).toContain("✓ dist-paths/cli");
 			expect(output).not.toContain("Run with --fix");
@@ -3503,7 +3505,7 @@ describe("CLI", () => {
 
 		it("should flag stuck lock file as fail", async () => {
 			const output = (await runDoctor(["doctor"], { lockStale: true })).join("\n");
-			expect(output).toContain("✗ Lock file");
+			expect(output).toContain("✗ Worker lock");
 			expect(output).toContain("stuck");
 		});
 
@@ -3585,12 +3587,12 @@ describe("CLI", () => {
 		});
 
 		it("should auto-fix stale lock when --fix is passed", async () => {
-			const { releaseLock } = await import("./core/SessionTracker.js");
-			vi.mocked(releaseLock).mockClear();
+			const { releaseWorkerLock } = await import("./core/Locks.js");
+			vi.mocked(releaseWorkerLock).mockClear();
 
 			const output = (await runDoctor(["doctor", "--fix"], { lockStale: true })).join("\n");
 
-			expect(releaseLock).toHaveBeenCalled();
+			expect(releaseWorkerLock).toHaveBeenCalled();
 			expect(output).toContain("Applying fixes");
 			expect(output).toContain("released");
 		});
@@ -3655,8 +3657,8 @@ describe("CLI", () => {
 
 		it("should set exitCode=1 when --fix repairs some but unfixable failures remain (Gap B)", async () => {
 			// Lock fixer succeeds; dist-path has no fixer and stays broken.
-			const { releaseLock } = await import("./core/SessionTracker.js");
-			vi.mocked(releaseLock).mockClear();
+			const { releaseWorkerLock } = await import("./core/Locks.js");
+			vi.mocked(releaseWorkerLock).mockClear();
 			process.exitCode = 0;
 
 			const output = (
@@ -3667,7 +3669,7 @@ describe("CLI", () => {
 				})
 			).join("\n");
 
-			expect(releaseLock).toHaveBeenCalled();
+			expect(releaseWorkerLock).toHaveBeenCalled();
 			expect(output).toContain("✗ dist-paths");
 			expect(process.exitCode).toBe(1);
 			process.exitCode = 0;
@@ -3677,9 +3679,9 @@ describe("CLI", () => {
 			// --fix passed, but the only failure is dist-path which has no fixer,
 			// so fixesToApply is empty. Must still exit 1 — the ✗ is still there.
 			const { install } = await import("./install/Installer.js");
-			const { releaseLock } = await import("./core/SessionTracker.js");
+			const { releaseWorkerLock } = await import("./core/Locks.js");
 			vi.mocked(install).mockClear();
-			vi.mocked(releaseLock).mockClear();
+			vi.mocked(releaseWorkerLock).mockClear();
 			process.exitCode = 0;
 
 			const output = (await runDoctor(["doctor", "--fix"], { distPaths: [], apiKey: "sk-ant-test" })).join("\n");
@@ -3688,7 +3690,7 @@ describe("CLI", () => {
 			expect(output).not.toContain("Applying fixes");
 			expect(output).toContain("✗ dist-paths");
 			expect(install).not.toHaveBeenCalled();
-			expect(releaseLock).not.toHaveBeenCalled();
+			expect(releaseWorkerLock).not.toHaveBeenCalled();
 			expect(process.exitCode).toBe(1);
 			process.exitCode = 0;
 		});
@@ -4800,14 +4802,13 @@ describe("CLI", () => {
 			vi.mocked(getGlobalConfigDir).mockReturnValue(tmpGlobalDir);
 
 			const { getStatus } = await import("./install/Installer.js");
-			const { isLockStale, countActiveQueueEntries, loadConfig, loadAllSessions } = await import(
-				"./core/SessionTracker.js"
-			);
+			const { countActiveQueueEntries, loadConfig, loadAllSessions } = await import("./core/SessionTracker.js");
+			const { isWorkerLockStale } = await import("./core/Locks.js");
 			const { orphanBranchExists } = await import("./core/GitOps.js");
 			const { traverseDistPaths } = await import("./install/DistPathResolver.js");
 
 			vi.mocked(getStatus).mockReset();
-			vi.mocked(isLockStale).mockReset();
+			vi.mocked(isWorkerLockStale).mockReset();
 			vi.mocked(countActiveQueueEntries).mockReset();
 			vi.mocked(loadConfig).mockReset();
 			vi.mocked(loadAllSessions).mockReset();
@@ -4824,7 +4825,7 @@ describe("CLI", () => {
 				summaryCount: 0,
 				orphanBranch: "jollimemory/summaries/v3",
 			});
-			vi.mocked(isLockStale).mockResolvedValue(false);
+			vi.mocked(isWorkerLockStale).mockResolvedValue(false);
 			vi.mocked(countActiveQueueEntries).mockResolvedValue(0);
 			vi.mocked(loadConfig).mockResolvedValue({ apiKey: "sk-ant-test" });
 			vi.mocked(loadAllSessions).mockResolvedValue([]);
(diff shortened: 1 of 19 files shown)


flyerli2025:
### Code review

Found 1 issue:

1. Third commit is missing the DCO `Signed-off-by:` trailer (CLAUDE.md says "DCO sign-off on every commit — `git commit -s`. CI rejects PRs without `Signed-off-by:`.")

   Commit [`04fc2b66`](https://github.com/jolliai/jolliai/commit/04fc2b6640223e8a8e2b1e88062450477eeb8534) is only a headline ("Split worker and orphan-write locks into separate directories with PID-ownership checks") with no body or `Signed-off-by:` line. The first two commits (`f4d41ca3`, `1a75bf1a`) both carry `Signed-off-by: sanshi <sanshi.zhang@jolli.ai>`, so this looks like a missed `-s` during an amend/rebase. The DCO check will reject the PR until the trailer is added (e.g. `git commit --amend -s` then force-push, or an interactive rebase that re-signs commit 3).

   Rule reference: [CLAUDE.md#L9](https://github.com/jolliai/jolliai/blob/04fc2b6640223e8a8e2b1e88062450477eeb8534/CLAUDE.md#L9)

flyerli2025 (approved):

PR #98, opened 2026-05-11
Add JM space binding dialog and 412 binding-required handling

jordangreen-jolli:
## Jolli Memory

https://jolli-local.me/main/articles?doc=27

## E2E Test (5)
<details>
<summary><strong>1. Space binding dialog opens and closes after successful bind</strong></summary>
<br>
<blockquote>

**Preconditions:** Have an IntelliJ project open with a Git repository that has no Jolli Memory space bound to it yet. Make sure at least one Memory space exists on your jolli.ai account.

**Steps:**
1. Open the Jolli Memory panel in IntelliJ and click the "Push" button to trigger a push for any plan or summary.
2. Wait for the "Choose a Memory Space" dialog to appear.
3. Click on one of the spaces shown in the list to select it.
4. Click the "Bind and Push" button.
5. Wait a few seconds while the button shows "Binding…" to indicate it is working.
6. Check that the dialog closes on its own and the push completes without the window staying frozen.

**Expected Results:**
- The dialog should close automatically after the binding is created.
- The push should proceed and finish normally — no spinner or frozen window left behind.
- The dialog should not remain stuck in a loading state after the server responds.

</blockquote>
</details>
<details>
<summary><strong>2. Space binding dialog shows race-winner banner when another user already bound the repo</strong></summary>
<br>
<blockquote>

**Preconditions:** Have two users or sessions ready so that a second binding can be created for the same repo while the dialog is open. Alternatively, ask a teammate to bind the repo from their machine just before you click "Bind and Push."

**Steps:**
1. Open the Jolli Memory panel and click "Push" to trigger the "Choose a Memory Space" dialog.
2. Select a space from the list but do NOT click "Bind and Push" yet.
3. Have a teammate (or second session) bind the same repo to any space from their machine.
4. Now click "Bind and Push" in your dialog.
5. Wait for the dialog to respond.
6. Check the message that appears inside the dialog.

**Expected Results:**
- A yellow banner should appear inside the dialog saying a teammate already bound the repo and naming the space they chose.
- The spaces list should be hidden and replaced by the banner.
- The button should change to "OK, Push Now."
- Clicking "OK, Push Now" should close the dialog and complete the push using the teammate's binding.

</blockquote>
</details>
<details>
<summary><strong>3. Space binding dialog shows empty state when no spaces exist</strong></summary>
<br>
<blockquote>

**Preconditions:** Have an IntelliJ project open with a Git repository that has no Jolli Memory space bound to it. Make sure the jolli.ai account has zero Memory spaces created.

**Steps:**
1. Open the Jolli Memory panel and click "Push" to trigger the "Choose a Memory Space" dialog.
2. Look at the area where the list of spaces would normally appear.
3. Check whether the "Bind and Push" button is clickable.

**Expected Results:**
- The list area should be empty with a message saying no Memory spaces are available and directing you to create one on jolli.ai.
- The "Bind and Push" button should be greyed out and not clickable.

</blockquote>
</details>
<details>
<summary><strong>4. Cancelling the space binding dialog does not push</strong></summary>
<br>
<blockquote>

**Preconditions:** Have an IntelliJ project open with a Git repository that has no Jolli Memory space bound to it.

**Steps:**
1. Open the Jolli Memory panel and click "Push" to trigger the "Choose a Memory Space" dialog.
2. Click the "Cancel" button without selecting any space.
3. Check the Jolli Memory panel for any push activity or success message.

**Expected Results:**
- The dialog should close immediately.
- No push should be sent — no success notification, no error, and no loading indicator should appear.

</blockquote>
</details>
<details>
<summary><strong>5. Push succeeds with correct repo identity after binding is set up</strong></summary>
<br>
<blockquote>

**Preconditions:** Have an IntelliJ project open with a Git repository that already has a Jolli Memory space bound to it (binding was created in a previous session or by a teammate).

**Steps:**
1. Open the Jolli Memory panel and click "Push" for any plan or summary.
2. Wait for the push to complete.
3. Open jolli.ai in a browser and navigate to the Memory space that is bound to this repo.
4. Check that the pushed document appears in the correct space.

**Expected Results:**
- The push should complete without any error or "binding required" dialog appearing.
- The document should appear in the correct Memory space on jolli.ai, confirming the repo was identified correctly regardless of how it was cloned (SSH or HTTPS).

</blockquote>
</details>

---

## Topics (3)
<details>
<summary><strong>01 · Dialog stuck on loading after successful space binding</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

After a user selected a JM space and clicked "Bind and Push," the backend successfully created the binding but the dialog remained frozen on a loading state and never closed.

- **Close directly vs. delegate to parent**: The parent's close method guards on whether the OK button is currently enabled, which is intentionally `false` during an in-flight request. Calling close directly bypasses that guard without removing the button-disable behavior, which still serves its purpose of preventing double-clicks.
- **Keep button-disable flag**: Removing the enabled/disabled toggle was considered since it no longer gates the close, but it still prevents users from firing duplicate binding requests while the HTTP call is in-flight, so it was retained with its scope narrowed to click-gating only.
- **Modality fix scope**: All three `invokeLater` calls in the same function received the fix, not just the success path, because the 409 race banner and error message had the identical modality problem and would have been invisible under the modal dialog.

**✅ What Was Implemented**

- Diagnosed root cause: `invokeLater` called from a background thread captures `NON_MODAL` modality state by default, causing the EDT to defer the close lambda indefinitely while the modal dialog is still showing (chicken-and-egg deadlock)
- Fixed by passing `ModalityState.any()` to all three `invokeLater` calls in `doOKAction()` (success path, 409 race path, and generic error path) in `BindingChooserDialog.kt`
- Replaced `super.doOKAction()` with `close(OK_EXIT_CODE)` on the success path to avoid the parent class's `isOKActionEnabled` guard silently blocking the close
- Added structured `LOG.info` / `LOG.warn` lines throughout the bind flow (enter, pool start, HTTP return, invokeLater fire, dispose) to make future diagnosis straightforward

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/BindingChooserDialog.kt`

</blockquote>
</details>
<details>
<summary><strong>02 · Wire new push payload fields to IntelliJ push flow</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The server's push contract was updated to require repo identity fields and a document type on every push request, but the IntelliJ plugin was still sending the old payload shape, causing pushes to be misrouted or rejected.

- **Compute once before the loop**: `repoUrl` and `relativePath` are derived from git config and branch name respectively — both are constant for a given push session, so computing them once avoids redundant git subprocess calls per plan.
- **Use shared utility over inline logic**: The existing ad-hoc repo-name derivation had three subtle differences from the VSCode client (case handling, empty fallback, unparseable URLs). Using the shared `GitRemoteUtils` ensures both IDE clients produce identical `repoName` values, preventing duplicate binding rows on the server.

**✅ What Was Implemented**

- Added `docType`, `repoUrl`, and `relativePath` to `JolliPushPayload`; removed `subFolder`
- In `SummaryPanel.kt`, compute `repoUrl` and `relativePath` once before the push loop using the new `GitRemoteUtils` utility
- Pass `docType = "plan"` to each per-plan push call and `docType = "summary"` to the summary push call
- Replaced ad-hoc repo-name derivation in `handleBindingRequired` with `GitRemoteUtils.deriveRepoNameFromUrl`
- Updated the payload defaults test to construct with the now-required `docType` field

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/SummaryPanel.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/GitRemoteUtils.kt`

</blockquote>
</details>
<details>
<summary><strong>03 · Add GitRemoteUtils for canonical repo URL normalization in IntelliJ</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The IntelliJ plugin needed to produce the same canonical repo URL as the VSCode extension so that both clients resolve to the same server-side binding row for a given repository, regardless of clone transport or casing differences.

- **Exact parity with VSCode implementation**: Any divergence in normalization would cause the same physical repo to produce two different URL keys, resulting in duplicate binding rows and broken cross-client push routing. The comment explicitly flags this constraint for future maintainers.
- **Case-insensitive path hosts allowlist**: Silently lowercasing all paths would merge distinct repos on case-sensitive self-hosted forges. Lowercasing only for the three major public hosts (GitHub, GitLab, Bitbucket) is the safe default that matches VSCode behavior.
- **Port handling asymmetry**: HTTP/HTTPS ports are always preserved because self-hosted forges on non-standard HTTPS ports are common. SSH/git default ports (22/9418) are stripped because they carry no identity information and would otherwise create duplicate keys.

**✅ What Was Implemented**

- Created `GitRemoteUtils.kt` in the `bridge` package, mirroring `vscode/src/util/GitRemoteUtils.ts` normalization rules
- Handles SSH scp form, SSH URL, git URL, HTTP/HTTPS, and local file paths
- Normalizes scheme and host to lowercase; strips trailing `.git`; preserves port for HTTP/HTTPS, drops default ports for SSH/git
- Lowercases path only for known case-insensitive hosts (GitHub, GitLab, Bitbucket); preserves case for self-hosted forges
- Includes `deriveRepoNameFromUrl` (display label from last path segment) and `sanitizeBranchSlug` (branch name to path-safe slug)

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/GitRemoteUtils.kt`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 11, 2026 at 3:38 PM*

### DIFF
--- a/intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/GitRemoteUtils.kt
+++ b/intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/GitRemoteUtils.kt
@@ -1,0 +1,196 @@
+package ai.jolli.jollimemory.bridge
+
+import java.net.URI
+
+/**
+ * GitRemoteUtils
+ *
+ * Resolves a workspace's canonical remote URL for the Jolli Memory push
+ * contract. The returned string is the stable identity key the server uses
+ * to look up `jolli_memory_repo_bindings`, so the same physical repo must
+ * always yield the same string regardless of clone transport or owner/repo
+ * casing the user happened to type in their `git clone` invocation.
+ *
+ * Mirrors `vscode/src/util/GitRemoteUtils.ts` byte-for-byte in normalization
+ * behavior — any change must be made in both places, otherwise a VS Code
+ * teammate and an IntelliJ teammate on the same repo will produce different
+ * `repoUrl` keys and end up with two separate bindings.
+ *
+ * Normalization rules:
+ *   - SSH scp form `git@host:owner/repo[.git]`          → `https://host/owner/repo`
+ *   - SSH URL  `ssh://git@host[:port]/path[.git]`       → `https://host[:port]/path`
+ *   - git URL  `git://host[:port]/path[.git]`           → `https://host[:port]/path`
+ *   - HTTP(S):  strip trailing `.git`, lower-case scheme + host
+ *   - No remote configured: fall back to `file://<workspaceRoot>` (forward slashes)
+ *
+ * Port handling:
+ *   - HTTP(S): always preserve the port (self-hosted forges on non-default
+ *     HTTPS ports are common).
+ *   - ssh / git: preserve the port unless it's the scheme's default (22 / 9418).
+ *
+ * Path-case handling:
+ *   - github.com / gitlab.com / bitbucket.org → lowercase path (these hosts
+ *     route owner/repo case-insensitively; not lowercasing would let two
+ *     clones with different casing fork into separate bindings).
+ *   - All other hosts → preserve path case (self-hosted Gitea/GitLab can be
+ *     case-sensitive; silently lowercasing would merge distinct repos).
+ *
+ * Trailing slashes and a single trailing `.git` are stripped.
+ */
+object GitRemoteUtils {
+
+    private val CASE_INSENSITIVE_PATH_HOSTS: Set<String> = setOf(
+        "github.com",
+        "gitlab.com",
+        "bitbucket.org",
+    )
+
+    /**
+     * Default wire ports for ssh / git. A port equal to the default carries
+     * no identity information, so `ssh://host:22/x` collapses with `ssh://host/x`.
+     * http/https are intentionally absent — keep whatever the user typed,
+     * since self-hosted forges sometimes serve on `:443`/`:80` explicitly.
+     */
+    private val SSH_GIT_DEFAULT_PORTS: Map<String, String> = mapOf(
+        "ssh" to "22",
+        "git" to "9418",
+    )
+
+    private val SSH_SCP_REGEX = Regex("""^([A-Za-z0-9_.+-]+@)([^:/\s]+):(.+)$""")
+
+    /** Returns the canonical, server-facing repo URL for the given workspace root. */
+    fun getCanonicalRepoUrl(workspaceRoot: String): String {
+        val remote = GitOps(workspaceRoot)
+            .exec("config", "--get", "remote.origin.url")
+            ?.trim()
+            .orEmpty()
+        if (remote.isEmpty()) {
+            return toFileUrl(workspaceRoot)
+        }
+        return normalizeRemoteUrl(remote, workspaceRoot)
+    }
+
+    /** Normalizes a remote URL string into the canonical form. Visible for tests. */
+    fun normalizeRemoteUrl(remote: String, workspaceRootForFallback: String): String {
+        val trimmed = remote.trim()
+        if (trimmed.isEmpty()) {
+            return toFileUrl(workspaceRootForFallback)
+        }
+
+        if (!trimmed.contains("://")) {
+            val sshMatch = SSH_SCP_REGEX.matchEntire(trimmed)
+            if (sshMatch != null) {
+                val host = sshMatch.groupValues[2].lowercase()
+                val pathPart = normalizePathCase(host, stripGitSuffixAndSlashes(sshMatch.groupValues[3]))
+                return "https://$host/$pathPart"
+            }
+        }
+
+        val parsed = try {
+            URI(trimmed)
+        } catch (_: Exception) {
+            return toFileUrl(workspaceRootForFallback)
+        }
+
+        val scheme = parsed.scheme?.lowercase()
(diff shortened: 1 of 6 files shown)


flyerli2025:
P1: JolliApiClient.kt (line 513) treats every HTTP 412 as BindingRequiredError. That is broader than the existing VS Code client, which only enters the binding flow for 412 plus error == "binding_required" at JolliPushService.ts (line 217). If the backend returns any other precondition failure under 412, IntelliJ will open the space chooser and retry a push that can never succeed.

P1: JolliApiClient.kt (line 514) drops repo identity when the 412 body omits repoUrl. The TS client falls back to the original request payload’s repoUrl at JolliPushService.ts (line 219), but this port throws BindingRequiredError(""). That empty value then drives the chooser flow in SummaryPanel.kt (line 350) and SummaryPanel.kt (line 382), so the dialog is keyed under "" and createBinding() can be sent an empty repoUrl, which breaks bind-and-retry on servers that omit that field.

flyerli2025 (approved):

PR #111, opened 2026-05-15
Add Cursor IDE as transcript source and surface SQLite scan errors

jordangreen-jolli:
## Quick recap

The plugin now successfully reads live conversation history from both Cursor IDE and OpenCode. Two correctness gaps in the post-commit hook were closed: transcripts are now filtered to exclude messages created after the commit, and a failure reading one source no longer aborts processing for all remaining sessions in the same run.

Getting the SQLite integration working inside the IntelliJ plugin required two separate fixes. The driver JAR was not being shipped with the plugin at all, causing an immediate crash on startup. After bundling it, a second failure surfaced: the IDE's standard Java class-loading mechanism could not see the newly bundled driver, so database connections still failed. An explicit one-time driver registration call at first use resolved this, and the status popup now shows the full error message inline when an integration is unavailable — making future diagnosis possible without opening a log file.

---

## E2E Test (2)
<details>
<summary><strong>1. Cursor IDE detected and shown in status panel</strong></summary>
<br>
<blockquote>

**Preconditions:** Have Cursor IDE installed on the test machine so its data folder exists. Open a project in the Jolli Memory plugin that you have previously worked on inside Cursor.

**Steps:**
1. Open your project in the IDE that has the Jolli Memory plugin installed.
2. Open the Jolli Memory status panel (look for the Jolli Memory icon or menu entry).
3. Check the list of detected AI tools shown in the panel.
4. Verify that "Cursor" appears in the list and shows as detected.
5. Now open the same project in Cursor, have a short conversation with the Cursor AI assistant, then switch back to the IDE with the plugin.
6. Trigger a git commit on the project and check that the commit summary includes content from your Cursor conversation.

**Expected Results:**
- The status panel should show Cursor as a detected AI source.
- After committing, the generated summary should reflect the conversation you had in Cursor, confirming that Cursor transcripts are being read and included.

</blockquote>
</details>
<details>
<summary><strong>2. SQLite scan error surfaced in status panel when Cursor or OpenCode database is locked</strong></summary>
<br>
<blockquote>

**Preconditions:** Have either Cursor IDE or OpenCode installed. The easiest way to trigger a locked database is to have Cursor actively running while you check the status panel.

**Steps:**
1. Make sure Cursor (or OpenCode) is open and actively in use so its database may be locked.
2. Open the Jolli Memory status panel in your IDE.
3. Look at the entry for Cursor (or OpenCode) in the panel.
4. Check whether any warning or error indicator appears next to that entry.
5. Close Cursor (or OpenCode) completely, then refresh or reopen the Jolli Memory status panel.
6. Check the same entry again.

**Expected Results:**
- While Cursor is running and the database is locked, the status panel should show a warning or error message next to the Cursor entry (such as "locked" or a similar indicator) rather than silently showing zero sessions.
- After closing Cursor, the error indicator should disappear and the panel should show Cursor sessions normally.

</blockquote>
</details>

---

## Topics (4)
<details>
<summary><strong>01 · SQLite driver not available to plugin at runtime causing crash on startup</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

After adding live SQLite scanning for OpenCode and Cursor sessions, the plugin crashed on startup with a "no class definition found" error for the SQLite driver. The driver JAR was present at compile time but not shipped inside the plugin, so the IDE's class loader couldn't find it when the status panel tried to open a database.

- **Bundle vs. rely on host JVM**: The SQLite driver had been marked compile-only on the assumption that all SQLite access happened inside the standalone hooks JAR, which bundles its own copy. Adding live scans inside the plugin broke that assumption. Bundling the driver into the plugin was the only correct path — the IDE's JVM has no SQLite driver of its own, and there was no other mechanism to make it available at runtime.
- **Explicit class loading over service discovery**: JDBC's standard auto-registration mechanism uses the system class loader, which cannot see JARs loaded by the plugin's isolated class loader. Explicitly loading the driver class by name at first use is the established workaround for this IntelliJ plugin constraint. The alternative — restructuring the plugin to use a different class loader delegation model — would have been far more invasive for a narrow problem.

**✅ What Was Implemented**

- Changed `org.xerial:sqlite-jdbc` from `compileOnly` to `implementation` in `intellij/build.gradle.kts` so the driver JAR is bundled into the plugin's `lib/` directory at build time. This raised the plugin zip from 17 MB to 31 MB.
- Added a lazy one-time `Class.forName("org.sqlite.JDBC")` call at the top of `withReadOnlyDb` in `intellij/src/main/kotlin/ai/jolli/jollimemory/core/SqliteHelpers.kt`. This forces the driver's static initializer to run from the plugin's class loader and register itself with the JDBC driver manager, resolving a second failure ("no suitable driver found") that appeared after the first fix.

**📁 FILES**
- `intellij/build.gradle.kts`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/core/SqliteHelpers.kt`

</blockquote>
</details>
<details>
<summary><strong>02 · Cursor IDE added as transcript source with per-session error isolation</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The post-commit hook needed to read conversation transcripts from Cursor IDE and OpenCode in addition to the existing sources. There were also two correctness gaps: transcripts newer than the commit timestamp were being included, and a failure reading one source could abort processing for all remaining sessions.

- **Fix both gaps together in one edit**: The timestamp cutoff fix was a two-character change (passing an already-available variable), and the error isolation was a small wrapper around the same block. Combining them into one edit kept the diff minimal and avoided a second build/test cycle. The risk of the combined change was assessed as very low since both paths were already exercised by existing tests.
- **Continue-on-error over fail-fast**: Aborting the entire post-commit run when one session source fails is a poor trade-off — a locked Cursor database should not prevent Claude or Gemini transcripts from being summarized. The continue-on-error pattern mirrors how the CLI worker already handles per-session failures, keeping behavior consistent across platforms.

**✅ What Was Implemented**

- Wrapped the per-session transcript reading block in `PostCommitHook.kt` in a try/catch that logs the error and continues to the next session on failure. This prevents a locked or corrupt database for one source from aborting the entire post-commit run.
- Passed `commitInfo.date` as the `beforeTimestamp` argument to `OpenCodeSupport.readTranscript` and `CursorSupport.readTranscript`. Transcript messages newer than the commit are now excluded, matching the behavior already implemented in the reader functions but previously not activated from the hook.

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/hooks/PostCommitHook.kt`

</blockquote>
</details>
<details>
<summary><strong>03 · Status popup now shows full error detail for unavailable integrations</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The hover popup on the status indicator was showing "unavailable — unknown" for OpenCode with no further detail, making it impossible to diagnose why the integration was failing without digging through log files.

Surfacing the error message directly in the popup was chosen over log-file inspection because it gives immediate, in-context feedback without requiring the developer to locate and tail a log file. The message is already captured in the structured error object at scan time, so no additional work was needed to produce it — only to display it.

**✅ What Was Implemented**

Added the raw error message as a gray sub-line beneath the "unavailable" label in the status popup HTML builder in `JolliMemoryToolWindowFactory.kt`, with HTML entity escaping applied. This was added for both the OpenCode and Cursor rows. The change was introduced as a diagnostic aid during the SQLite driver investigation and left in place as a permanent improvement.

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/JolliMemoryToolWindowFactory.kt`

</blockquote>
</details>
<details>
<summary><strong>04 · IntelliJ plugin build requires JDK 21 — path documented for future sessions</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The plugin build failed repeatedly because the default Java installation on the machine was version 17 (too old to emit the required bytecode) and the newest available version was 25 (too new for the Gradle version in use). The correct JDK had to be located and the build re-run manually.

The JDK path was recorded in project memory rather than hardcoded into the build configuration. Adding a Gradle toolchain block to `build.gradle.kts` would make the build self-contained and eliminate the manual override, but that change was deferred — the memory entry is sufficient for now and avoids touching build configuration during an unrelated debugging session.

**✅ What Was Implemented**

Identified the working JDK at `/opt/homebrew/Cellar/openjdk@21/21.0.11` and confirmed that setting `JAVA_HOME` to that path before running the Gradle build resolves both the "JVM target mismatch" and "unsupported Java version" errors. The path and override pattern were saved to the project memory file so future sessions do not need to rediscover it.

**📋 Future Enhancements**

Consider adding a Gradle Java toolchain declaration to `intellij/build.gradle.kts` so the build automatically selects JDK 21 without requiring a manual `JAVA_HOME` override.

**📁 FILES**
- `intellij/build.gradle.kts`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 15, 2026 at 2:06 PM · via Anthropic*

### DIFF
--- a/intellij/build.gradle.kts
+++ b/intellij/build.gradle.kts
@@ -33,7 +33,7 @@ dependencies {
     // The standalone hooks JAR bundles its own copies via the hooksRuntime configuration.
     compileOnly("com.google.code.gson:gson:2.12.1")
     compileOnly("org.jetbrains.kotlin:kotlin-stdlib")
-    compileOnly("org.xerial:sqlite-jdbc:3.49.1.0")
+    implementation("org.xerial:sqlite-jdbc:3.49.1.0")
     hooksRuntime("com.google.code.gson:gson:2.12.1")
     hooksRuntime("org.jetbrains.kotlin:kotlin-stdlib:2.1.20")
     hooksRuntime("org.xerial:sqlite-jdbc:3.49.1.0")
@@ -41,6 +41,7 @@ dependencies {
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
     testImplementation("io.mockk:mockk:1.13.16")
     testImplementation("io.kotest:kotest-assertions-core:5.9.1")
+    testImplementation("org.xerial:sqlite-jdbc:3.49.1.0")
 }
 
 intellijPlatform {
--- a/intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/SummaryReader.kt
+++ b/intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/SummaryReader.kt
@@ -2,6 +2,7 @@ package ai.jolli.jollimemory.bridge
 
 import ai.jolli.jollimemory.core.CodexSessionDiscoverer
 import ai.jolli.jollimemory.core.CommitSummary
+import ai.jolli.jollimemory.core.CursorSupport
 import ai.jolli.jollimemory.core.GeminiSupport
 import ai.jolli.jollimemory.core.JmLogger
 import ai.jolli.jollimemory.core.OpenCodeSupport
@@ -30,6 +31,15 @@ class SummaryReader(private val projectDir: String, private val git: GitOps) {
         val branchExists = git.branchExists(ORPHAN_BRANCH)
         val summaryCount = if (branchExists) countSummaries() else 0
 
+        // Lightweight DB health checks — no full scan, just open + trivial query
+        val openCodeInstalled = OpenCodeSupport.isOpenCodeInstalled()
+        val openCodeError = if (openCodeInstalled && config.openCodeEnabled != false)
+            OpenCodeSupport.checkDbHealth() else null
+
+        val cursorInstalled = CursorSupport.isCursorInstalled()
+        val cursorError = if (cursorInstalled && config.cursorEnabled != false)
+            CursorSupport.checkDbHealth() else null
+
         return StatusInfo(
             enabled = hooksInstalled,
             claudeHookInstalled = installer.isClaudeHookInstalled(),
@@ -45,8 +55,12 @@ class SummaryReader(private val projectDir: String, private val git: GitOps) {
             codexEnabled = config.codexEnabled,
             geminiDetected = GeminiSupport.isGeminiInstalled(),
             geminiEnabled = config.geminiEnabled,
-            openCodeDetected = OpenCodeSupport.isOpenCodeInstalled(),
+            openCodeDetected = openCodeInstalled,
             openCodeEnabled = config.openCodeEnabled,
+            openCodeScanError = openCodeError,
+            cursorDetected = cursorInstalled,
+            cursorEnabled = config.cursorEnabled,
+            cursorScanError = cursorError,
         )
     }
 
--- a/intellij/src/main/kotlin/ai/jolli/jollimemory/core/CursorSupport.kt
+++ b/intellij/src/main/kotlin/ai/jolli/jollimemory/core/CursorSupport.kt
@@ -1,0 +1,409 @@
+package ai.jolli.jollimemory.core
+
+import com.google.gson.JsonParser
+import java.io.File
+import java.net.URI
+import java.nio.file.Paths
+import java.time.Instant
+
+/**
+ * Cursor Support — session discovery and transcript reading for Cursor's
+ * Composer AI agent.
+ *
+ * Cursor stores all Composer conversations in a *global* SQLite database at:
+ *   ~/Library/Application Support/Cursor/User/globalStorage/state.vscdb  (macOS)
+ *   ~/.config/Cursor/User/globalStorage/state.vscdb                      (Linux)
+ *   %APPDATA%/Cursor/User/globalStorage/state.vscdb                      (Windows)
+ *
+ * Rows in the `cursorDiskKV` table are JSON BLOBs keyed by:
+ *   composerData:<composerId>        — full composer metadata + bubble headers
+ *   bubbleId:<composerId>:<bubbleId> — individual message blobs
+ *
+ * There is NO authoritative "this composer belongs to this workspace" pointer in
+ * the global DB. Per-workspace `state.vscdb` files (under
+ * User/workspaceStorage/<wsHash>/) DO contain a `composer.composerData` row in
+ * their `ItemTable` with `lastFocusedComposerIds` and `selectedComposerIds`.
+ *
+ * Attribution algorithm (anchor-only with staleness cutoff):
+ *   1. Workspace lookup — scan each <wsHash>/workspace.json for a `folder` URI
+ *      that resolves to projectDir. Stop at the first match.
+ *   2. Anchor extraction — read the per-workspace state.vscdb and union the two
+ *      pointer arrays (`lastFocusedComposerIds` + `selectedComposerIds`).
+ *   3. Filter — include only anchored composers whose `lastUpdatedAt` is within
+ *      the last 48 h. Non-anchored composers are never included, preventing
+ *      cross-project transcript contamination on multi-repo machines.
+ *
+ * Synthetic transcript path: "<globalDbPath>#<composerId>" (matches OpenCode).
+ *
+ * Kotlin port of `cli/src/core/{CursorDetector,CursorSessionDiscoverer,CursorTranscriptReader,VscodeWorkspaceLocator}.ts`.
+ */
+object CursorSupport {
+
+	private val log = JmLogger.create("CursorSupport")
+
+	/** Sessions older than 48 hours are considered stale (matches other sources). */
+	private const val SESSION_STALE_MS = 48 * 60 * 60 * 1000L
+
+	/** Cursor bubble.type → transcript role. Other values (system messages, tool calls) are skipped. */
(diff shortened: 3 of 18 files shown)


flyerli2025:
Code review findings:

1. `intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/SummaryReader.kt:35-38` ignores `config.openCodeEnabled` / `config.cursorEnabled` and still scans both SQLite backends whenever the DB exists. Because `StatusPanel.kt:210-252` renders `scanError` before checking the enabled flag, a user who explicitly disabled Cursor/OpenCode can still get `unavailable — locked/corrupt/...` warnings for that integration. That breaks the meaning of the toggle and keeps touching external DBs even after opt-out.

2. `intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliMemoryService.kt:321-355` performs `refreshStatus()` synchronously, and that now calls `SummaryReader.getStatus()`, which in turn runs `CursorSupport.discoverSessions()` on every refresh. `CursorSupport.kt:105-150` does a full `SELECT ... WHERE key LIKE 'composerData:%'` scan over the global Cursor DB. Since `refreshStatus()` is triggered from initialization and repo-change listeners, this adds an O(total Cursor conversations) SQLite walk to routine IDE refreshes and is likely to cause noticeable UI stalls on larger Cursor databases.\n\n3. `intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/SummaryReader.kt:29-46` still sets `activeSessions = SessionTracker.loadAllSessions(projectDir).size`, which excludes on-demand sources like Cursor and OpenCode entirely. But `StatusPanel.kt:135-140` now labels that value as sessions "across all integrations". After this change, a user can have Cursor detected and actively contributing transcripts while the STATUS panel still shows `Sessions: 0`, which is misleading product behavior.

intellij-debug.log
flyerli2025, line 1: This debug log file shouldn't be in the PR

flyerli2025:
Findings:
P1 PostCommitHook.kt (line 174): the per-session loop still has no try/catch around transcript reads. OpenCodeSupport.readTranscript(...) and CursorSupport.readTranscript(...) both throw on read failure, so one locked/corrupt SQLite session will still abort the whole worker and prevent later sessions from being summarized. That means the advertised continue-on-error fix has not actually landed.

P1 PostCommitHook.kt (line 180): the new SQLite-backed readers are still invoked without the commit cutoff timestamp. Both OpenCodeSupport.readTranscript(...) and CursorSupport.readTranscript(...) accept beforeTimestamp, but the hook calls only pass (transcriptPath, cursor), so messages created after the commit can still be pulled into that commit’s summary.

P2 JolliMemoryService.kt (line 121) and SummaryReader.kt (line 37): the new status scans run against resolvedRoot/main worktree root, not the currently opened worktree path. Cursor and OpenCode both attribute sessions by exact workspace directory, so in linked worktrees the status panel can now report no sessions / no scan result for the wrong path even though commits made inside that worktree would be attributed correctly.

P3 JolliMemoryToolWindowFactory.kt (line 545): the MEMORIES-header hover popup still only shows scanError.kind, not the detailed scanError.message. The PR summary says the full SQLite error is surfaced inline for diagnosis, but this implementation still hides the actionable part of the failure in that popup.

flyerli2025 (dismissed):

flyerli2025:
Approved with below issues:

High: StatusInfo.openCodeScanError / cursorScanError are never actually populated, so the new “unavailable” UI path can’t trigger. SummaryReader.getStatus() still only checks isOpenCodeInstalled() / isCursorInstalled() and returns detection booleans, but it never calls discoverSessions() or carries the ScanResult.error forward. That means a locked/corrupt SQLite DB will still look like a normal detected integration in both the status list and popup, which breaks the main user-facing goal of this PR. Refs: intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/SummaryReader.kt:27-53, intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/StatusPanel.kt:207-253, intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/JolliMemoryToolWindowFactory.kt:545-563.

Medium: The session count was relabeled to “Claude only”, but the value is still SessionTracker.loadAllSessions().size, and that registry includes Gemini sessions too. GeminiAfterAgentHook persists source = TranscriptSource.gemini, so the count will be wrong as soon as a user has active Gemini sessions. Refs: intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/SummaryReader.kt:29-41, intellij/src/main/kotlin/ai/jolli/jollimemory/core/SessionTracker.kt:66-70, intellij/src/main/kotlin/ai/jolli/jollimemory/hooks/GeminiAfterAgentHook.kt:45-52, intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/StatusPanel.kt:135-140, intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/JolliMemoryToolWindowFactory.kt:511-512.

Medium: Even if the scan errors are wired up, the header health dot will still stay green because hasWarnings() never considers openCodeScanError or cursorScanError. So the detailed popup could say “Cursor unavailable”, while the top-level status indicator still advertises “all good”. Refs: intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/JolliMemoryToolWindowFactory.kt:410-441, intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/JolliMemoryToolWindowFactory.kt:545-563.

summerfang-jolli (approved):

PR #124, opened 2026-05-19
Add healMissingVisibleMarkdown to restore deleted Memory Bank files

flyerli2025:
## Quick recap

The developer added automatic recovery for Memory Bank files that were deleted by external causes while their metadata remained intact. When users expand a repository in the sidebar, missing visible Markdown files are automatically regenerated from the hidden JSON source. A new `jolli heal-folder` command also allows users to manually trigger recovery from the terminal. The heal operation is idempotent and safe to run on every panel refresh, imposing no performance penalty on the hot path.

The healer anchors all path reconstruction to the hidden JSON summary files, which contain the authoritative branch and commit message and are never modified by user edits. This prevents the healer from writing files to the wrong folder even when the manifest has drifted or is incomplete due to legacy data or user edits. A critical safety improvement prevents permanent data loss: manifest entries are preserved by default when their hidden JSON source is also missing, and only dropped when the caller explicitly has a truth source (the orphan branch in dual-write mode) to repopulate from. Entries that fail to heal due to transient I/O errors are now distinguished from truly missing files and are preserved for retry, while path drift is detected and logged as a warning rather than silently overwriting user edits.

The sidebar now heals missing files synchronously within the file-listing call, so recovered files appear immediately without requiring a manual refresh. The per-session cache that tracks clean repos is now keyed by absolute path instead of directory name, eliminating collisions when users switch their "Local Folder" setting to a different parent. The cache is automatically invalidated on every sidebar refresh, covering all refresh paths with a single choke point. Manifest drops are batched into a single write operation to improve performance from O(N²) to O(N). Comprehensive test coverage reproduces the actual user-reported scenario and validates all edge cases, including idempotency, handling of ghost entries where the hidden JSON is also missing, and the new opt-in drop behavior that prevents accidental data loss in folder-only mode.

---

## E2E Test (6)
<details>
<summary><strong>1. Recover missing Memory Bank files from hidden backup</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a Memory Bank folder where some commit summary files have been accidentally deleted from disk (but the hidden backup files still exist). You can simulate this by:
1. Opening your Memory Bank folder in a file explorer
2. Navigating to a branch folder (e.g., "main" or "feature-xyz")
3. Deleting one or more `.md` files manually

**Steps:**
1. Open the VS Code sidebar and expand the Memory Bank Folders panel
2. Click on the repository that has missing files
3. Wait 2-3 seconds and observe the file list
4. Verify that the deleted files now reappear in the branch folders
5. Open one of the recovered files to confirm it contains the correct commit message and content

**Expected Results:**
- The missing `.md` files should automatically regenerate in their original locations
- The recovered files should display the same commit messages and content as before deletion
- No manual refresh or restart should be required — recovery happens automatically when you expand the folder

</blockquote>
</details>
<details>
<summary><strong>2. Run explicit recovery command from terminal</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a Memory Bank folder with missing commit summary files (same setup as the previous scenario)

**Steps:**
1. Open a terminal and navigate to your Memory Bank project directory
2. Run the command: `jolli heal-folder`
3. Wait for the command to complete and read the output
4. Verify the counts reported (e.g., "Healed: 3 visible .md file(s) regenerated from hidden JSON")
5. Check the Memory Bank folder in your file explorer to confirm the files are now present

**Expected Results:**
- The command should print a summary showing how many files were recovered
- If all files were already present, it should print "No heal needed"
- If some files could not be recovered, it should list the count of failed entries and explain why
- All recovered files should appear in the correct branch folders with correct names

</blockquote>
</details>
<details>
<summary><strong>3. Prevent accidental data loss when switching Memory Bank locations</strong></summary>
<br>
<blockquote>

**Preconditions:** Have two different parent directories, each containing a Memory Bank folder with the same name (e.g., both have a "my-project" folder). The first location should have some missing files that were previously healed.

**Steps:**
1. Open VS Code with the first Memory Bank location configured
2. Expand the Memory Bank Folders panel and verify files are present
3. Go to Settings and change the Memory Bank parent folder to the second location
4. Wait for the sidebar to refresh
5. Expand the Memory Bank Folders panel again and verify the file list

**Expected Results:**
- The sidebar should show the correct files for the second location
- Files from the first location should NOT appear in the second location's file list
- If the second location has missing files, they should be healed independently without interference from the first location's state

</blockquote>
</details>
<details>
<summary><strong>4. Understand recovery limitations in folder-only mode</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a Memory Bank configured in folder-only mode (not dual-write). Create a missing file scenario by deleting both the visible `.md` file AND the hidden backup file for one commit.

**Steps:**
1. Open a terminal and run: `jolli heal-folder`
2. Read the output carefully, looking for the "Failed" count
3. Check the message about what to do with failed entries
4. Verify that the manifest entry still exists (the commit is still listed in `.jolli/manifest.json`)

**Expected Results:**
- The command should report the failed entry count
- The message should explain that folder-only mode has no backup source to restore from
- The manifest entry should be preserved (not deleted) so you can manually restore the hidden backup if needed
- The message should NOT mention "Re-run `jolli enable`" (that only appears in dual-write mode)

</blockquote>
</details>
<details>
<summary><strong>5. Verify path drift detection prevents silent overwrites</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a Memory Bank where you manually edited a commit summary file's location or name in the manifest, and the hidden backup file still exists with the original path information.

**Steps:**
1. Open a terminal and run: `jolli heal-folder`
2. Read the output and look for the "Skipped" count
3. Check the message about path-drifted entries
4. Verify that the manually-edited file location was NOT overwritten

**Expected Results:**
- The command should report the entry as "Skipped" (not "Healed")
- The message should mention "path-drifted" entries and suggest running `jolli reconcile` to resolve drift
- Your manually-edited file location should remain unchanged
- The hidden backup should still be intact for later recovery if needed

</blockquote>
</details>
<details>
<summary><strong>6. Distinguish transient read errors from permanent file loss</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a Memory Bank folder where a hidden backup file exists but is temporarily inaccessible (e.g., locked by another process, permission denied, or disk full). You can simulate this by temporarily changing folder permissions or having the file open in another application.

**Steps:**
1. Open a terminal and run: `jolli heal-folder`
2. Wait for the command to complete
3. Read the output and look for the "Failed" count
4. Verify that the manifest entry was NOT deleted
5. Restore normal access to the file (close the lock, restore permissions, or free disk space)
6. Run `jolli heal-folder` again

**Expected Results:**
- The first run should report the entry as "Failed" (not dropped from manifest)
- The message should indicate a transient read error (permission, disk space, or file locked)
- The manifest entry should still exist after the first run
- The second run (after restoring access) should successfully recover the file

</blockquote>
</details>

---

## Topics (10)
<details>
<summary><strong>01 · Restore deleted Memory Bank files from hidden JSON source</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Users reported that commit summaries appeared to be missing from their Memory Bank after external factors such as iCloud eviction, manual deletion, or prior-version cleanup bugs removed the visible Markdown files, even though the underlying data in hidden JSON and the orphan branch remained intact.

- **Hidden JSON as single source of truth**: The hidden JSON (summaries/{hash}.json) is the authoritative source for branch and commit message because it is written once at initial summary generation and never modified by user edits or manifest updates. Reading the hidden JSON first ensures the healer recovers files to their original locations regardless of manifest drift.
- **Idempotent and safe for hot paths**: The heal method is designed to be called on every panel refresh without performance penalty. The fast path is M `existsSync` calls (where M is manifest size), and the slow path only fires for genuinely-missing files. Calling twice in a row produces zero healed counts on the second pass because file regeneration short-circuits when the `.md` already exists on disk.
- **Automatic healing on first access with per-session caching**: Healing runs transparently when a user first expands a repository in the sidebar, rather than requiring manual invocation. A per-session cache tracks clean repos to avoid repeated scans on every list operation when no external changes have occurred.

**✅ What Was Implemented**

- Added `healMissingVisibleMarkdown` method to `FolderStorage` that walks the manifest, identifies commit entries whose visible `.md` files are missing, and regenerates them from the hidden `.jolli/summaries/<hash>.json` source. The healer reads the hidden JSON summary file first to extract the authoritative branch and commit message, then uses those values to construct the synthetic entry and write the markdown file to the correct location.
- Integrated heal into `KbFoldersService` as a call after each `reconcile()` pass, so users see missing files restored automatically when they expand a repository node in the sidebar. The sidebar now heals missing files synchronously within the file-listing call, so recovered files appear immediately without requiring a manual refresh.
- Added `jolli heal-folder` CLI command that invokes the heal method and reports counts of healed, skipped, and failed entries, with mode-specific guidance about whether failed entries can be repopulated from the orphan branch.
- Delegated heal through `DualWriteStorage` to the folder-side provider, with error handling that marks the storage dirty if the heal attempt fails.

**📋 Future Enhancements**

- The per-session memo tracking clean repos is never invalidated when external changes occur (e.g., when the user clicks Refresh or when a post-commit hook writes new files). A follow-up should wire invalidation into all external-change entry points.

**📁 FILES**
- `cli/src/core/FolderStorage.ts`
- `cli/src/commands/HealFolderCommand.ts`
- `vscode/src/services/KbFoldersService.ts`

</blockquote>
</details>
<details>
<summary><strong>02 · Fix silent data loss when manifest rows are dropped without a truth source</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The heal-folder recovery command was permanently deleting manifest entries whose hidden JSON files were also missing, even in folder-only mode where the manifest is the last record of that data, making data loss irreversible for users without an orphan branch to repopulate from.

- **Opt-in deletion is safer than opt-out**: The default behavior preserves manifest rows even when hidden JSON is missing, because folder-only mode has no truth source to repopulate from. Only callers that explicitly have a truth source (dual-write mode via the orphan branch) request the drop. This fail-safe direction prevents accidental permanent data loss.
- **Error channel distinguishes true no-ops from swallowed exceptions**: Adding an `error` field to `HealResult` lets the CLI avoid lying to users who explicitly ran a recovery command. A user sees "Heal errored" instead of "No heal needed" when the heal pass threw an exception, even if zero files were healed.
- **Storage mode awareness in CLI guidance**: The command now reads `storageMode` from config and adjusts its output accordingly. Dual-write mode tells users to re-run `jolli enable` to repopulate from the orphan branch; folder-only mode tells them to manually restore the hidden JSON or remove the manifest row. This prevents misleading advice.

**✅ What Was Implemented**

- Modified `FolderStorage.healMissingVisibleMarkdown` to accept an `opts` parameter with `dropOrphanedManifestEntries` flag (defaults to false), and only drop manifest rows when explicitly opted in. Entries that fail to heal are now preserved by default, with the failed count reported to the user.
- Updated `DualWriteStorage` to pass `dropOrphanedManifestEntries: true` by default when calling the folder storage heal method, since dual-write mode has the orphan branch as a truth source. Added error channel to `HealResult` so CLI can distinguish between "no heal needed" and "heal errored".
- Modified `HealFolderCommand` to read the storage mode from config and only request manifest drops in dual-write mode, with mode-specific guidance printed to the user about whether failed entries can be repopulated.

**📁 FILES**
- `cli/src/core/FolderStorage.ts`
- `cli/src/core/DualWriteStorage.ts`
- `cli/src/commands/HealFolderCommand.ts`

</blockquote>
</details>
<details>
<summary><strong>03 · Distinguish transient read errors from missing files during heal</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The heal process was treating all file read errors (permission denied, disk full, file locked) the same as missing files, which could cause manifest entries to be dropped when the real problem was a transient I/O error that might resolve on retry.

Separating errno categories prevents silent data loss from transient conditions. A permission error or disk-full condition is not the same as a deleted file — the manifest row should survive so the user can retry after fixing the underlying issue. This distinction is critical for safety in production environments where temporary I/O glitches are common.

**✅ What Was Implemented**

- Modified `FolderStorage.readFile` to distinguish `ENOENT` (file truly missing) from other errors like `EACCES`, `EBUSY`, `EIO`, and `EISDIR`. Only `ENOENT` is treated as "missing"; other errors are now re-thrown so they propagate as failures rather than silently becoming null.
- Updated heal logic to count read errors that are not `ENOENT` as failed entries without dropping the manifest row, preserving the data for later retry when the transient condition clears. Added categorized error logging for transient errors (`ENOSPC`, `EACCES`, `EROFS`, `EBUSY`) so users and support can diagnose permission or disk-space issues.

**📁 FILES**
- `cli/src/core/FolderStorage.ts`

</blockquote>
</details>
<details>
<summary><strong>04 · Detect and warn when manifest paths drift during heal</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The heal process was silently overwriting manifest path entries with newly computed paths, which could lose the user's manual edits or cause mismatches if branch folder names changed or slugification rules evolved between versions.

Skipping the update when paths diverge is more conservative than silently overwriting. The manifest entry is preserved as-is, and the user gets visibility into the mismatch via the WARN log. This prevents accidental loss of user-edited branch folder names or other path customizations, while still allowing the user to manually resolve the drift if needed. Path drift is not a heal failure — the hidden JSON is readable and intact, so reconcile is the appropriate tool to resolve the drift.

**✅ What Was Implemented**

Added path comparison in `FolderStorage.healMissingVisibleMarkdown`: before calling `regenerateVisibleMarkdown`, the heal loop now compares the manifest's stored path against the newly computed path from the hidden JSON. If they differ, a WARN is logged and the entry is skipped (not updated). The heal counter now correctly classifies path-drifted entries as skipped rather than failed, since the hidden JSON is intact and the entry is fully recoverable via reconcile.

**📁 FILES**
- `cli/src/core/FolderStorage.ts`

</blockquote>
</details>
<details>
<summary><strong>05 · Fix cleanRepos memo to use absolute paths and prevent cross-root leaks</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

When users changed their "Local Folder" setting to point to a different parent directory, the per-session memo tracking which repos had clean heal passes would leak across roots — if the new parent contained a repo with the same folder name as one already memoized under the old parent, the new repo would incorrectly skip reconcile and heal based on stale state from a completely different repo.

- **Absolute path keying over directory name**: Using the full `kbRoot` path as the memo key eliminates the collision risk entirely. Directory names are not unique across different parent folders, but absolute paths are globally unique within a session. This is simpler and safer than trying to composite the parent path with the directory name.
- **Invalidation at the refresh boundary**: Rather than scattering `notifyDirty()` calls across five different call sites (settings save, migration, rebuild, kbParent change, orphan listener), wiring it into `refreshKnowledgeBaseFolders()` creates a single choke point that automatically covers all refresh paths. This reduces the risk of missing an invalidation path in the future.

**✅ What Was Implemented**

- Changed `KbFoldersService` to key the clean repos memo by absolute `kbRoot` path instead of directory name, preventing same-basename repos under different parents from colliding in the memo.
- Updated `notifyDirty()` method signature to accept an optional `kbRoot` parameter (absolute path) instead of `repoDirName`, allowing callers to invalidate the memo for a specific repo or clear it entirely.
- Wired `notifyDirty()` into `SidebarWebviewProvider.refreshKnowledgeBaseFolders()` so every refresh path (manual Refresh button, settings save, migration, rebuild, kbParent move) re-arms the heal check on the next listing.
- Added regression tests verifying that `notifyDirty()` re-arms heal after a clean pass and that switching `kbParent` to a new root with a same-basename repo does not leak the old memo.

**📁 FILES**
- `vscode/src/services/KbFoldersService.ts`
- `vscode/src/views/SidebarWebviewProvider.ts`

</blockquote>
</details>
<details>
<summary><strong>06 · Batch manifest drops into a single write to avoid O(N²) performance</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The heal process was dropping orphaned manifest entries one at a time, causing the manifest file to be read and written once per dropped entry. For repos with many missing files, this created unnecessary I/O overhead.

Batching the drops reduces I/O from O(N²) to O(N) and ensures the manifest is never left in a partially-cleaned state if a write fails mid-way. This is especially important for large repos where heal might need to drop dozens of entries.

**✅ What Was Implemented**

- Added `MetadataManager.replaceFiles` method to replace the entire manifest entries list in a single read-modify-write cycle. Modified `FolderStorage.healMissingVisibleMarkdown` to collect all dropped IDs during the heal loop and call `replaceFiles` once at the end instead of calling `removeFromManifest` repeatedly.

**📁 FILES**
- `cli/src/core/MetadataManager.ts`
- `cli/src/core/FolderStorage.ts`

</blockquote>
</details>
<details>
<summary><strong>07 · End-to-end regression test for missing visible Markdown recovery</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The heal feature needed test coverage that reproduced the actual user-reported scenario: manifest entry present, hidden JSON intact, but visible `.md` file deleted by external causes.

The regression test was written as a "golden master" that documents the current broken behavior (visible `.md` stays missing after reconcile) with an explicit comment saying "flip this assertion when heal lands." This approach pins the bug in CI and makes the fix verifiable: when heal is implemented, the test assertion changes from `.toBe(false)` to `.toBe(true)`, proving the gap is closed. Both test suites from the feature branch and main were preserved during merge conflict resolution because they were independent and complementary—neither was a duplicate or override of the other, ensuring the final code is tested against both the new healing logic and the existing fallback behavior.

**✅ What Was Implemented**

- Added `reconcile heal gap (regression repro)` test suite in `FolderStorage.test.ts` that pins the current non-self-healing behavior and verifies that `regenerateVisibleMarkdown` can recover the same file that `reconcile` leaves missing.
- Added `healMissingVisibleMarkdown` test suite with six cases covering: empty manifest, successful regeneration, skipped existing files, dropped ghost entries (hidden JSON missing), ignored plan/note entries (heal scope is commits only), and idempotency.
- Added two regression tests that verify files are recovered to their correct locations even when the manifest has drifted or is incomplete: one verifies that a manifest entry with missing `source.branch` still writes to the correct branch folder (not "default"), and another verifies that a manifest entry whose title has diverged from the original commit message still writes under the original commit message's slug.
- Added integration test in `KbFoldersService.test.ts` that simulates the full user workflow: manifest entry exists, hidden JSON exists, visible `.md` is deleted, then `listChildren` is called and the file is regenerated in the background.
- Added test to `SidebarWebviewProvider.test.ts` verifying that `refreshKnowledgeBaseFolders()` calls `notifyDirty()` to invalidate the memo.

**📁 FILES**
- `cli/src/core/FolderStorage.test.ts`
- `vscode/src/services/KbFoldersService.test.ts`

</blockquote>
</details>
<details>
<summary><strong>08 · Improve error logging clarity in dual-write storage heal operations</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

When heal operations failed in dual-write storage, the error log message always said "Shadow healMissingVisibleMarkdown failed" even when the primary storage was the one that actually failed, making post-mortems harder because the log did not match reality.

Naming the actual target in the log message is a low-cost improvement that eliminates ambiguity in error investigation. The label is computed once at the start of the try block, so there is no performance cost, and it makes the log message immediately actionable without requiring the reader to understand the dual-write architecture.

**✅ What Was Implemented**

Updated `DualWriteStorage` to compute a `targetLabel` variable that names which storage target (shadow or primary) actually ran, and use that label in the error log message instead of hardcoding "Shadow". This makes error logs self-documenting so readers do not have to re-derive which side failed from context.

**📁 FILES**
- `cli/src/core/DualWriteStorage.ts`

</blockquote>
</details>
<details>
<summary><strong>09 · Clarify recovery path for stale child markdown cleanup and bug history</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The docstring in StaleChildMarkdownCleanup claimed to "undo" a prior cleanup bug, but the actual implementation cannot restore heads that were already deleted from disk — it can only prevent future deletions. This misleading documentation could cause future maintainers to misdiagnose data loss issues or attempt recovery through the wrong code path.

The docstring was rewritten to separate two distinct concerns: (1) what the cleanup pass accomplishes (draining backlog of children with `parentCommitHash != null`), and (2) what it cannot do (restore deleted heads). This prevents future readers from attempting recovery through the wrong code path and makes the relationship between cleanup and heal explicit. The change acknowledges prior bugs as historical fact rather than claiming to undo them, which is more honest about the system's recovery capabilities.

**✅ What Was Implemented**

Updated the docstring in `cli/src/core/StaleChildMarkdownCleanup.ts` to accurately describe what the cleanup functions do and do not do. The revised text clarifies that `cleanupAllBranchesStaleChildMarkdown` drains backlog of hoisted children but cannot restore heads already removed by prior bugs, and explicitly directs readers to `FolderStorage.healMissingVisibleMarkdown` as the correct recovery path for missing visible Markdown files.

**📁 FILES**
- `cli/src/core/StaleChildMarkdownCleanup.ts`

</blockquote>
</details>
<details>
<summary><strong>10 · Test coverage for plain commits across multiple branches in cleanup</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The StaleChildMarkdownCleanup test suite needed verification that the cleanup process correctly handles scenarios where multiple branches have plain commits (commits with no parent) in the index simultaneously.

The test validates a critical safety property: cleanup operations must be branch-scoped and never touch commits belonging to other branches, even when those commits share the same structural shape (parent=null). This prevents accidental data loss when multiple feature branches coexist in the index with similar commit patterns.

**✅ What Was Implemented**

Added a new test case to `cli/src/core/StaleChildMarkdownCleanup.test.ts` that verifies the cleanup operation preserves plain commits on non-active branches while processing the active branch. The test creates an index with plain commits from three different branches and confirms that running cleanup scoped to one branch leaves all entries untouched, with zero deletions and zero failures.

**📁 FILES**
- `cli/src/core/StaleChildMarkdownCleanup.test.ts`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 19, 2026 at 7:16 PM · via Anthropic*

### DIFF
--- a/cli/src/Cli.ts
+++ b/cli/src/Cli.ts
@@ -33,6 +33,7 @@ import { registerConvertCommand } from "./commands/ConvertCommand.js";
 import { registerDoctorCommand } from "./commands/DoctorCommand.js";
 import { registerDisableCommand, registerEnableCommand } from "./commands/EnableCommand.js";
 import { registerExportCommand, registerExportPromptCommand } from "./commands/ExportCommand.js";
+import { registerHealFolderCommand } from "./commands/HealFolderCommand.js";
 import { registerMigrateCommand } from "./commands/MigrateCommand.js";
 import { registerNewCommand } from "./commands/NewCommand.js";
 import { registerRecallCommand } from "./commands/RecallCommand.js";
@@ -182,6 +183,7 @@ export async function main(args?: ReadonlyArray<string>): Promise<void> {
 	registerRecallCommand(program);
 	registerSearchCommand(program);
 	registerMigrateCommand(program);
+	registerHealFolderCommand(program);
 	registerExportPromptCommand(program);
 	registerExportCommand(program);
 	registerAuthCommands(program);
--- a/cli/src/commands/HealFolderCommand.test.ts
+++ b/cli/src/commands/HealFolderCommand.test.ts
@@ -1,0 +1,225 @@
+import { Command } from "commander";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { HealResult, StorageProvider } from "../core/StorageProvider.js";
+import { registerHealFolderCommand } from "./HealFolderCommand.js";
+
+// Mock the storage factory so the command never touches a real ~/.jolli config
+// or a real git repo. Each test installs its own implementation via mockResolvedValue.
+const { mockCreateStorage, mockLoadConfig } = vi.hoisted(() => ({
+	mockCreateStorage: vi.fn(),
+	mockLoadConfig: vi.fn(),
+}));
+
+vi.mock("../core/StorageFactory.js", () => ({
+	createStorage: mockCreateStorage,
+}));
+
+vi.mock("../core/SessionTracker.js", async () => ({
+	loadConfig: mockLoadConfig,
+}));
+
+// Logger writes to disk in setLogDir; skip that path so tests don't create files.
+vi.mock("../Logger.js", async (importOriginal) => {
+	const actual = await importOriginal<typeof import("../Logger.js")>();
+	return {
+		...actual,
+		setLogDir: vi.fn(),
+		createLogger: () => ({
+			info: vi.fn(),
+			warn: vi.fn(),
+			error: vi.fn(),
+			debug: vi.fn(),
+		}),
+	};
+});
+
+function makeProgram(): Command {
+	const program = new Command();
+	program.exitOverride();
+	registerHealFolderCommand(program);
+	return program;
+}
+
+async function runCommand(args: string[]): Promise<{ stdout: string; stderr: string }> {
+	let stdout = "";
+	let stderr = "";
+	const origLog = console.log;
+	const origErr = console.error;
+	console.log = (msg: string) => {
+		stdout += `${msg}\n`;
+	};
+	console.error = (msg: string) => {
+		stderr += `${msg}\n`;
+	};
+	try {
+		await makeProgram().parseAsync(["node", "jolli", "heal-folder", ...args]);
+	} finally {
+		console.log = origLog;
+		console.error = origErr;
+	}
+	return { stdout, stderr };
+}
+
+function makeStorageWith(result: HealResult | null): StorageProvider {
+	const base: StorageProvider = {
+		readFile: vi.fn(),
+		writeFiles: vi.fn(),
+		listFiles: vi.fn(),
+		exists: vi.fn().mockResolvedValue(true),
+		ensure: vi.fn(),
+	};
+	if (result !== null) {
+		return {
+			...base,
+			healMissingVisibleMarkdown: vi.fn().mockResolvedValue(result),
+		};
+	}
+	return base;
+}
+
+describe("registerHealFolderCommand", () => {
+	beforeEach(() => {
+		vi.clearAllMocks();
+		mockLoadConfig.mockResolvedValue({ storageMode: "dual-write" });
+	});
+
+	afterEach(() => {
+		process.exitCode = undefined;
+	});
+
+	it("prints 'Heal not available' when the active storage lacks the heal method", async () => {
+		mockCreateStorage.mockResolvedValue(makeStorageWith(null));
+		const { stdout } = await runCommand(["--cwd", "/tmp/jolli-heal-test-orphan"]);
+		expect(stdout).toContain("Heal not available");
+		expect(stdout).toContain("storageMode=dual-write");
+	});
+
+	it("prints 'Manifest is empty' when the manifest had no entries", async () => {
+		mockCreateStorage.mockResolvedValue(makeStorageWith({ healed: 0, skipped: 0, failed: 0 }));
+		const { stdout } = await runCommand(["--cwd", "/tmp/jolli-heal-test-empty"]);
+		expect(stdout).toContain("Manifest is empty");
+	});
+
+	it("prints 'No heal needed' when entries exist but everything is on disk", async () => {
+		mockCreateStorage.mockResolvedValue(makeStorageWith({ healed: 0, skipped: 5, failed: 0 }));
+		const { stdout } = await runCommand(["--cwd", "/tmp/jolli-heal-test-clean"]);
+		expect(stdout).toContain("No heal needed");
+		expect(stdout).toContain("Skipped: 5");
+	});
+
+	it("reports healed counts and skipped files when files were regenerated", async () => {
+		mockCreateStorage.mockResolvedValue(makeStorageWith({ healed: 3, skipped: 18, failed: 0 }));
+		const { stdout } = await runCommand(["--cwd", "/tmp/jolli-heal-test-success"]);
+		expect(stdout).toContain("Healed:   3");
+		expect(stdout).toContain("Skipped:  18");
+		expect(stdout).not.toContain("Failed:");
+		expect(stdout).not.toContain("Re-run `jolli enable`");
+	});
+
+	it("reports failed entries with dropped IDs when manifest rows were dropped", async () => {
+		mockCreateStorage.mockResolvedValue(
+			makeStorageWith({
+				healed: 0,
+				skipped: 4,
+				failed: 2,
+				droppedIds: ["aabbccdd00112233", "eeff445566778899"],
+			}),
+		);
+		const { stdout } = await runCommand(["--cwd", "/tmp/jolli-heal-test-partial"]);
+		// Heal line is omitted when 0 — the CLI only prints non-zero counts.
+		expect(stdout).not.toMatch(/Healed:\s+0/);
+		expect(stdout).toContain("Failed:   2");
+		expect(stdout).toContain("Dropped from manifest: 2");
+		expect(stdout).toContain("aabbccdd, eeff4455");
+		expect(stdout).toContain("Re-run `jolli enable`");
+	});
+
+	it("folder-only mode keeps failed entries and tells the user not to expect orphan repopulation", async () => {
+		mockLoadConfig.mockResolvedValue({ storageMode: "folder" });
+		mockCreateStorage.mockResolvedValue(makeStorageWith({ healed: 0, skipped: 4, failed: 2 }));
+		const { stdout } = await runCommand(["--cwd", "/tmp/jolli-heal-test-folder-only"]);
+		expect(stdout).toContain("Failed:   2");
+		expect(stdout).not.toContain("Dropped from manifest");
+		expect(stdout).toContain("folder-only mode has no truth source");
+		expect(stdout).not.toContain("Re-run `jolli enable`");
+	});
+
+	it("reports both healed and failed counts together", async () => {
+		mockCreateStorage.mockResolvedValue(
+			makeStorageWith({ healed: 5, skipped: 10, failed: 1, droppedIds: ["dead00000000beef"] }),
+		);
+		const { stdout } = await runCommand(["--cwd", "/tmp/jolli-heal-test-mixed"]);
+		expect(stdout).toContain("Healed:   5");
+		expect(stdout).toContain("Skipped:  10");
+		expect(stdout).toContain("Failed:   1");
+	});
+
+	// When DualWriteStorage swallows a shadow exception it surfaces it via
+	// HealResult.error. The CLI must NOT report "No heal needed" in that case —
+	// the user explicitly ran a recovery command and deserves a true status.
+	it("prints an explicit error message when the heal pass aborted with an exception", async () => {
+		mockCreateStorage.mockResolvedValue(
+			makeStorageWith({ healed: 0, skipped: 0, failed: 0, error: "shadow storage manifest write failed" }),
+		);
+		const { stdout, stderr } = await runCommand(["--cwd", "/tmp/jolli-heal-test-errored"]);
+		expect(stderr).toContain("Heal errored");
+		expect(stderr).toContain("shadow storage manifest write failed");
+		expect(stdout).not.toContain("No heal needed");
+		expect(process.exitCode).toBe(1);
+	});
+
+	it("aborts with exit 1 when createStorage throws", async () => {
+		mockCreateStorage.mockRejectedValue(new Error("config.json corrupted"));
+		const { stderr } = await runCommand(["--cwd", "/tmp/jolli-heal-test-config-error"]);
+		expect(stderr).toContain("Heal aborted");
+		expect(stderr).toContain("config.json corrupted");
+		expect(stderr).toContain("jolli doctor");
+		expect(process.exitCode).toBe(1);
+	});
+
+	// Folder-only mode talks directly to FolderStorage which does NOT
+	// self-catch its manifest read/replace/regenerate throws. The CLI must
+	// wrap that path so a corrupted manifest / EACCES / ENOSPC becomes the
+	// same "Heal errored:" exit-1 path as a DualWriteStorage-surfaced error,
+	// instead of bubbling up as an uncaught promise rejection.
+	it("catches a synchronous throw from healMissingVisibleMarkdown (folder-only path)", async () => {
+		mockLoadConfig.mockResolvedValue({ storageMode: "folder" });
+		const base: StorageProvider = {
+			readFile: vi.fn(),
+			writeFiles: vi.fn(),
+			listFiles: vi.fn(),
+			exists: vi.fn().mockResolvedValue(true),
+			ensure: vi.fn(),
+		};
+		mockCreateStorage.mockResolvedValue({
+			...base,
+			healMissingVisibleMarkdown: vi.fn().mockRejectedValue(new Error("manifest.json is unreadable")),
+		});
+		const { stderr } = await runCommand(["--cwd", "/tmp/jolli-heal-test-folder-throw"]);
+		expect(stderr).toContain("Heal errored");
+		expect(stderr).toContain("manifest.json is unreadable");
+		expect(process.exitCode).toBe(1);
+	});
+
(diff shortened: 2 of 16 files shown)


flyerli2025:
### Code review

Found 2 issues:

1. Commit `12643e99` is missing the `Signed-off-by:` DCO trailer; CI's DCO check will reject the PR (CLAUDE.md says "DCO sign-off on every commit — `git commit -s`. CI rejects PRs without `Signed-off-by:`").

https://github.com/jolliai/jolliai/blob/12643e996f956bffe7a312025171a2720c131d2a/CLAUDE.md#L7-L12

Commit: https://github.com/jolliai/jolliai/commit/12643e996f956bffe7a312025171a2720c131d2a

2. `KbFoldersService.cleanRepos` is set after the first clean heal pass but is never invalidated — `notifyDirty()` is defined and documented as the invalidation hook ("Call from the consumer when external state changes (manual refresh, KB folder watcher fires, post-commit hook signals a write)"), but it has zero callers in `vscode/src/**`. After the first clean pass, `listChildren` permanently skips reconcile+heal for that repo for the rest of the VS Code session. If a `.md` file is later evicted (iCloud) or deleted externally, the sidebar Refresh button and `kb:foldersReset` will not restore it — only a window reload clears the `Set`. This silently undoes the rationale captured in the prior reconcile-on-every-list comment ("Runs on every list-inside-repo … per-call cost stays bounded because reconcile() short-circuits when every manifest path is still on disk").

https://github.com/jolliai/jolliai/blob/12643e996f956bffe7a312025171a2720c131d2a/vscode/src/services/KbFoldersService.ts#L91-L115

Skip site:

https://github.com/jolliai/jolliai/blob/12643e996f956bffe7a312025171a2720c131d2a/vscode/src/services/KbFoldersService.ts#L183-L210

sanshizhang-jolli (approved):

PR #127, opened 2026-05-20
Replace stale-commit panel disposal with persistent stale-readonly mode and warning banner

sanshizhang-jolli:
## Commits in this PR (3)

1. Closes JOLLI-1550: Guard all write paths against stale rewritten commits (`7f71000`)
2. Add race-window re-checks to SummaryWebviewPanel write paths (`7cba9e2`)
3. Replace stale-commit panel disposal with persistent stale-readonly mode and warning banner (`7b5dd42`)

### Commit 1 of 3: Closes JOLLI-1550: Guard all write paths against stale rewritten commits (`7f71000`)

A new safety mechanism was added to the commit summary panel to prevent writes from landing on commits that have since been amended, squashed, or rebased. Previously, if a summary panel remained open after its commit was rewritten, actions like "Push to Jolli," topic edits, recap generation, and end-to-end test operations would silently write to the orphaned commit, leaving the live commit's summary without its document link or other fields. The panel now checks the commit index before any write, identifies whether the commit has been superseded, and if so shows a warning naming both the stale commit and the live root it was folded into, then closes itself so the user can open the correct summary from the branch panel.

Coverage was extended well beyond the initial set of write operations. Plan management, note management, and Linear issue removal were all added to the protected set, and save and translate operations for plans and notes are guarded as well. Every handler that could produce a write tied to the panel's current commit now checks for a rewrite first, and every check fires before any picker, confirmation dialog, or archive call — the panel behaves as fully read-only once its commit has been superseded.

The index lookup that powers the guard was also corrected to always consult the same storage backend as the rest of the panel's write operations. In folder-storage mode the original implementation read from the wrong location, which could cause real rewrites to go undetected. The lookup now routes through the same bridge layer that handles all other writes, so the guard and the writers always see the same data regardless of which storage mode is active.

### Commit 2 of 3: Add race-window re-checks to SummaryWebviewPanel write paths (`7cba9e2`)

The Summary panel now guards against a class of data-corruption bug where a commit could be rewritten while a slow operation was in progress. Previously, the safety check ran only at the moment a user clicked an action button. If the user then waited through an AI generation call or interacted with a confirmation dialog — during which they or another tool amended the commit — the result would be silently written to the now-outdated commit record. A second check was added immediately before each final write in every handler with a meaningful wait, covering AI generation, translation, confirmation dialogs, and selection menus. The push flow was deliberately left with only the entry check: blocking a write after a successful remote call would leave the newly created article without a local record, producing a duplicate on the next push.

### Commit 3 of 3: Replace stale-commit panel disposal with persistent stale-readonly mode and warning banner (`7b5dd42`)

When a commit summary panel detected that its underlying commit had been rewritten by an amend, squash, or rebase, it previously closed itself immediately and showed a brief corner notification. The panel now stays open and transitions into a persistent read-only state: a yellow warning banner appears at the top of the summary explaining what happened, all editing controls are hidden, and a centered modal dialog gives the user a direct button to open the new commit's summary. Dismissing the modal leaves the panel intact so the user can keep reading the orphaned commit's content. Any further attempts to trigger a write action on the stale panel are silently blocked without raising another dialog.

## E2E Test (8)
<details>
<summary><strong>1. [7f71000] Stale-commit guard blocks plan and note actions on a rewritten commit</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a commit whose summary is open in the Jolli Memory panel. The commit must have been rewritten (for example, amended) so that it is no longer the live root — the extension should be in folder-storage mode or any non-default storage configuration to exercise the corrected code path.

**Steps:**
1. Open the Jolli Memory panel and navigate to the summary for the rewritten commit.
2. Click "Add Plan" and wait for a response.
3. Check whether a warning message appears instead of the plan picker opening.
4. Click "Add Markdown Note" and wait for a response.
5. Check whether a warning message appears instead of a file picker opening.
6. Try "Remove Plan" (if a plan is listed) and verify a warning appears before any confirmation dialog.

**Expected Results:**
- Each action should immediately show a warning message containing the name of the blocked operation (for example, "add plan", "add markdown note", "remove plan").
- No plan picker, file picker, or confirmation dialog should open for any of these actions.
- No changes should be saved to the summary.

</blockquote>
</details>
<details>
<summary><strong>2. [7f71000] Stale-commit guard blocks save, translate, and remove actions on a rewritten commit</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a commit summary open in the Jolli Memory panel where the commit has been rewritten. The summary should include at least one plan, one note, and one Linear issue so all removal paths can be tested.

**Steps:**
1. Open the summary for the rewritten commit in the Jolli Memory panel.
2. Open an existing plan and click "Save" — check that a warning appears and no save occurs.
3. Click "Translate" on the same plan — check that a warning appears and no translation starts.
4. Open an existing note and click "Save" — check that a warning appears and no save occurs.
5. Click "Remove" next to a Linear issue — check that a warning appears before any confirmation dialog.
6. Click "Remove" next to a note — check that a warning appears before any confirmation dialog.

**Expected Results:**
- Every action should be blocked with a warning message that names the specific operation (for example, "save plan", "translate plan", "save note", "remove note", "remove Linear issue").
- No confirmation dialogs should appear for removal actions.
- No content should be written or updated in the summary.

</blockquote>
</details>
<details>
<summary><strong>3. [7f71000] Stale panel blocked after amending a commit</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a Git repository open in VS Code with at least one commit that has a Jolli Memory summary. The commit's summary panel should be open and visible.

**Steps:**
1. Open the summary panel for a commit (the panel showing memories, recap, and the "Push to Jolli" button).
2. Without closing the panel, amend that commit in your terminal (for example, run `git commit --amend --no-edit`), so the original commit is replaced by a new one.
3. Return to VS Code and click the "Push to Jolli" button in the still-open summary panel for the old commit.
4. Wait a moment for the action to process.
5. Check whether a warning message appears in the bottom-right corner of VS Code.
6. Check whether the summary panel closes automatically.

**Expected Results:**
- A warning notification should appear saying the commit was "rewritten into" a newer commit hash, and that you should open that commit's summary instead.
- The warning should mention the operation "push to Jolli".
- The panel for the old (amended) commit should close on its own.
- No data should be written — the old commit's summary should not gain a Jolli document ID or URL, and the new commit's summary should be unaffected.

</blockquote>
</details>
<details>
<summary><strong>4. [7f71000] Stale panel blocked for destructive actions (delete memory, delete E2E scenario) before confirmation dialog</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a Git repository open with a commit summary that contains at least one memory entry and one E2E test scenario. Keep the summary panel open, then amend the commit in the terminal so the panel becomes stale.

**Steps:**
1. With the stale summary panel still open (after the commit was amended), click the delete button next to one of the memory entries.
2. Check whether a "Delete memory?" confirmation dialog appears, or whether a warning notification appears instead.
3. Dismiss any notification, then click the delete button next to one of the E2E test scenarios.
4. Again check whether a confirmation dialog appears or a warning notification appears instead.

**Expected Results:**
- For both delete actions, a warning notification should appear saying the commit was rewritten, mentioning "delete memory" or "delete E2E scenario" respectively.
- No confirmation dialog ("Delete memory?" or "Delete scenario…?") should appear at any point — the guard should stop the action before asking the user to confirm.
- Nothing should be deleted; the summary data should remain unchanged.

</blockquote>
</details>
<details>
<summary><strong>5. [7cba9e2] Race-window block: AI generation (E2E test guide) after commit is rewritten mid-operation</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a commit open in the Jolli sidebar panel. Arrange a way to amend the commit (e.g. stage a small change ready to go) so you can trigger an amend while the AI is working.

**Steps:**
1. Open the Jolli sidebar and navigate to the commit's summary panel.
2. Click the button to generate an E2E test guide for the commit.
3. While the AI generation is in progress (before it finishes), amend the commit in your terminal or via the source control sidebar (e.g. run `git commit --amend --no-edit`).
4. Wait for the AI generation to finish.
5. Check whether a warning message appears in the panel and whether the E2E test guide was saved.

**Expected Results:**
- A warning message should appear indicating the commit was rewritten (mentioning the new commit hash).
- The E2E test guide should NOT be saved to the panel — no new test guide content should appear.
- The panel should not crash or show a silent failure; the warning should be clearly visible.

</blockquote>
</details>
<details>
<summary><strong>6. [7cba9e2] Race-window block: deleting a memory topic after commit is rewritten while confirm dialog is open</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a commit open in the Jolli sidebar panel that has at least one memory topic listed. Arrange a way to amend the commit while the confirmation dialog is open.

**Steps:**
1. Open the Jolli sidebar and navigate to a commit that has at least one memory topic.
2. Click the delete button next to a memory topic — a confirmation dialog should appear asking you to confirm the deletion.
3. While the confirmation dialog is still open, amend the commit in your terminal (e.g. `git commit --amend --no-edit`).
4. Click "Delete" to confirm the deletion in the dialog.
5. Check whether a warning message appears and whether the memory topic was actually removed.

**Expected Results:**
- A warning message should appear indicating the commit was rewritten.
- The memory topic should NOT be deleted — it should still appear in the panel unchanged.
- No silent data loss should occur.

</blockquote>
</details>
<details>
<summary><strong>7. [7cba9e2] Race-window block: adding a plan after commit is rewritten while the plan picker is open</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a commit open in the Jolli sidebar panel. Have at least one plan available to add. Arrange a way to amend the commit while the plan selection menu is open.

**Steps:**
1. Open the Jolli sidebar and navigate to a commit's summary panel.
2. Click the button to add a plan — a selection menu (QuickPick) should appear listing available plans.
3. While the selection menu is still open, amend the commit in your terminal (e.g. `git commit --amend --no-edit`).
4. Select a plan from the menu to confirm your choice.
5. Check whether a warning message appears and whether the plan was added to the commit.

**Expected Results:**
- A warning message should appear indicating the commit was rewritten.
- The plan should NOT be added to the commit — the plans list should remain unchanged.
- No plan should be incorrectly linked to the old (orphaned) commit.

</blockquote>
</details>
<details>
<summary><strong>8. [7cba9e2] Race-window block: translating a plan after commit is rewritten during the AI translation call</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a commit open in the Jolli sidebar panel that has at least one plan written in a non-English language. Arrange a way to amend the commit while the translation is in progress.

**Steps:**
1. Open the Jolli sidebar and navigate to a commit that has a plan with non-English content.
2. Click the translate button for that plan to start AI translation.
3. While the translation is in progress, amend the commit in your terminal (e.g. `git commit --amend --no-edit`).
4. Wait for the translation to finish.
5. Check whether a warning message appears and whether the translated title was saved to the commit summary.

**Expected Results:**
- A warning message should appear indicating the commit was rewritten.
- The translated plan title should NOT be synced back to the commit summary — the summary should remain unchanged.
- The translation itself may be saved to the plan's content file (this is expected and intentional), but the commit summary should not be updated.

</blockquote>
</details>

## Topics (6)
<details>
<summary><strong>01 · [7f71000] Stale-commit guard reads index through bridge for correct storage backend</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The guard that prevents writing to a commit that has since been rewritten was reading the commit index directly, bypassing the layer that knows which storage backend is active. In folder-storage mode this caused the guard to read from the wrong location, meaning real rewrites could go undetected or false blocks could fire.

- **Route through the bridge rather than expose storage directly**: The bridge already owns the authoritative storage instance for all webview-driven writes. Keeping the index read on the same path as every write means the guard and the writers always see the same backend. Exposing the storage object directly to the panel would have created a second ownership point and risked the two drifting apart in future refactors.
- **Add a dedicated regression-guard test**: The fix is a routing change rather than a behavior change, so a normal functional test would not catch a future developer reverting to the direct import. A test that explicitly asserts the bridge method was called makes the storage-threading contract visible and machine-checked.

**✅ What Was Implemented**

- Added `getSummaryIndexEntryMap()` to `JolliMemoryBridge.ts`: this wrapper retrieves the active storage instance (dual-write or folder) and passes it through to the underlying index read, ensuring the correct file is consulted regardless of which storage mode is configured.
- Updated `ensureCommitNotRewritten` in `SummaryWebviewPanel.ts` to call `this.bridge.getSummaryIndexEntryMap()` instead of importing and calling the store function directly. The direct import of `getIndexEntryMap` from `SummaryStore` was removed from the panel entirely.
- Updated test infrastructure in both `SummaryWebviewPanel.test.ts` and `SummaryWebviewPanel.handlePush.test.ts`: removed the `SummaryStore` mock for `getIndexEntryMap`, added `getSummaryIndexEntryMap` as a stub on the bridge object, and added a regression-guard test that explicitly asserts the bridge method is called rather than the direct import.

**📁 FILES**
- `vscode/src/JolliMemoryBridge.ts`

</blockquote>
</details>
<details>
<summary><strong>02 · [7f71000] Stale-commit guard extended to cover plan, note, and Linear issue write paths</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The guard that blocks writes to a rewritten commit only covered push, topic and recap editing, and end-to-end test operations. Plan management, note management, and Linear issue removal could still write data tied to an orphaned commit, silently associating new content with a commit hash that no longer exists as a live root.

- **Block before any user interaction, not after**: For handlers that open a picker or confirmation dialog, the guard fires first. Letting the user complete a multi-step flow only to reject it at the final write would be confusing and wasteful; blocking at the entry point keeps the experience consistent with the existing push and delete-topic guards.
- **Treat save and translate as writes against the stale commit**: Plan and note bodies are global files, so one could argue the body write is safe. The decision was to guard the whole handler anyway, because the title-sync step that follows always writes back to the panel's stale summary. Splitting the handler into a safe body-write and a blocked sync would add complexity with no user benefit — the panel is stale and should be treated as read-only in its entirety.

**✅ What Was Implemented**

- Added the stale-commit check to six handlers that were previously unguarded: remove plan, add plan (before the picker opens), add markdown note (before the file dialog opens), save snippet, remove note, and remove Linear issue. Each check fires before any user-facing dialog or archive call so the user is never prompted to complete an action that will be blocked.
- Added the guard to four save and translate handlers whose side effects write back to the panel's current summary: save plan, save note, translate plan, and translate note. Even though plan and note files are stored globally, each of these handlers synchronises the title or content back into the stale summary, so the whole handler is treated as a write against the orphaned commit.
- Added eleven new tests in `SummaryWebviewPanel.test.ts` covering each newly guarded handler, verifying that the warning is shown and that no downstream write (archive, store, or summary update) is called.

**📁 FILES**

</blockquote>
</details>
<details>
<summary><strong>03 · [7f71000] Block all writes to summary panels whose commit has been rewritten</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A user reported that after amending a commit, the "Push to Jolli" document ID and URL disappeared from the summary. Investigation revealed the panel for the old commit stayed open after the amend, and when the user clicked "Push to Jolli" the write went to the orphaned commit rather than the live one, leaving the current commit's summary without those fields.

- **Index-based detection over git reachability check**: Using the index's `parentCommitHash` field was chosen over spawning a `git branch --contains` subprocess. The index check is a pure in-memory lookup against data jollimemory already maintains, while the git approach would add subprocess latency and could produce false positives when branches are deleted or switched. Critically, the index also carries the identity of the live root, enabling the warning message to tell the user exactly where to go.
- **Guard placement before confirmation dialogs on destructive operations**: For delete handlers, the stale-commit check was placed before the "are you sure?" dialog. Asking a user to confirm a deletion and then silently doing nothing would be confusing and erode trust; surfacing the stale-commit warning first makes the failure reason immediately clear and avoids a misleading interaction.
- **Dispose the panel on detection**: When a stale commit is detected the panel is disposed rather than just blocking the single operation. Keeping the panel open would leave the user with a tab that rejects every action with no clear path forward. Disposing it forces the user to open the current commit's summary from the branch panel, which is the correct recovery path.

**✅ What Was Implemented**

- Added `ensureCommitNotRewritten` to `SummaryWebviewPanel.ts`, a pre-write guard that reads the summary index through the bridge and checks whether the panel's commit hash has a non-null `parentCommitHash` entry — the marker jollimemory writes when a commit is folded into a new root by amend, squash, or rebase. If the commit has been rewritten, the guard walks the parent chain to find the live root, shows a warning naming both the stale hash and the live root hash, disposes the panel, and returns false so the caller short-circuits immediately.
- Wired the guard into ten write-path handlers in the initial implementation: Push to Jolli, edit memory, edit recap, generate recap, delete memory, generate E2E test guide, edit E2E test guide, edit E2E scenario, delete E2E scenario, and delete E2E test guide. For destructive operations the guard fires before the confirmation dialog so users are never asked to confirm an action that cannot take effect.
- Added thirteen new tests in `SummaryWebviewPanel.test.ts` and `SummaryWebviewPanel.handlePush.test.ts` covering: root entry allowed through, absent-from-index allowed through, single-hop rewrite blocked, multi-hop chain walked to final root, cyclic chain broken safely, and per-handler operation label verification.

**📁 FILES**

</blockquote>
</details>
<details>
<summary><strong>04 · [7cba9e2] Guard against stale commit writes after long async operations</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A reviewer flagged that the safety check preventing writes to outdated commits only ran once at the start of each action. During slow operations like AI generation (30–60 seconds) or while a user interacts with a confirmation dialog or selection menu, the underlying commit could be rewritten. When the operation finished, data would silently be saved to the now-orphaned old commit instead of the current one.

- **Selective second-guard placement**: The re-check was added only to handlers with a meaningful async gap — AI calls (tens of seconds) and user-facing dialogs (unbounded wait). Handlers that only do millisecond-level file I/O were left with just the entry guard, keeping the change focused and avoiding unnecessary complexity.
- **Push handler exception — write through rather than block**: For the push flow, blocking the local save after a successful remote call would be worse than writing to a stale commit. The remote article would exist with no local record, and the next push of the new commit would create a duplicate. Accepting the small race risk on a 1–3 second window is the safer trade-off for users, and the decision is documented in the code for future maintainers.
- **Global content (plan/note body) written regardless of race**: For translate operations, the translated text itself is stored globally and not tied to a specific commit, so that write is allowed to proceed even when the re-check fails. Only the commit-specific metadata sync is blocked, preserving the user's translation work.

**✅ What Was Implemented**

- Added a second safety re-check immediately before the final write in every handler that involves a long wait: `handleGenerateRecap`, `handleGenerateE2eTest`, `handleTranslatePlan`, and `handleTranslateNote` each re-check after the AI call completes; `handleDeleteTopic`, `handleDeleteE2eScenario`, `handleDeleteE2eTest`, `handleRemovePlan`, `handleRemoveNote`, and `handleRemoveLinearIssue` re-check after the confirmation modal closes; `handleAddPlan` and `handleAddMarkdownNote` re-check after the selection UI closes. If the commit has been rewritten in the interim, the operation is aborted and a warning is shown.
- Added a code comment to `handlePush` documenting why it intentionally does NOT get a second re-check: the HTTP call is short (1–3 seconds), and blocking the write after a successful push would leave the newly created remote article without a local record, causing a duplicate article on the next push of the new commit.
- Added four new race-window tests using a `makeRacingEntryMapMock` helper that returns the original commit map on the first call and a rewritten map on subsequent calls, precisely simulating an amend landing between the entry check and the pre-write re-check.

**📁 FILES**

</blockquote>
</details>
<details>
<summary><strong>05 · [7cba9e2] Fix invalid TypeScript types in test fixtures for references</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

A reviewer noted that several test fixtures introduced in a previous session were missing required fields, meaning the tests passed at runtime but would fail a TypeScript type check. This created a gap between what the tests validated and what the actual data model required.

The fixes were purely mechanical — each field was either missing from the type definition or using a name that did not exist on the type. No design trade-offs were involved; the goal was simply to make the test fixtures match the actual data model so that TypeScript type checking passes alongside the runtime tests.

**✅ What Was Implemented**

Fixed four fixture objects in `SummaryWebviewPanel.test.ts`: added `editCount`, `addedAt`, and `updatedAt` to two plan reference fixtures; added `addedAt` and `updatedAt` to two note reference fixtures; replaced a non-existent `capturedAt` field on a Linear issue reference fixture with the correct `referencedAt` and `sourceToolName` fields.

**📁 FILES**
- `vscode/src/views/SummaryWebviewPanel.test.ts`

</blockquote>
</details>
<details>
<summary><strong>06 · [7b5dd42] Replace stale commit panel close with persistent read-only warning mode</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

When a commit had been rewritten by an amend, squash, or rebase, the summary panel was immediately closed and a brief toast notification appeared. The user lost their place, had no way to navigate to the new commit's summary, and the toast disappeared before they could act on it.

- **Keep the panel open instead of closing it**: Closing the tab on a stale detection was disorienting, especially when the user had been waiting on a long-running operation. The existing foreign-repo read-only mode already proved that a panel can remain useful in a locked-down state, so the same pattern was applied here. The user retains their reading context and gets a clear in-panel explanation of what changed.
- **Modal dialog over toast notification**: A toast auto-dismisses in roughly ten seconds and appears in a corner, making it easy to miss during a busy workflow. A modal is center-screen and requires an explicit action, ensuring the user sees the old and new commit hashes and the "Open new commit's summary" button before continuing. The trade-off is a small interruption, which is acceptable for a one-time, high-stakes state change.
- **Reuse the foreign-readonly CSS whitelist rather than a separate stale-specific one**: Both read-only modes share the same concept of "buttons that are safe to show even when writes are blocked." Introducing a second whitelist would require every future button author to remember two attributes. A single `data-foreign-safe` marker covers both modes, and a panel that is simultaneously foreign and stale (a cross-repo summary whose source commit was rewritten) gets both hook classes with no extra logic.

**✅ What Was Implemented**

- `SummaryWebviewPanel.ts` gained a `staleRewrittenInto` field and a new `enterStaleReadonlyMode()` method. On first detection of a rewritten commit, the panel re-renders with the stale state set, then raises a modal dialog (blocking, center-screen) naming the old and new short hashes and the blocked operation. The user can click "Open new commit's summary" to jump directly to the live commit's panel, or dismiss to stay. The `panel.dispose()` call was removed entirely. Subsequent write attempts short-circuit silently on the already-set field without re-raising the modal.
- `SummaryHtmlBuilder.ts` received a new `staleRewrittenInto` option on `BuildHtmlOptions`. When set, the page root element gains a `stale-readonly` hook class (parallel to the existing `foreign-readonly` class) and a persistent yellow warning banner is injected at the top of the page, including a button that posts an `openRewrittenCommit` message back to the extension.
- `SummaryCssBuilder.ts` and `SummaryScriptBuilder.ts` were extended to hide all destructive buttons under the `stale-readonly` class (reusing the existing `data-foreign-safe` whitelist) and to wire the banner button's click event to post the navigation message.

**📁 FILES**
- `vscode/src/views/SummaryHtmlBuilder.ts`
- `vscode/src/views/SummaryCssBuilder.ts`
- `vscode/src/views/SummaryScriptBuilder.ts`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 20, 2026 at 3:58 PM*

### DIFF
--- a/vscode/src/JolliMemoryBridge.ts
+++ b/vscode/src/JolliMemoryBridge.ts
@@ -1653,6 +1653,22 @@ export class JolliMemoryBridge {
 	// `SummaryWebviewPanel` use these instead of importing the SummaryStore /
 	// PlanService / NoteService writers directly.
 
+	/**
+	 * Loads the orphan-branch index entry map via the Bridge's storage
+	 * instance. Threading storage matters in folder-mode (or any non-default
+	 * `storageMode`) because the default `resolveStorage()` falls back to a
+	 * fresh `OrphanBranchStorage` that would read the wrong file. Callers like
+	 * the SummaryWebviewPanel's stale-commit guard rely on the live index, so
+	 * routing through the Bridge keeps the read consistent with every
+	 * webview-driven write.
+	 */
+	async getSummaryIndexEntryMap(): Promise<
+		ReadonlyMap<string, import("../../cli/src/Types.js").SummaryIndexEntry>
+	> {
+		const storage = await this.getStorage();
+		return getIndexEntryMap(this.cwd, storage);
+	}
+
 	/** Writes a summary via the Bridge's DualWriteStorage instance. */
 	async storeSummary(
 		summary: CommitSummary,
--- a/vscode/src/views/SummaryCssBuilder.ts
+++ b/vscode/src/views/SummaryCssBuilder.ts
@@ -1235,6 +1235,64 @@ export function buildCss(): string {
     display: none !important;
   }
 
+  /* Stale-readonly mode.
+     Triggered when the panel's commit has been rewritten by amend / squash /
+     rebase during the panel's lifetime (SummaryWebviewPanel sets it from
+     ensureCommitNotRewritten). Reuses the data-foreign-safe whitelist so
+     both readonly modes share a single set of "always allowed" controls
+     (Copy / Save Markdown / Toggle / copy-hash / the stale banner's
+     Open-new-commit action button). The destructive-message guards in
+     SummaryWebviewPanel handlers already short-circuit the writes; this
+     layer removes the affordance so the user has no clickable trap.
+
+     Unlike foreign-readonly, this mode also renders a persistent banner
+     (.stale-banner) at the top of the page — the rewrite is a state
+     change during the panel's lifetime, so the user needs in-panel
+     context for why their buttons just disappeared. */
+  .page.stale-readonly button:not([data-foreign-safe]) {
+    display: none !important;
+  }
+
+  .stale-banner {
+    display: flex;
+    align-items: center;
+    gap: 12px;
+    margin: 0 0 16px 0;
+    padding: 12px 16px;
+    background: var(--vscode-inputValidation-warningBackground, #5a4a1a);
+    border: 1px solid var(--vscode-inputValidation-warningBorder, #cca700);
+    border-radius: 6px;
+    color: var(--vscode-foreground);
+  }
+  .stale-banner-icon {
+    font-size: 18px;
+    line-height: 1;
+  }
+  .stale-banner-text {
+    flex: 1;
+    line-height: 1.5;
+  }
+  .stale-banner-hash {
+    background: var(--vscode-textCodeBlock-background, rgba(0, 0, 0, 0.2));
+    padding: 1px 6px;
+    border-radius: 3px;
+    font-family: var(--vscode-editor-font-family, monospace);
+    font-size: 0.92em;
+  }
+  .stale-banner-action {
+    background: var(--vscode-button-background);
+    color: var(--vscode-button-foreground);
+    border: none;
+    padding: 6px 14px;
+    border-radius: 4px;
+    cursor: pointer;
+    font-size: 13px;
+    white-space: nowrap;
+  }
+  .stale-banner-action:hover {
+    background: var(--vscode-button-hoverBackground);
+  }
+
 ${buildPrSectionCss()}
 `;
 }
--- a/vscode/src/views/SummaryHtmlBuilder.ts
+++ b/vscode/src/views/SummaryHtmlBuilder.ts
@@ -53,6 +53,25 @@ export interface BuildHtmlOptions {
 	 * the source repo, so this layer does not render an additional banner.
 	 */
 	readonly foreignRepoName?: string | null;
+	/**
+	 * Full 40-char hash of the live root commit when the panel's commit has
+	 * been rewritten by amend / squash / rebase. The `.page` root receives
+	 * a `stale-readonly` hook class (parallel to `foreign-readonly`) so the
+	 * same CSS rule that hides destructive buttons takes effect, and a
+	 * persistent banner is rendered at the top explaining where to go.
+	 *
+	 * The banner displays an 8-char short form for readability while the
+	 * action button's `data-target-hash` attribute carries the FULL hash.
+	 * Navigation needs the full hash because `getSummary()` only resolves
+	 * `commitAliases` for 40-char inputs and throws `AmbiguousHashError`
+	 * on prefix collisions.
+	 *
+	 * The banner is needed (unlike foreign-readonly which relies on the
+	 * tab-title prefix) because the rewrite happens during the panel's
+	 * lifetime — the user already had it open and would otherwise just see
+	 * buttons disappear without knowing why.
+	 */
+	readonly staleRewrittenInto?: string | null;
 }
 
 /**
@@ -65,7 +84,14 @@ export function buildHtml(
 	opts: BuildHtmlOptions = {},
 ): string {
 	const { transcriptHashSet, planTranslateSet, noteTranslateSet, nonce } = opts;
-	const pageClass = opts.foreignRepoName ? "page foreign-readonly" : "page";
+	// Both readonly modes share the same CSS rule that hides destructive
+	// buttons (see SummaryCssBuilder). A panel can in principle be both
+	// foreign AND stale (cross-repo summary whose commit was rewritten in
+	// the source repo) — emit both hook classes so the rule still matches.
+	const pageClasses = ["page"];
+	if (opts.foreignRepoName) pageClasses.push("foreign-readonly");
+	if (opts.staleRewrittenInto) pageClasses.push("stale-readonly");
+	const pageClass = pageClasses.join(" ");
 	const { topics: allTopics, sourceNodes } = collectSortedTopics(summary);
 	const stats = resolveDiffStats(summary);
 	const totalInsertions = stats.insertions;
@@ -84,6 +110,20 @@ export function buildHtml(
 		: "";
 	const nonceAttr = nonce ? ` nonce="${nonce}"` : "";
 
+	// Display short-form (8 chars) for readability; the action button keeps
+	// the full hash in data-target-hash so navigation goes through
+	// getSummary()'s 40-char alias-resolution path rather than the
+	// prefix-scan path that throws on collisions.
+	const staleFullHash = opts.staleRewrittenInto ?? "";
+	const staleShortHash = staleFullHash.substring(0, 8);
+	const staleBannerHtml = opts.staleRewrittenInto
+		? `<div class="stale-banner" role="status">
+  <span class="stale-banner-icon" aria-hidden="true">&#x26A0;&#xFE0F;</span>
+  <span class="stale-banner-text">This commit was rewritten into <code class="stale-banner-hash">${escHtml(staleShortHash)}</code>. This view is read-only; open the new commit's summary to make changes.</span>
+  <button class="stale-banner-action" id="staleOpenNewBtn" data-foreign-safe data-target-hash="${escHtml(staleFullHash)}">Open new commit's summary</button>
+</div>`
+		: "";
+
 	return `<!DOCTYPE html>
 <html lang="en">
 <head>
@@ -95,6 +135,7 @@ ${csp}
 </head>
 <body>
 <div class="${pageClass}">
+${staleBannerHtml}
 ${buildAllConversationsSection(transcriptHashSet)}
 ${buildHeader(summary, totalFiles, totalInsertions, totalDeletions)}
 <hr class="separator" />
--- a/vscode/src/views/SummaryScriptBuilder.ts
+++ b/vscode/src/views/SummaryScriptBuilder.ts
@@ -91,6 +91,19 @@ export function buildScript(): string {
       vscode.postMessage({ command: 'push' });
     });
   }
+
+  // Stale-readonly banner: "Open new commit's summary" button. The live
+  // root hash is rendered into data-target-hash by buildHtml; the
+  // extension routes the message to jollimemory.viewSummary.
+  var staleOpenBtn = document.getElementById('staleOpenNewBtn');
+  if (staleOpenBtn) {
+    staleOpenBtn.addEventListener('click', function() {
+      var hash = staleOpenBtn.getAttribute('data-target-hash');
+      if (hash) {
+        vscode.postMessage({ command: 'openRewrittenCommit', hash: hash });
+      }
+    });
+  }
 ${buildPrSectionScript()}
 
   // Listen for messages from the extension (push + topic edit status updates).
--- a/vscode/src/views/SummaryWebviewPanel.handlePush.test.ts
+++ b/vscode/src/views/SummaryWebviewPanel.handlePush.test.ts
@@ -458,6 +458,9 @@ const stubBridge = {
 						artifacts,
 					) as Promise<void>),
 	),
+	// Stale-commit guard reads the index via this bridge wrapper. Default to
+	// "empty index" so push tests stay on the happy path.
+	getSummaryIndexEntryMap: vi.fn().mockResolvedValue(new Map()),
 } as unknown as import("../JolliMemoryBridge.js").JolliMemoryBridge;
 
 // ── Tests ────────────────────────────────────────────────────────────────────
--- a/vscode/src/views/SummaryWebviewPanel.test.ts
+++ b/vscode/src/views/SummaryWebviewPanel.test.ts
@@ -551,6 +551,11 @@ const stubBridge = {
 	archiveNoteForCommit: vi.fn((id: string, commitHash: string) =>
 		mockArchiveNoteForCommit(id, commitHash, workspaceRoot),
 	),
+	// Stale-commit guard reads the index via this bridge wrapper (storage
+	// threading — see JolliMemoryBridge.getSummaryIndexEntryMap docstring).
+	// Default to "empty index" so existing tests are unaffected; the
+	// stale-commit describe block overrides per test.
+	getSummaryIndexEntryMap: vi.fn().mockResolvedValue(new Map()),
 } as unknown as import("../JolliMemoryBridge.js").JolliMemoryBridge;
 const mainBranch = "main";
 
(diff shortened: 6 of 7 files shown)


sanshizhang-jolli:
First version: 
image

Final version:
image

image

flyerli2025 (dismissed):

foster-han (approved):

PR #131, opened 2026-05-21
Fix nested-article rendering and root index flash for site generation

summerfang-jolli:
## Summary

Validated the schema in `cli/sites-schema_0514.md` against a real sample
site (`/Users/jianbinfang/jolli.sites/sample-site`) and fixed the gaps
that surfaced. Six bugs in one commit:

- A navigation article with both an `href`-matching source file and
  nested `articles` produced a Nextra v4 layout conflict — Nextra
  collapsed the parent into a non-clickable expandable folder, the
  children 404'd, and an auto-discovered "index" appeared as a duplicate
  child. `ContentPlanner` now re-homes the parent to
  `<folder>/index.<ext>` and injects `asIndexPage: true` into the
  frontmatter so Nextra promotes it to the folder's representative page.
- The schema-documented `expanded: true` field was silently ignored.
  `StructureParser` now emits `theme.collapsed: false` on the parent's
  `_meta` entry so the folder defaults to open.
- The `asIndexPage` promotion exposed a `MetaGenerator` bug: it kept
  writing `{ "index": { display: "hidden" } }` for promoted indexes,
  causing Nextra to throw _"field key 'index' refers to a page that
  cannot be found"_ at build time. `MetaGenerator` now reads the index's
  frontmatter and omits the `"index"` key when `asIndexPage` is set.
- When the navigation is in page mode and no page is rooted at `/`, the
  auto-mirrored root index used to render briefly before `SidebarTabs`
  client-side redirected to the first tab. The root index is now
  replaced with an inline-`<script>` redirect stub written before
  `MetaGenerator` runs so the file is hidden from the sidebar. The
  inline script encodes `<` as `<` so a hostile href cannot
  break the surrounding `<script>` tag at HTML parse time; the
  noscript anchor uses HTML attribute escaping.

## Test plan

- [x] `npm run all` passes (statements 98.67% / branches 96.08% / funcs
  99.04% / lines 98.97% — all above CLI thresholds)
- [x] Added regression tests:
  - `ContentPlanner.test.ts` — parent article with nested children
    re-homes to `<folder>/index.<ext>`; non-index source gets
    `asIndexPage: true` injected, real index sources are not touched
  - `StructureParser.test.ts` — `expanded: true` + nested children emits
    `theme.collapsed: false`; missing/false `expanded` keeps the plain
    string entry
  - `MetaGenerator.test.ts` — `_meta.js` omits the `"index"` key when
    the index has `asIndexPage: true`
  - `StartCommand.test.ts` — `pickRootRedirectHref` skips menu pages
    and respects user-claimed roots; `writeRootRedirectIndex` escapes
    `</script>` in JS context and HTML-escapes the noscript anchor
- [x] Manually rebuilt the sample site (`jolli build`): 16 pages, no
  Nextra validation errors. Verified in browser (incognito) that
  Deployment renders content + defaults to expanded; Docker/Kubernetes
  resolve; visiting `/` redirects to `/docs` without the prior flash.


Finn-coder2026:
[Uploading jolli-pr131-test.zip…]()
Verified the fix on a sample site: menu entries can be clicked to render their own markdown content, and they also expand/collapse to reveal nested children — both jobs work as expected.

One observation: `expanded: false` doesn't appear to take effect. Since either value (`true` or `false`) still lets users manually toggle a menu open or closed, my mental model is that this field is meant to control the *default* open/collapsed state. If that's the intent, would a name like `defaultExpanded` (or `defaultOpen`) be clearer than `expanded`?

cli/src/commands/StartCommand.ts
```diff
@@ -243,6 +257,65 @@ async function syncContent(
...
+ * Exported for unit testing — call site is in `syncContent`.
+ */
+export function pickRootRedirectHref(
+	parsedNavigation: ReturnType<typeof parseNavigation> | undefined,
+): string | undefined {
+	if (!parsedNavigation?.pages?.length) return undefined;
+	const claimsRoot = parsedNavigation.pages.some((p) => p.href === "/");
+	if (claimsRoot) return undefined;
+	const firstNavigable = parsedNavigation.pages.find((p) => p.type !== "menu" && p.href && p.href !== "#");
+	return firstNavigable?.href;
```
Finn-coder2026: The selected `href` is forwarded straight to `window.location.replace(...)` and rendered into a `<a href="...">` without any scheme check. Since the value originates in `site.json`, an entry such as `nav: [{ root: "javascript:alert(1)" }]` would currently execute as JS on every visit to `/`.

The repo already has the canonical allow-list sanitiser in `cli/src/site/Sanitize.ts` — `_meta.js` / footer / layout writers route every user-supplied URL through it for exactly this reason. Suggest:

```ts
const safe = sanitizeUrl(firstNavigable.href);
return safe === "#" ? undefined : safe;
```

With that, a malicious href becomes `"#"` and `pickRootRedirectHref` simply returns `undefined`, leaving the original index in place.

cli/src/commands/StartCommand.ts
```diff
@@ -243,6 +257,65 @@ async function syncContent(
...
+	// escaping is different from JS string escaping — a JSON-encoded value
+	// dropped into an HTML attribute could still let a `</script>` etc. land
+	// in the document as raw markup once a hostile href passes through.
+	const htmlHref = escapeHtml(href);
+	const mdx = `---
+title: Redirecting…
+sidebarTitle: ' '
+---
+
+<script dangerouslySetInnerHTML={{ __html: \`window.location.replace(${jsHref})\` }} />
```
Finn-coder2026, line 340: The inline script body is a **JavaScript template literal**: `` `window.location.replace(${jsHref})` ``. `JSON.stringify` does not escape backticks or `${`, and your `\u003c` step only neutralises `</script>`. So a hostile href containing `${...}` (e.g. `/x${alert(1)}`) survives JSON encoding, lands inside the template literal at MDX render time, and the `${alert(1)}` is evaluated as a template substitution — i.e. arbitrary JS executes on every visit to `/`. A trailing backtick is the matching DoS variant (page crash on syntax error).

Two simple fixes, pick one:

```ts
// Option A — build the script body as a normal string and JSON-encode the whole thing once.
const scriptBody = `window.location.replace(${jsHref})`;
const innerHtml = JSON.stringify(scriptBody).replace(/</g, "\\u003c");
// …__html: ${innerHtml}…

// Option B — extend the post-JSON.stringify sanitiser to cover template-literal syntax.
const jsHref = JSON.stringify(href)
  .replace(/</g, "\\u003c")
  .replace(/`/g, "\\u0060")
  .replace(/\$\{/g, "\\${");
```

The `writeRootRedirectIndex` test below proves `</script>` injection is blocked but not backtick / `${...}` injection — please add a regression case using `/x${alert(1)}` (and one with a literal backtick) so this is locked in.

cli/src/commands/StartCommand.ts
```diff
@@ -243,6 +257,65 @@ async function syncContent(
...
+
+<noscript>
+  Redirecting to <a href="${htmlHref}">${htmlHref}</a>…
```
Finn-coder2026, line 343: `htmlHref` is interpolated as **MDX text content** between `<a>` and `</a>` (and again before `…`). MDX parses `{...}` in text content as a JSX expression, so an href like `/foo{alert(1)}` lands as `Redirecting to <a href="...">/foo{alert(1)}</a>…` and the `{alert(1)}` runs at render time. The local `escapeHtml` defined below doesn't escape `{` or `}`, so it doesn't defuse this.

`cli/src/site/Sanitize.ts` already exports an `escapeHtml` that does the right thing for both attribute and JSX-text contexts (`&`, `<`, `>`, `"`, `'`, `{`, `}`). Please import that one and delete the local copy below.

Finn-coder2026 (commented):
Nice tight regression PR — the four root causes are well isolated and the test coverage on the planner / meta-generator paths is solid. Most of my comments are clustered around the new redirect-stub writer in `StartCommand.ts`, which has a couple of XSS / template-literal injection issues that the current test only partially covers, plus a few suggestions to lean on existing helpers (`sanitizeUrl` / `escapeHtml` in `cli/src/site/Sanitize.ts`) instead of reimplementing weaker local copies. The frontmatter regex used by both `ContentPlanner` and `MetaGenerator` is also worth tightening — it currently matches at any indentation level so a nested YAML key falsely registers as `asIndexPage: true`.

Nothing here blocks the core fix; treat them as items to fold in before merge.

cli/src/commands/StartCommand.ts
```diff
@@ -243,6 +257,65 @@ async function syncContent(
...
+</noscript>
+`;
+	await mkdir(contentDir, { recursive: true });
+	await rm(join(contentDir, "index.md"), { force: true });
+	await writeFile(join(contentDir, "index.mdx"), mdx, "utf-8");
+}
+
+function escapeHtml(s: string): string {
+	return s.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;").replace(/"/g, "&quot;");
+}
```
Finn-coder2026: Drop this helper and import `escapeHtml` from `../site/Sanitize.js`. The Sanitize copy is documented as the single source of truth (see the module docstring — previous duplicates were specifically consolidated there) and is strictly stronger than this one: it also escapes `'`, `{`, and `}`, which is what makes the noscript text on line 307 actually safe in MDX.

cli/src/site/ContentPlanner.ts
```diff
@@ -387,11 +395,39 @@ async function rewritePlannedPage(
...
+ * the file has no frontmatter, prepends a new block. If the frontmatter
+ * already declares `asIndexPage`, leaves it as authored.
+ */
+function injectAsIndexPageFrontmatter(content: string): string {
+	const fmMatch = content.match(/^---\r?\n([\s\S]*?)\r?\n---\r?\n?/);
+	if (!fmMatch) {
+		return `---\nasIndexPage: true\n---\n${content}`;
+	}
+	const body = fmMatch[1];
+	if (/^\s*asIndexPage\s*:/m.test(body)) {
```
Finn-coder2026: The `^\s*asIndexPage\s*:/m` regex matches at **any** indentation, so a nested key like

```yaml
things:
  asIndexPage: true
```

falsely registers as "the page already declares `asIndexPage`" and the function skips injection — the resulting file has no top-level flag and Nextra v4 falls back to its layout-conflict behaviour. Anchoring the pattern to column 0 would fix it:

```ts
if (/^asIndexPage\s*:/m.test(body)) {
  return content;
}
```

Two broader notes while you're here:

1. The exact same fragile YAML-detection pattern lives in `MetaGenerator.detectAsIndexPage` (see linked comment). Worth extracting to a shared helper — both functions are answering the same question, and they currently disagree on what counts as "declared" (one matches any value, the other only `true`).
2. Consider using a real YAML parser. `js-yaml` is already in many of the site-gen deps; one-line `yaml.load(body).asIndexPage === true` would be both safer and clearer than two slightly-different regexes.

cli/src/site/MetaGenerator.ts
```diff
@@ -180,6 +197,26 @@ function buildOverriddenEntries(filenames: string[], override: Record<string, Si
...
+	if (!indexFile) return false;
+	let raw: string;
+	try {
+		raw = await readFile(join(dir, indexFile), "utf-8");
+	} catch {
+		return false;
+	}
+	const fm = raw.match(/^---\r?\n([\s\S]*?)\r?\n---/);
+	if (!fm) return false;
+	return /^\s*asIndexPage\s*:\s*true\s*$/m.test(fm[1]);
```
Finn-coder2026: Same fragility / duplication as `ContentPlanner.injectAsIndexPageFrontmatter` — see the longer comment there. In short: the `^\s*` allows any indentation so a nested key falsely triggers, and the regex only accepts lowercase `true` (YAML also accepts `True`, `TRUE`, `yes`).

Also: the order-of-preference for picking which `index.*` to read is `contentItems` array order, which is `readdir` order — typically alphabetical, so `index.md` wins over `index.mdx`. Nextra prefers `.mdx` in case of a tie; if both exist the detector can disagree with what Nextra actually compiles. Prefer `.mdx` explicitly when both are present.

cli/src/commands/StartCommand.test.ts
```diff
@@ -1110,3 +1110,99 @@ describe("buildRootInjectionInput", () => {
...
+		const { tmpdir } = await import("node:os");
+		contentDir = await mkdtemp(join(tmpdir(), "jolli-redirect-test-"));
+		cleanup = [];
+	});
+
+	afterEach(async () => {
+		const { rm } = await import("node:fs/promises");
+		await rm(contentDir, { recursive: true, force: true });
+		for (const fn of cleanup) await fn();
+	});
```
Finn-coder2026, line 1235: `cleanup` is declared, reset to `[]` in `beforeEach`, and iterated in `afterEach`, but nothing is ever pushed into it — dead code. Either remove the variable + the loop, or wire it up to the cleanup it was meant to track.

Minor while in the area: every `it` block does `await import("./StartCommand.js")` for symbols you could pull in once at the top of the file. Module imports are cached so it's not a correctness issue, just noise — a single `import { pickRootRedirectHref, writeRootRedirectIndex } from "./StartCommand.js"` at the top would do.

cli/src/commands/StartCommand.ts
summerfang-jolli: Fixed in 444e3be — `pickRootRedirectHref` now routes the selected href through `sanitizeUrl` (the canonical allow-list helper from `cli/src/site/Sanitize.ts`). A `javascript:` / `data:` / `vbscript:` / scheme-relative URL collapses to `"#"`, which the caller maps back to `undefined` so the original index renders.

Regression tests added in `StartCommand.test.ts` for `javascript:alert(1)` and `//evil.com/x` inputs to `pickRootRedirectHref`.

cli/src/commands/StartCommand.ts
summerfang-jolli: Fixed in 444e3be — went with Option B (extend the post-`JSON.stringify` sanitiser) since it keeps the existing single-pass shape:

```ts
const jsHref = JSON.stringify(href)
    .replace(/</g, "\\u003c")
    .replace(/`/g, "\\u0060")
    .replace(/\$\{/g, "\\u0024{");
```

Two regression tests added in `StartCommand.test.ts`:

- `/x${alert(1)}` — asserts the inline-script body contains `${alert(1)}` (escaped) and no literal `${alert` survives.
- `/x\`alert(1)\`` — asserts the only backticks on the redirect line are the two template-literal delimiters; the hostile backticks are encoded as ```.

cli/src/commands/StartCommand.ts
summerfang-jolli: Fixed in 444e3be — dropped the local `escapeHtml` and imported the canonical one from `cli/src/site/Sanitize.ts`, which escapes `& < > " ' { }` (so the `{alert(1)}` JSX-expression injection in the noscript text is defused along with everything else).

Regression test added: `writeRootRedirectIndex` called with `/foo{alert(1)}` — the noscript block contains `&#123;alert(1)&#125;` and no literal `{alert(1)}`.

cli/src/commands/StartCommand.ts
summerfang-jolli: Done in 444e3be — local helper removed, `escapeHtml` imported from `../site/Sanitize.js`. Sanitize.ts is the single source of truth as documented in its module header.

cli/src/site/ContentPlanner.ts
summerfang-jolli: Fixed in 444e3be — anchored the detection regex to column 0:

```ts
if (/^asIndexPage\s*:/m.test(body)) {
    return content;
}
```

Regression test added in `ContentPlanner.test.ts`: a source file with frontmatter `things:\n  asIndexPage: true` now correctly gets a top-level `asIndexPage: true` injected, and the original nested key is preserved.

On the two broader notes:

- **Shared helper between ContentPlanner and MetaGenerator**: I considered it but kept them separate because they answer different questions — injection asks "is the key present at all so I shouldn't double-write" (so it skips on any value), while detection asks "is it set truthy so Nextra promoted it" (so it only matches truthy values). A single helper would need a mode flag and obscure the intent. Anchoring both regexes the same way removes the false-positive divergence without forcing them to share a body.
- **Real YAML parser**: `yaml` is already a CLI dependency, but adding `yaml.load(body).asIndexPage === true` would mean handling parse errors (silent fallback vs. failing the build) and re-deriving the same anchored truthy-match semantics. I'd rather do that as a separate cleanup if/when the regex is no longer expressive enough — happy to revisit.

cli/src/site/MetaGenerator.ts
summerfang-jolli: Fixed in 444e3be. Three changes to `detectAsIndexPage`:

1. Anchored the regex to column 0 so a nested YAML key no longer falsely registers.
2. Extended the truthy match to the YAML 1.1 variants gray-matter recognises (`true | True | TRUE | yes | Yes | YES`).
3. Replaced the `find()` over `contentItems` with an explicit `index.mdx`-first lookup so the detector agrees with Nextra's compilation order when both `.mdx` and `.md` exist.

Two regression tests added in `MetaGenerator.test.ts`:

- A nested `asIndexPage: true` inside a top-level `things:` key keeps the index hidden via the standard `display: "hidden"` shape (does NOT promote).
- With both `index.mdx` (carrying `asIndexPage: true`) and `index.md` (without it) present, the index key is omitted — i.e. the `.mdx` wins.

(Per the longer reply on `ContentPlanner.ts`: deliberately kept the two regexes separate rather than extracting a shared helper, since they answer different questions.)

cli/src/commands/StartCommand.test.ts
summerfang-jolli: Removed in 444e3be — `cleanup` and its `for` loop are gone.

On the minor `await import(...)` inside each `it`: left as-is for now. The dynamic-import pattern is consistent with every other `describe` in this file (they all rely on the module-level `vi.mock` setup running before the import lands), and switching just the `writeRootRedirectIndex` block to top-level imports would be inconsistent without a broader refactor. Happy to do that pass separately if you want me to.

summerfang-jolli:
Thanks for the verification, Finn — good catch on `expanded: false`.

You're right that the current implementation only emits a `theme.collapsed: false` override when `expanded === true`; `expanded: false` is treated the same as omitted (falls back to Nextra's `defaultMenuCollapseLevel`). So today the field only really has one effective value, which matches your mental model that it's meant to express the default state.

I held off on the rename in 444e3be because that commit is the security/correctness pass on the review comments and a field rename is a separate schema decision. Two options for the follow-up — happy with either, slight preference for (1):

1. **Rename to `defaultExpanded` and honor both values.** When `defaultExpanded: false` is set on an article with nested children, emit `theme.collapsed: true` so the user can also force-collapse a folder that Nextra would otherwise leave open at shallow depths. This makes the field do useful work in both states and matches what the name implies.
2. **Keep `expanded` but document the asymmetry.** Note in the schema doc that `expanded: true` opens by default and `expanded: false` is a no-op (defers to Nextra). Cheaper, but the field name still misleads.

I'll send (1) as a separate PR on top of this branch unless you'd rather see it folded in here before merge — let me know which you prefer.

Finn-coder2026 (dismissed):

Finn-coder2026 (approved):

Finn-coder2026 (approved):

Finn-coder2026 (approved):

PR #143, opened 2026-05-23
Replace tar dependency with built-in tar.gz parser

summerfang-jolli:
## Summary
- Replaced the `tar` npm package with a built-in tar.gz extractor using `node:zlib` (gunzipSync) and manual tar header parsing in `ThemeCommand.ts`
- Removed `tar` from `cli/package.json` dependencies and `cli/vite.config.ts` externals
- Fixes `ERR_MODULE_NOT_FOUND: Cannot find package 'tar'` when running `jolli dev` from a user project directory

## Why
The `tar` package was marked as external in the Vite build config, so it was not bundled into `dist/Cli.js`. When customers ran `jolli dev` in their own project, Node.js could not find `tar` in their `node_modules`. Rather than bundling `tar` (which may have bundler compatibility issues), we replaced it with a zero-dependency tar.gz parser that uses only Node.js built-ins — works on all platforms including Windows.

## Test plan
- [x] `npm run all` passes (build, lint, all 6764 tests, coverage thresholds met)
- [ ] Manual: `jolli theme install <name>` downloads and extracts a theme correctly
- [ ] Manual: `jolli dev` no longer crashes with `ERR_MODULE_NOT_FOUND`

## Quick recap

Theme downloads in the CLI no longer depend on any third-party archive library. Previously, the CLI crashed on startup — even for commands completely unrelated to themes — because a missing archive package caused Node.js to abort before any command could run. The fix removes that package entirely and replaces it with a compact archive parser built from Node.js built-in modules, so the CLI works correctly on Windows, macOS, and Linux with no additional installation requirements.

Theme files are still fetched from the same GitHub endpoint as before, in a single network request with no rate limiting. The new parser processes the downloaded archive directly in memory, extracts only the files belonging to the requested theme, and writes them to disk with path traversal protection. Users installing or updating themes will see no change in behavior — the command works the same way, just reliably across all platforms.

---

## Topic (1)
<details>
<summary><strong>01 · Replace third-party tar package with built-in archive parser for theme downloads</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The CLI crashed on every command — including unrelated ones like starting a dev server — with a "cannot find package 'tar'" error. The root cause was that the tar library was listed as an external dependency not bundled into the CLI, so it was unavailable in users' project directories. Even if bundled, tar support on Windows is unreliable.

- **Zero external dependencies over convenience**: The team considered keeping the `tar` npm package (either as a bundled or external dependency) and also evaluated the GitHub Contents API as a replacement. The `tar` package was rejected on cross-platform grounds — Windows support is fragile — and the Contents API was rejected due to a 60-requests-per-hour unauthenticated rate limit that would break theme installs in CI or for active users. Writing a minimal parser using only built-in Node.js APIs eliminates both risks with a small, auditable implementation.
- **Single tarball download over per-file API calls**: The `codeload.github.com` tarball endpoint delivers the entire repository in one HTTP request with no rate limiting, which is the same URL the previous `tar`-based approach used. This was preserved deliberately — the only thing that changed is how the downloaded bytes are processed, not where they come from.
- **In-memory buffering over streaming extraction**: The previous implementation streamed the tarball response directly into the `tar` library's extraction pipeline. The new approach buffers the full response into memory before parsing. Theme tarballs are small enough that memory pressure is not a concern, and buffering simplifies the implementation significantly by allowing two synchronous passes over the data.

**✅ What Was Implemented**

- Removed the `tar` npm package entirely from `cli/package.json` and the Vite external list in `cli/vite.config.ts`. The dependency and all its transitive packages (`@isaacs/fs-minipass`, `minizlib`, `yallist`, `chownr`) were removed from the lockfile.
- Added a self-contained tar.gz parser (~80 lines) directly in `cli/src/commands/ThemeCommand.ts` using only Node.js built-ins. The implementation uses `node:zlib`'s `gunzipSync` to decompress the gzip layer, then manually walks 512-byte tar headers to extract file name, size, type flag, and prefix fields. Path traversal attacks are blocked by normalizing each output path and verifying it stays within the destination directory.
- `downloadTheme()` in `ThemeCommand.ts` was rewritten to buffer the full tarball response via `arrayBuffer()`, run it through the new `extractTarGz()` function twice (once to detect the repo root directory name, once filtered to the requested theme subdirectory), then write files directly to the destination — eliminating the temp-directory-and-copy pattern used with the old `tar` library.

**📁 FILES**
- `cli/src/commands/ThemeCommand.ts`
- `cli/vite.config.ts`
- `cli/package.json`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 23, 2026 at 10:56 AM · via Anthropic*

### DIFF
--- a/cli/src/commands/ThemeCommand.ts
+++ b/cli/src/commands/ThemeCommand.ts
@@ -8,13 +8,11 @@
  */
 
 import { existsSync } from "node:fs";
-import { cp, mkdir, mkdtemp, readdir, readFile, rm, writeFile } from "node:fs/promises";
-import { homedir, tmpdir } from "node:os";
-import { join } from "node:path";
-import { Readable } from "node:stream";
-import { pipeline } from "node:stream/promises";
+import { mkdir, readFile, rm, writeFile } from "node:fs/promises";
+import { homedir } from "node:os";
+import { join, normalize } from "node:path";
+import { gunzipSync } from "node:zlib";
 import type { Command } from "commander";
-import { extract } from "tar";
 import { scaffoldProject } from "../site/StarterKit.js";
 
 // ─── Constants ───────────────────────────────────────────────────────────────
@@ -106,6 +104,83 @@ async function fetchRegistry(): Promise<RegistryData> {
 	return (await res.json()) as RegistryData;
 }
 
+// ─── Minimal tar.gz extractor (zero dependencies) ──────────────────────────
+//
+// tar format: each entry = 512-byte header + ceil(size/512)*512 bytes of data.
+// Header fields are NUL-terminated octal ASCII strings at fixed offsets.
+// We only need: name (0, 100), size (124, 12), typeflag (156, 1), prefix (345, 155).
+
+interface TarEntry {
+	path: string;
+	type: "file" | "dir";
+	data: Buffer;
+}
+
+/** Parse a NUL-terminated octal string from a tar header field. */
+function parseOctal(buf: Buffer, offset: number, length: number): number {
+	const slice = buf.subarray(offset, offset + length);
+	const str = slice.toString("ascii").replace(/\0.*$/, "").trim();
+	return str.length === 0 ? 0 : Number.parseInt(str, 8);
+}
+
+/** Read a NUL-terminated ASCII string from a tar header field. */
+function parseString(buf: Buffer, offset: number, length: number): string {
+	const slice = buf.subarray(offset, offset + length);
+	const idx = slice.indexOf(0);
+	return (idx >= 0 ? slice.subarray(0, idx) : slice).toString("ascii");
+}
+
+/**
+ * Extracts all entries from a .tar.gz buffer using only `node:zlib`.
+ * Returns entries whose path starts with `prefix/` (if given), with
+ * the prefix stripped so callers get repo-relative paths.
+ */
+function extractTarGz(gzBuf: Buffer, filterPrefix?: string): TarEntry[] {
+	// Decompress gzip → raw tar bytes
+	const tar = gunzipSync(gzBuf);
+
+	const entries: TarEntry[] = [];
+	let offset = 0;
+
+	while (offset + 512 <= tar.length) {
+		const header = tar.subarray(offset, offset + 512);
+		// Two consecutive zero blocks = end of archive
+		if (header.every((b) => b === 0)) break;
+
+		const prefix = parseString(header, 345, 155);
+		const rawName = parseString(header, 0, 100);
+		const fullPath = prefix ? `${prefix}/${rawName}` : rawName;
+		const size = parseOctal(header, 124, 12);
+		const typeflag = header[156];
+
+		offset += 512; // move past header
+
+		// typeflag: 0 or ASCII '0' = file, '5' = directory, others skipped
+		const isFile = typeflag === 0 || typeflag === 0x30; // 0x30 = '0'
+		const isDir = typeflag === 0x35; // '5'
+
+		if ((isFile || isDir) && fullPath.length > 0) {
+			const data = isFile ? Buffer.from(tar.subarray(offset, offset + size)) : Buffer.alloc(0);
+			let entryPath = fullPath;
+			if (filterPrefix) {
+				if (!entryPath.startsWith(filterPrefix)) {
+					offset += Math.ceil(size / 512) * 512;
+					continue;
+				}
+				entryPath = entryPath.slice(filterPrefix.length);
+			}
+			if (entryPath.length > 0) {
+				entries.push({ path: entryPath, type: isFile ? "file" : "dir", data });
+			}
+		}
+
+		// Advance past data blocks (rounded up to 512-byte boundary)
+		offset += Math.ceil(size / 512) * 512;
+	}
+
+	return entries;
+}
+
 /**
  * Downloads a theme to `~/.jolli/themes/<name>/` from the canonical GitHub
  * repo (`github.com/jolliai/themes`). Saves version metadata from the
@@ -113,9 +188,9 @@ async function fetchRegistry(): Promise<RegistryData> {
  * not present in the repo — callers decide whether to fall back to the
  * already-cached copy under `~/.jolli/themes/<name>/`.
  *
- * Uses the public `codeload.github.com` tarball endpoint (one request for
- * the whole repo, no `api.github.com` rate-limit consumption) and extracts
- * only the requested `<name>/` directory into the destination.
+ * Uses the public `codeload.github.com` tarball endpoint (single request,
+ * no API rate-limit) and a built-in tar.gz parser — zero external
+ * dependencies, works on all platforms.
  */
 export async function downloadTheme(name: string, version?: string): Promise<string> {
 	const destDir = join(USER_THEMES_DIR, name);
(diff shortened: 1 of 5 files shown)


cli/src/commands/ThemeCommand.ts
```diff
@@ -128,40 +203,43 @@ export async function downloadTheme(name: string, version?: string): Promise<str
...
 	}
+
+	// Replace the destination so stale files don't linger.
+	await rm(destDir, { recursive: true, force: true });
+
+	for (const entry of themeEntries) {
+		// Guard against path traversal (e.g. `../../etc/passwd`)
+		const dest = normalize(join(destDir, entry.path));
+		if (!dest.startsWith(destDir)) continue;
+
```
Finn-coder2026, line 253: **Potential path-traversal bypass.** `startsWith(destDir)` is a plain string-prefix check, so it doesn't catch a *sibling* directory that happens to share the same prefix.

Example: if `destDir = "/Users/x/.jolli/themes/foo"` and the entry path resolves to `"../foobar/evil.txt"`, the normalized result is `/Users/x/.jolli/themes/foobar/evil.txt`. That path *does* satisfy `startsWith("/Users/x/.jolli/themes/foo")` (because `foo` is a prefix of `foobar`), but it actually lives outside the `foo/` directory.

Safer form requires the separator boundary:

```ts
import { sep } from "node:path";
// ...
if (dest !== destDir && !dest.startsWith(destDir + sep)) continue;
```

Not currently exploitable since the only data source is our own `jolliai/themes` repo, but the comment one line above advertises the check as a path-traversal guard, so it should actually be one.

cli/src/commands/ThemeCommand.ts
```diff
@@ -106,16 +104,93 @@ async function fetchRegistry(): Promise<RegistryData> {
...
+		const typeflag = header[156];
+
+		offset += 512; // move past header
+
+		// typeflag: 0 or ASCII '0' = file, '5' = directory, others skipped
+		const isFile = typeflag === 0 || typeflag === 0x30; // 0x30 = '0'
+		const isDir = typeflag === 0x35; // '5'
+
+		if ((isFile || isDir) && fullPath.length > 0) {
+			const data = isFile ? Buffer.from(tar.subarray(offset, offset + size)) : Buffer.alloc(0);
```
Finn-coder2026: **Unknown typeflags are silently dropped.** Only `\0` / `'0'` (regular file) and `'5'` (directory) are recognized; everything else falls through the `if ((isFile || isDir) && …)` block and the parser just advances past the data block. A few cases worth thinking about:

- **`'L'` (GNU LongName)** — its data block carries the long path for the *next* entry. With this implementation that next entry would be parsed using the truncated 100-byte `name` field (typically `././@LongLink`), so the real filename is lost.
- **`'x'` / `'g'` (PAX extended / global header)** — similar idea; carries long paths and other metadata for the following entry.
- **`'1'` / `'2'` (hardlink / symlink)** — silently skipping these is actually fine for a theme installer (and a bit safer than honoring them), but worth being intentional about.

UStar `prefix(155) + name(100)` covers up to 255 bytes, which should be enough for everything in `jolliai/themes` today. Two reasonable options:

1. Document the constraint in the parser header comment: themes repo must not contain paths > 255 bytes; PAX / GNU long-name entries are not supported.
2. Throw on unknown typeflags so we fail loudly instead of producing a half-extracted theme directory.

I'd slightly prefer (2) — silent drops are very hard to debug after the fact.

cli/src/commands/ThemeCommand.ts
```diff
@@ -128,40 +203,43 @@ export async function downloadTheme(name: string, version?: string): Promise<str
...
+	if (allEntries.length === 0) throw new Error("Empty tarball");
+	const rootDir = allEntries[0].path.split("/")[0];
+	const prefix = `${rootDir}/${name}/`;
 
-		if (version) await writeInstalledMeta(name, version);
-		return destDir;
-	} finally {
-		await rm(tmpRoot, { recursive: true, force: true });
+	const themeEntries = extractTarGz(gzBuf, prefix);
+	if (themeEntries.length === 0) {
```
Finn-coder2026, line 239: **Minor: same buffer is gunzipped + walked twice.** `extractTarGz` is invoked once to discover the root dir name and once again with `filterPrefix` to extract the actual entries, so `gunzipSync(gzBuf)` runs twice on the same input and every header is parsed twice.

For current theme sizes this is negligible, but it scales linearly with archive size. The single-pass version is also a touch simpler:

```ts
const allEntries = extractTarGz(gzBuf);
if (allEntries.length === 0) throw new Error("Empty tarball");
const rootDir = allEntries[0].path.split("/")[0];
const prefix = `${rootDir}/${name}/`;
const themeEntries = allEntries.flatMap((e) => {
    if (!e.path.startsWith(prefix)) return [];
    const relPath = e.path.slice(prefix.length);
    return relPath.length === 0 ? [] : [{ ...e, path: relPath }];
});
```

Bonus: the `filterPrefix` parameter on `extractTarGz` can then be dropped, making the parser smaller and easier to unit-test in isolation. Totally optional, not blocking.

cli/src/commands/ThemeCommand.ts
```diff
@@ -106,16 +104,93 @@ async function fetchRegistry(): Promise<RegistryData> {
...
+	const tar = gunzipSync(gzBuf);
+
+	const entries: TarEntry[] = [];
+	let offset = 0;
+
+	while (offset + 512 <= tar.length) {
+		const header = tar.subarray(offset, offset + 512);
+		// Two consecutive zero blocks = end of archive
+		if (header.every((b) => b === 0)) break;
+
```
Finn-coder2026, line 159: **Nit:** the comment above correctly says "two consecutive zero blocks", but the check only looks at one. POSIX tar terminates on two 512-byte zero blocks; some producers (rare) emit a single zero block in the middle of an archive without it being the end marker.

Unlikely to bite with GitHub `codeload` tarballs in practice, but the code and the comment should at least agree. Either tighten the check to look at two blocks, or change the comment to "a zero block ends the archive" so we're not lying about what we do.

cli/src/commands/ThemeCommand.ts
```diff
@@ -106,16 +104,93 @@ async function fetchRegistry(): Promise<RegistryData> {
...
+	return (idx >= 0 ? slice.subarray(0, idx) : slice).toString("ascii");
+}
+
+/**
+ * Extracts all entries from a .tar.gz buffer using only `node:zlib`.
+ * Returns entries whose path starts with `prefix/` (if given), with
+ * the prefix stripped so callers get repo-relative paths.
+ */
+function extractTarGz(gzBuf: Buffer, filterPrefix?: string): TarEntry[] {
+	// Decompress gzip → raw tar bytes
```
Finn-coder2026: **No tests for the new parser.** `cli/src/commands/ThemeCommand.ts` is in the `coverage.exclude` list in `cli/vite.config.ts` (line 57), which is the only reason CI is green — the new ~80-line tar parser has zero coverage.

This is exactly the kind of code that benefits most from tests: it parses untrusted-by-default binary input from the network, and we're now maintaining it ourselves rather than relying on a battle-tested library. A small fixture-based suite would cover a lot of ground:

- A committed real `.tar.gz` fixture (e.g. under `cli/src/commands/__fixtures__/`) extracts to the expected file tree.
- An entry whose normalized path escapes the destination (`../evil`) is rejected — proves the path-traversal guard works.
- A sibling-prefix entry (`destDir = ".../foo"`, entry path `../foobar/x`) — once the `startsWith` bypass below is fixed, this should also be rejected.
- An empty / single-zero-block tarball: `extractTarGz` returns `[]`, and `downloadTheme` throws "Empty tarball".

Since this PR is fixing a startup-breaking bug, I'd be fine landing it as-is and adding tests in a follow-up — please open a tracking issue so it doesn't get forgotten.

Finn-coder2026 (commented):
Nice fix for the startup crash — pulling `tar` out of the runtime dependencies and using only `node:zlib` is the right call given the Vite-external situation and `tar`'s flaky Windows story. The single-tarball-download-via-codeload approach is preserved verbatim, which keeps the rate-limit characteristics unchanged.

A few things on the new in-house parser worth a second look, most important first:

1. **`downloadTheme` path-traversal guard has a sibling-prefix bypass** — `startsWith(destDir)` matches `"/.../foobar/..."` when `destDir = "/.../foo"`. Inline comment with a fix.
2. **PAX / GNU long-name headers (`'L'`, `'x'`, `'g'`) are silently dropped** — entries with paths > 100 bytes would parse as `././@LongLink` and the real filename would be lost.
3. **Same gzip buffer is decompressed and walked twice** — minor perf nit with a single-pass rewrite suggestion.
4. **Single zero block terminates the archive**, but the comment correctly says it should be two.
5. **The new ~80-line parser has zero unit tests** — `ThemeCommand.ts` is in `coverage.exclude` (`cli/vite.config.ts:57`) so CI doesn't notice.

Also: the two manual test items in the PR description are both still unticked. `npm run all` doesn't exercise the codeload fetch path end-to-end, so could you run `jolli theme install <name>` and `jolli dev` against a fresh `node_modules` and tick them off before merging?

None of this needs to block landing the fix quickly if users are hitting the crash — but at minimum I'd want the path-traversal `startsWith` tightened up, and a follow-up issue for tests.

cli/src/commands/ThemeCommand.ts
summerfang-jolli: Fixed in d8fcb35. Now uses `dest !== destDir && !dest.startsWith(destDir + sep)` which requires the separator boundary. Added a test case that verifies sibling-prefix paths like `foobar` are rejected when `destDir` is `foo`.

cli/src/commands/ThemeCommand.ts
summerfang-jolli: Fixed in d8fcb35. The parser now throws on typeflags `L`, `K`, `x`, `g`, `1`, and `2` with a clear error message. Went with option (2) — fail loudly rather than silently drop entries. Added a doc comment listing the limitations (no PAX/GNU long-name, paths up to 255 bytes via UStar prefix+name). Tests cover all six rejected typeflags.

cli/src/commands/ThemeCommand.ts
summerfang-jolli: Fixed in d8fcb35. Adopted your suggested single-pass approach — `extractTarGz` is now called once, and the prefix filtering is done in the caller with `allEntries.flatMap()`. The `filterPrefix` parameter has been removed from `extractTarGz`, making the parser simpler and easier to test in isolation.

cli/src/commands/ThemeCommand.ts
summerfang-jolli: Fixed in d8fcb35. The loop now checks `offset + 1024 <= tar.length` and inspects both blocks: two consecutive zero blocks terminate the archive, a single zero block is skipped. Added tests for both cases.

cli/src/commands/ThemeCommand.ts
summerfang-jolli: Added in d8fcb35. 12 tests in `ThemeCommand.test.ts` covering: valid file/directory extraction, NUL typeflag handling, empty archives, all six rejected typeflags (L, K, x, g, 1, 2), two-consecutive-zero-block termination, single-zero-block skip, multi-block files, and path-traversal guard (both sibling-prefix and parent traversal). `extractTarGz` is now exported for direct testing.

Finn-coder2026 (approved):

PR #147, opened 2026-05-25
Fix PAX global header 'g' skipped and 'default' pack short-circuits

Finn-coder2026:
## Quick recap

This commit fixed two distinct bugs in the theme download and installation pipeline. First, the tar archive parser was crashing on every real-world theme download from GitHub because GitHub's download service prepends a PAX global header entry to every tarball it serves — the parser was treating this as a fatal error instead of silently skipping it. Second, the theme discovery system was performing an unnecessary and always-failing network round-trip to GitHub whenever the theme was set to the built-in "default" layout, because there is no corresponding theme package to find. Both fixes are accompanied by new tests that pin the corrected behavior.

---

## Topics (2)
<details>
<summary><strong>01 · Theme downloads broken because GitHub tarball header was rejected as an error</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Every theme download from GitHub was failing at the tar parsing step. GitHub's download service always prepends a special archive-wide metadata header to its tarballs, and the parser was treating this header type as a fatal unsupported format error rather than ignoring it.

- **Silent skip over hard error**: The PAX global header carries no information that affects individual file entries — it is purely archive-wide metadata. Throwing an error on it would make every GitHub-hosted theme permanently uninstallable, since GitHub unconditionally prepends this header. Skipping it is both safe and necessary for the feature to work at all.
- **Narrow the rejection list rather than broaden the skip list**: The other rejected types (`L`, `K`, `x`, `1`, `2`) remain fatal because they either rename the next entry (causing silent data corruption if ignored) or create filesystem links that could escape the destination directory. The PAX global header has neither of those risks, so it warrants different treatment rather than a blanket policy change.

**✅ What Was Implemented**

- In `ThemeCommand.ts`, the PAX global header typeflag was removed from the list of entry types that trigger a thrown error. It was moved into the "skip silently" path instead, with an updated comment explaining that this header carries archive-wide metadata (such as the git commit hash GitHub embeds) and has no effect on the files that follow it.
- A new test case was added to `ThemeCommand.test.ts` that constructs a synthetic tarball containing a PAX global header entry followed by a directory and a regular file, then asserts that the parser returns exactly the two real entries and ignores the header entirely.

**📁 FILES**
- `cli/src/commands/ThemeCommand.ts`

</blockquote>
</details>
<details>
<summary><strong>02 · Unnecessary network call made when using the built-in default theme layout</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

When a project was configured to use the built-in default Nextra layout (rather than any installable theme pack), the theme discovery system was still going through the full lookup sequence — including a network request to GitHub — before concluding that nothing was found. This was a guaranteed-failing round-trip on every invocation.

- **Hard-coded sentinel check over a general "known-missing" registry**: The "default" name has a specific semantic meaning in the system — it signals that the user wants the vanilla Nextra layout with no pack at all. No discovery step can ever satisfy it, making the lookup sequence purely wasteful. A targeted early exit is simpler and more self-documenting than a general mechanism for pre-registering known-absent names.
- **Short-circuit before any I/O rather than after the first miss**: Stopping before the registry lookup (rather than, say, after the npm check fails) eliminates all I/O for this case. The comment in the code explicitly calls out that the GitHub round-trip is the most expensive step being avoided, and that the intent is to let the caller fall straight through to the built-in layout path.

**✅ What Was Implemented**

- In `ThemeRegistry.ts`, an early-return guard was added at the top of the discovery function: if the resolved pack name is the sentinel value "default", the function returns undefined immediately without consulting the in-memory registry, the user themes directory, npm, or GitHub.
- A new test in `ThemeRegistry.test.ts` verifies this short-circuit by spying on the global fetch function and asserting it is never called when the pack name is "default", while also confirming the return value is undefined.

**📁 FILES**
- `cli/src/site/themes/ThemeRegistry.ts`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 25, 2026 at 10:59 AM · via Jolli proxy*

### DIFF
--- a/cli/src/commands/ReverseCommand.ts
+++ b/cli/src/commands/ReverseCommand.ts
@@ -36,10 +36,11 @@ function parseMetaFile(filePath: string): ParsedMeta | null {
 	if (!match) return null;
 	try {
 		let obj = match[1];
-		// Remove trailing commas for eval compatibility
+		// Remove trailing commas
 		obj = obj.replace(/,(\s*[}\]])/g, "$1");
-		// biome-ignore lint/security/noGlobalEval: _meta files are trusted build output
-		return eval(`(${obj})`) as ParsedMeta;
+		// _meta files are trusted build output. Function constructor (not direct
+		// eval) so a bundler can statically analyze the call without scope-capture warnings.
+		return new Function(`return (${obj})`)() as ParsedMeta;
 	} catch {
 		return null;
 	}
@@ -316,8 +317,9 @@ function extractFooter(buildDir: string): Record<string, unknown> {
 		const colsMatch = content.match(/FOOTER_COLUMNS\s*=\s*(\[[\s\S]*?\])\s*\n\n/);
 		if (colsMatch) {
 			try {
-				// biome-ignore lint/security/noGlobalEval: trusted build output
-				const cols = eval(colsMatch[1]) as Array<{
+				// Trusted build output. Function constructor (not direct eval) so
+				// a bundler can statically analyze the call without scope-capture warnings.
+				const cols = new Function(`return (${colsMatch[1]})`)() as Array<{
 					title: string;
 					links: Array<{ label: string; url: string }>;
 				}>;
@@ -336,8 +338,9 @@ function extractFooter(buildDir: string): Record<string, unknown> {
 		const socialMatch = content.match(/FOOTER_SOCIAL_LINKS\s*=\s*(\{[\s\S]*?\})\s*\n/);
 		if (socialMatch) {
 			try {
-				// biome-ignore lint/security/noGlobalEval: trusted build output
-				const social = eval(`(${socialMatch[1]})`) as Record<string, string>;
+				// Trusted build output. Function constructor (not direct eval) so
+				// a bundler can statically analyze the call without scope-capture warnings.
+				const social = new Function(`return (${socialMatch[1]})`)() as Record<string, string>;
 				const socialLinks: Record<string, string> = {};
 				for (const [key, value] of Object.entries(social)) {
 					if (value) socialLinks[key] = value;
--- a/cli/src/commands/ThemeCommand.test.ts
+++ b/cli/src/commands/ThemeCommand.test.ts
@@ -106,11 +106,13 @@ describe("extractTarGz", () => {
 		expect(() => extractTarGz(gz)).toThrow(/Unsupported tar entry type 'x'/);
 	});
 
-	it("skips PAX global header typeflag 'g' and keeps parsing later entries", () => {
+	it("skips PAX global header typeflag 'g' carrying only comment= and keeps parsing later entries", () => {
 		// GitHub's codeload.github.com prepends a `pax_global_header` entry
-		// (typeflag 'g') to every tarball. Treating it as fatal would break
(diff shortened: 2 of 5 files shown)


flyerli2025 (dismissed):
P2 [ThemeRegistry.ts#L148](https://github.com/jolliai/jolliai/blob/d16c2a48324c295c34cb4cb8fa2364b0da678120/cli/src/site/themes/ThemeRegistry.ts#L148): the new if (name === "default") return undefined; runs before the runtime registry lookup. That makes discoverPack() inconsistent with registerPack() / resolvePack(): a pack whose manifest.name is "default" can still be registered, but initNextraProject() will never load it because it relies on discoverPack()’s return value for layout generation and capability validation. If "default" is meant to be reserved, that needs to be enforced at registration/API boundaries; otherwise this guard should move after the registry check.

P2 [ThemeCommand.ts#L174](https://github.com/jolliai/jolliai/blob/d16c2a48324c295c34cb4cb8fa2364b0da678120/cli/src/commands/ThemeCommand.ts#L174): treating every PAX global header (typeflag 'g') as a benign no-op is too broad. In the pax format, g records can apply metadata to all subsequent entries, including path-related fields, so skipping them wholesale is not generally semantics-preserving; it only happens to work for GitHub’s current tarballs if they contain a harmless comment= record. The safer fix is to special-case the known GitHub header shape or parse/allowlist supported keys rather than declaring all g headers safe. Reference: [POSIX pax(1p)](https://www.unix.com/man_page/posix/1p/pax/).

Finn-coder2026:
Replying to @flyerli2025 on https://github.com/jolliai/jolliai/pull/147#pullrequestreview-4354313659

Thanks for the careful read. Both P2 items were design issues rather than pure correctness bugs, and both are now fixed in commit 1921de5.

**P2 — `ThemeRegistry.ts` `if (name === "default") return undefined;` placement**

You're right that running the short-circuit before the registry lookup made `discoverPack()` inconsistent with `registerPack()` / `resolvePack()`. The guard is now placed *after* `packs.get(name)` so the registry stays the source of truth: a user-registered pack named `default` wins everywhere, and the sentinel short-circuit only kicks in for the (still extremely common) case where nothing is registered. A new test `a registered 'default' pack wins over the sentinel short-circuit` pins this contract.

**P2 — `ThemeCommand.ts` PAX `g` global header**

You're also right that treating every `g` header as a no-op is too broad. The replacement now parses each `<length> <key>=<value>\n` record in the global-header data block and rejects any key that could rewrite later entries — `path`, `linkpath`, `size`, and anything else outside an explicit allowlist (`comment`, `charset`, `hdrcharset`, `mtime`, `atime`, `ctime`, `uname`, `gname`, `uid`, `gid`). GitHub's `comment=<sha>` tarballs stay on the allowlist and continue to extract cleanly; an archive that tries to slip a `path=` past us now fails loud (`Refusing to skip PAX global header key "path"…`). New tests cover both the allowlist success path (single + multi-record) and four failure paths (`path=`, `linkpath=`, `size=`, and an unknown key).

flyerli2025 (dismissed):

flyerli2025 (approved):

PR #156, opened 2026-05-26
Make navigation group always a separator, never a folder

summerfang-jolli:
## Summary
- `parseContent` no longer branches on `group.root`. Groups always emit a `__group-<slug>` separator at the parent `_meta.js` level — never a clickable folder.
- When `root` is set, child articles are hoisted up as flat `{ title, href }` link entries and the first root segment gets a `display: "hidden"` entry so the auto-discovered nested directory doesn't surface alongside the separator.
- `ContentPlanner` is untouched: article MDX files still land at `<root>/<href>.mdx`, so URLs remain stable.

## Why
The schema reference has always described a group as a "non-clickable section heading" and `root` as a path-resolution aid only. The old code violated the contract by branching on `root` and producing a clickable folder, giving the same primitive two different sidebar shapes. Discussed with @luke; aligning code to the doc.

## Trade-off
Nested `articles: [...]` hierarchies inside a *rooted* group flatten to sibling href-links — Nextra's collapsible folders are filesystem-bound, so an entry cannot be both a link and a collapsible parent. Users wanting collapsibility can drop the group's root or wrap the section in a page.

## Test plan
- [x] `npm run all` passes (5237 cli tests + vscode tests; global coverage 99.1% stmts / 97.86% branches)
- [x] `StructureParser.test.ts` rewritten — every test that codified the old "group+root → folder" behavior now asserts separator + hidden first-segment + flat href-link entries
- [x] New coverage for external links and absolute-href articles inside rooted groups
- [ ] Smoke-test a generated site with a real `group + root` site.json to confirm sidebar renders as separator + flat links in Nextra

## Quick recap

Navigation groups in the sidebar now always appear as non-clickable section headings, matching the behavior the product documentation had always promised. Previously, adding a URL root to a group would silently promote it into a clickable folder — a behavior the schema explicitly ruled out but the code had been producing. The fix ensures the two concerns stay independent: a group's visual treatment in the sidebar is always a separator, and any URL prefix it carries applies only to how child article addresses are resolved.

When a group carries a URL root, its child articles now appear as flat links directly beneath the separator in the sidebar, with their full nested URLs intact. The physical content files remain at their nested locations on disk, keeping all existing URLs stable. Users who need a genuinely collapsible folder section in the sidebar can use the existing article-with-children or page constructs, both of which were already part of the schema and are unaffected by this change.

---

## E2E Test (3)
<details>
<summary><strong>1. Navigation group with a root always appears as a section heading, never a clickable folder</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a documentation site set up with at least one navigation group that has a sub-path (root) configured, such as a "Getting Started" group pointing to the "getting-started" URL path, containing two or more articles (e.g. "Quickstart" and "Install").

**Steps:**
1. Open the documentation site in a browser and look at the left-hand sidebar.
2. Find the navigation group label (e.g. "Getting Started") in the sidebar.
3. Check whether the group label is clickable — try clicking on it.
4. Verify that the articles listed under the group (e.g. "Quickstart", "Install") appear directly in the sidebar beneath the group heading, not nested inside a collapsible folder.
5. Click one of the article links (e.g. "Quickstart") and confirm it opens the correct page at the expected URL (e.g. /getting-started/quickstart).
6. Return to the sidebar and confirm the group heading itself never navigated anywhere when clicked in step 3.

**Expected Results:**
- The group label (e.g. "Getting Started") appears as a plain, non-clickable section heading — clicking it does nothing and the page does not change.
- The articles beneath it (e.g. "Quickstart", "Install") are listed as direct links in the sidebar, not hidden inside a collapsible sub-folder.
- Clicking an article link opens the correct page at the nested URL (e.g. /getting-started/quickstart).
- There is no separate clickable folder entry for the group's sub-path (e.g. no "getting-started" folder item appears in the sidebar).

</blockquote>
</details>
<details>
<summary><strong>2. Navigation group without a root still shows articles as collapsible sidebar entries</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a documentation site with at least one navigation group that has no sub-path (root) configured, containing articles that themselves have child articles (e.g. a "SQL" group with an "Operations" article that has "Aggregate" and "Window" children).

**Steps:**
1. Open the documentation site and look at the left-hand sidebar.
2. Find the navigation group label that has no root configured (e.g. "SQL").
3. Confirm the group label appears as a non-clickable section heading.
4. Look for the top-level article beneath it (e.g. "Operations") and check whether it appears as a collapsible folder that can be expanded.
5. Click or expand the article folder and verify the child articles (e.g. "Aggregate", "Window") appear nested inside it.

**Expected Results:**
- The group label appears as a plain section heading (not clickable).
- Articles with children appear as collapsible folders in the sidebar, preserving the nested hierarchy.
- Expanding a parent article reveals its child articles nested beneath it, not flattened to the same level.

</blockquote>
</details>
<details>
<summary><strong>3. Articles inside a rooted group link to the correct nested URLs</strong></summary>
<br>
<blockquote>

**Preconditions:** Have a documentation site with a navigation group that has a multi-segment root (e.g. "Guides" with root "docs/guides") and at least one article inside it (e.g. "Intro").

**Steps:**
1. Open the documentation site and locate the group heading in the sidebar (e.g. "Guides").
2. Find the article link listed beneath the group heading (e.g. "Intro").
3. Click the article link.
4. Check the URL in the browser address bar after the page loads.

**Expected Results:**
- The article link is visible directly in the sidebar beneath the group heading, not inside a nested folder.
- Clicking the link navigates to the full nested URL (e.g. /docs/guides/intro), confirming the root path is correctly applied to the article's URL even though the sidebar shows it flat.
- No intermediate folder entries (e.g. "docs" or "guides" as clickable items) appear in the sidebar.

</blockquote>
</details>

---

## Topic (1)
<details>
<summary><strong>01 · Navigation groups always render as separators, never as clickable folders</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The product manager identified a mismatch between the documented behavior and the actual code: the schema promised that a navigation group would always appear as a non-clickable section heading in the sidebar, but the code was silently turning groups into clickable folders whenever a URL root was specified alongside them.

- **Separating sidebar shape from URL structure**: The old design coupled two orthogonal concerns — what the sidebar looks like and what URL prefix child articles get — into a single field. This made it impossible to have a group that added a URL prefix while still rendering flat, or a separator group whose articles shared a path root. Decoupling them restores full expressiveness and matches the documented contract without adding new fields.
- **Flattening nested children in rooted groups rather than preserving hierarchy**: Nextra's collapsible folder mechanism is bound to the filesystem, so a sidebar entry cannot simultaneously be a clickable link and a collapsible parent. Rather than introduce a new filesystem-level construct, nested children inside a rooted group are flattened into sibling link entries. Users who need collapsible hierarchy can use a non-rooted group with multi-segment hrefs, or a page wrapper — both already existed in the schema.
- **Hiding the auto-discovered directory rather than removing it**: The physical content files still live under the root path on disk, so Nextra would auto-discover the nested directory and surface it as a duplicate entry in the sidebar. Emitting a `display: "hidden"` marker on the first path segment suppresses that auto-discovery without breaking URL routing, keeping the sidebar clean while leaving the content addressable.

**✅ What Was Implemented**

- `StructureParser.ts` `parseContent` was refactored to eliminate the two-branch logic that produced different sidebar shapes depending on whether a group had a root field. Groups now unconditionally emit a separator entry at the parent meta level. When a root is present, the first path segment gets a `display: "hidden"` marker to suppress the auto-discovered nested directory, and all child articles are hoisted up as flat link entries (title + href) at the parent level via a new `addArticleAsLink` helper. The URL paths for those articles still nest under the root as before.
- `Types.ts` gained a `display?: "hidden"` field on `SidebarItemValue` to typecheck the new hidden-folder entries, and the `NavigationGroup` interface and its `root` field received updated documentation explicitly stating that groups are always separators and `root` only affects URL resolution.
- All tests in `StructureParser.test.ts` that encoded the old "group + root → clickable folder" behavior were rewritten to assert the new shape: separator entry present, first root segment hidden, articles appearing as flat href-link entries at the parent level, and no nested sidebar override objects. Two new tests were added covering external links and absolute-href articles inside rooted groups.

**📁 FILES**
- `cli/src/site/StructureParser.ts`
- `cli/src/site/Types.ts`

</blockquote>
</details>

---

*Generated by Jolli Memory · May 25, 2026 at 11:51 PM · via Anthropic*

### DIFF
--- a/cli/schemas/site-config.json
+++ b/cli/schemas/site-config.json
@@ -439,7 +439,7 @@
 		},
 		"NavigationGroup": {
 			"type": "object",
-			"description": "A non-clickable section heading that clusters articles.",
+			"description": "A non-clickable section heading that clusters articles. Always renders as a sidebar separator.",
 			"required": ["group", "content"],
 			"additionalProperties": false,
 			"properties": {
@@ -449,11 +449,11 @@
 				},
 				"root": {
 					"type": "string",
-					"description": "Optional sub-path. Article href values in this group are resolved relative to this."
+					"description": "Source-location hint. Source files for this group's articles live under <page.root>/<group.root>/. Does NOT affect generated URLs or the sidebar tree."
 				},
 				"sourceRoot": {
 					"type": "string",
-					"description": "Physical source directory for this group."
+					"description": "Physical source directory for this group. Takes precedence over `root` for source lookup; does not affect URLs."
 				},
 				"content": {
 					"type": "array",
--- a/cli/schemas/site-config.json
+++ b/cli/schemas/site-config.json
@@ -449,11 +449,11 @@
 				},
 				"root": {
 					"type": "string",
-					"description": "Source-location hint. Source files for this group's articles live under <page.root>/<group.root>/. Does NOT affect generated URLs or the sidebar tree."
+					"description": "For groups, a source-location hint only. Source files for this group's articles live under <page.root>/<group.root>/. Does NOT affect generated URLs or the sidebar tree (unlike page.root, which does affect both)."
 				},
 				"sourceRoot": {
 					"type": "string",
-					"description": "Physical source directory for this group. Takes precedence over `root` for source lookup; does not affect URLs."
+					"description": "For groups, the physical source directory. Takes precedence over `root` for source lookup; does not affect URLs or the sidebar tree."
 				},
 				"content": {
 					"type": "array",
--- a/cli/src/commands/StartCommand.test.ts
+++ b/cli/src/commands/StartCommand.test.ts
@@ -1213,6 +1213,17 @@ describe("navigationClaimsRootIndex", () => {
 		expect(navigationClaimsRootIndex([{ article: "Home", href: "/" }])).toBe(true);
 	});
 
+	it("returns true for relative index hrefs ('index', 'index.md', 'index.mdx')", async () => {
+		const { navigationClaimsRootIndex } = await import("./StartCommand.js");
+		// Article hrefs are conventionally relative, so a bare `index` should
+		// also count as claiming root. Today no observable issue because
+		// `pickRootRedirectHref` returns undefined in simple mode, but pinning
+		// this prevents a future tweak from re-introducing a flash.
+		expect(navigationClaimsRootIndex([{ article: "Home", href: "index" }])).toBe(true);
+		expect(navigationClaimsRootIndex([{ article: "Home", href: "index.md" }])).toBe(true);
+		expect(navigationClaimsRootIndex([{ article: "Home", href: "index.mdx" }])).toBe(true);
+	});
+
 	it("returns false when an external link uses '/'-like href (external never claims root)", async () => {
 		const { navigationClaimsRootIndex } = await import("./StartCommand.js");
 		expect(
--- a/cli/src/commands/StartCommand.test.ts
+++ b/cli/src/commands/StartCommand.test.ts
@@ -1111,41 +1111,131 @@ describe("buildRootInjectionInput", () => {
 	});
 });
 
-// ─── hasUserAuthoredRootIndex / pickRootRedirectHref / writeRootRedirectIndex unit tests ────
+// ─── navigationClaimsRootIndex / pickRootRedirectHref / writeRootRedirectIndex unit tests ────
 
-describe("hasUserAuthoredRootIndex", () => {
-	it("returns true when the mirror has a root index.md", async () => {
-		const { hasUserAuthoredRootIndex } = await import("./StartCommand.js");
-		expect(hasUserAuthoredRootIndex({ markdownFiles: ["index.md", "foo.md"] })).toBe(true);
+describe("navigationClaimsRootIndex", () => {
+	it("returns false when navigation is undefined", async () => {
+		const { navigationClaimsRootIndex } = await import("./StartCommand.js");
+		expect(navigationClaimsRootIndex(undefined)).toBe(false);
 	});
 
-	it("returns true when the mirror has a root index.mdx", async () => {
-		// Covers user-authored MDX landing pages.
-		const { hasUserAuthoredRootIndex } = await import("./StartCommand.js");
-		expect(hasUserAuthoredRootIndex({ markdownFiles: ["index.mdx", "foo.md"] })).toBe(true);
+	it("returns false when navigation is empty", async () => {
+		const { navigationClaimsRootIndex } = await import("./StartCommand.js");
+		expect(navigationClaimsRootIndex([])).toBe(false);
 	});
 
-	it("returns false when no index file is present", async () => {
-		const { hasUserAuthoredRootIndex } = await import("./StartCommand.js");
-		expect(hasUserAuthoredRootIndex({ markdownFiles: ["docs/getting-started.md", "api.md"] })).toBe(false);
+	it("returns true when a page is rooted at /", async () => {
+		const { navigationClaimsRootIndex } = await import("./StartCommand.js");
+		expect(
+			navigationClaimsRootIndex([
+				{ page: "Home", root: "/", content: [{ article: "Welcome", href: "welcome" }] },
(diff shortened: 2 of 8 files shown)


Finn-coder2026:
## Code Review: `fix/group-always-separator`

**Files reviewed:** 8
**Lines changed:** +417 / -180
**Commits:** 1 — *"Make group a separator and let schema, not filesystem, decide who claims `/`"*

---

### Show-stoppers
None found.

### High Priority
None found.

### Medium Priority

1. **`navigationClaimsRootIndex` ignores hrefs without a leading slash** — `cli/src/commands/StartCommand.ts:331-335`
   `isRootHref` only matches `"/"`, `"/index"`, `"/index.md"`, `"/index.mdx"`. In article context, hrefs are conventionally relative (e.g. `"intro"`, `"index"`). A user writing `{ page: "Docs", root: "/", content: [{ article: "Home", href: "index" }] }` is already covered because the page's `root: "/"` short-circuits. But a simple-mode config `[{ article: "Home", href: "index" }]` would not — `pickRootRedirectHref` returns undefined in simple mode anyway, so today there's no observable regression. Worth a one-line guard or a test that pins this behavior so a future tweak to `pickRootRedirectHref` doesn't silently re-introduce a flash. — Fix: extend `isRootHref` to also match `"index"` / `"index.md"` / `"index.mdx"`, or add a regression test that locks the current simple-mode outcome.

### Low Priority

2. **Stale wording in the suppression comment** — `cli/src/commands/StartCommand.ts:232-245`
   The block comment still says “When the site is in page mode and no page is rooted at `/`…”, but `navigationClaimsRootIndex` now also fires on `href: "/"` in simple mode. The behavior is equivalent because `pickRootRedirectHref` returns `undefined` in simple mode, but the comment reads as if simple mode were out of scope. — Fix: tighten the comment to say “in page mode, and equivalently a no-op in simple mode.”

3. **Schema description mentions URL behavior that the code doesn’t enforce** — `cli/schemas/site-config.json:449-456`
   The new `root` / `sourceRoot` descriptions say “Does NOT affect generated URLs or the sidebar tree.” That’s true for `group.root` but easy to misread as a statement about `page.root`. The page-level definitions a few lines up still rely on `root` for both URL and sidebar. — Fix: prefix with “For groups,” to scope the claim.

### Suggestions

4. **`articleClaimsRoot` recurses without a depth cap** — `cli/src/commands/StartCommand.ts:320-329`
   The schema doesn’t reject pathological nesting. Trusted input today, but a `for…of` over `article.articles` with a `Set`-based visited guard or a depth cap would harden it. Low risk; mention only.

5. **Test churn could be reduced with a small factory** — `cli/src/site/StructureParser.test.ts`
   Several new tests repeat the `{ group: "…", content: [{ article, href, articles: [...] }] }` skeleton with minor variants. A `makeGroup({...})` helper would tighten the diff for future flattening changes. Optional.

6. **`navigationClaimsRootIndex` could short-circuit on the first `claimsRoot` rather than iterating** — already does, but the helper `nodesClaimRoot` re-walks via two near-identical loops. A single recursive walker over `NavigationGroup | NavigationArticle` would dedupe ~10 lines. Cosmetic.

7. **`Types.ts` `display?: "hidden"` doc lives on `SidebarItemValue`** — `cli/src/site/Types.ts:97-105`
   Good addition. Worth a `@see MetaGenerator` link or a one-line cross-reference so the next reader can find the writer side without grepping.

---

**Summary:** Solid, well-scoped refactor — moves the "who claims `/`" decision from filesystem to schema, removes the special-cased "rooted group" branch in `StructureParser`, and the test coverage tracks the behavior change cleanly. Only nits are stale comment wording and an `isRootHref` edge case that's currently latent.

Want me to apply any of these? (Only #1, #2, and #3 are worth a follow-up; #4–#7 are stylistic.)

Finn-coder2026 (approved):

PR #167, opened 2026-05-29
Normalize article href against inherited root (dedup + leading-slash)

summerfang-jolli:
## Summary

Two normalization rules added to `article.href`, applied once in the navigation parser and reused by sidebar generation, content planning, and the mismatch validator:

- **Leading-slash equivalence** — `/intro` ≡ `intro`. Both mean "relative to the inherited root" (`page.root` + enclosing `group.root`), never the OS filesystem root. The previous `article.href.startsWith("/")` branches in `addArticlePlan` / `checkArticle` are gone.
- **Prefix dedup against the inherited root** — if `href` starts with segments that duplicate the *end* of the inherited root, strip them. So `href: "guides/introduction"` inside `group{root:"guides"}` under `page{root:"/docs"}` normalizes to `["introduction"]` instead of producing a phantom `guides` folder next to the group separator. Only one matching prefix is stripped; `guides/guides/intro` keeps the second `guides/`.

The rule extends the schema-intent principle already established for `group.root` (sidebar marker, not URL prefix): when an author repeats the group's source-location prefix inside `href`, the parser treats it as the same path, not as additional nesting. Multi-segment hrefs that *don't* match the inherited-root tail (e.g. `deployment/docker` inside `group{root:"guides"}`) keep the existing behavior — no dedup, intermediate folder created as authored.

`normalizeHrefSegments` is exported from `StructureParser.ts` so `ContentPlanner.ts` reuses the exact same normalization across source lookup, target write, and validation. `inheritedRootSegs` threads through `addNodesPlan` / `checkNodes` alongside the existing `sourceDir` / `targetDir` — no new state.

## Before / after on a typical phantom-folder site.json

```json
{ "page": "Get Started", "root": "/docs", "content": [
  { "group": "Guides", "root": "guides", "content": [
    { "article": "Introduction", "href": "guides/introduction" },
    { "article": "Authentication", "href": "guides/authentication" }
  ]}
]}
```

**Before:** sidebar shows `━ Guides ━` separator **plus a phantom `guides` folder** containing the articles.

**After:** `Introduction` and `Authentication` render flat under the `━ Guides ━` separator. Source still resolves to `docs/guides/introduction.mdx` (via `group.root` source hint), target written to `content/docs/introduction.mdx`.

## Test plan

- [x] 11 new tests in `StructureParser.test.ts`: 5 integration scenarios (group-root dedup, full-root dedup, no-match passthrough, leading-slash equivalence, single-pass conservatism) + 6 direct unit tests for `normalizeHrefSegments`
- [x] All existing navigation tests (44 → 55 in this file) pass — multi-segment hrefs, external links inside rooted groups, `expanded` folder collapse, etc.
- [x] `npm run all` clean: build → lint → test → coverage (CLI 98.7% stmts, 97.23% branches, 98.62% funcs, 99.05% lines)
- [ ] Validated against a real customer site.json with multi-segment hrefs that duplicate group root

### DIFF
--- a/cli/schemas/site-config.json
+++ b/cli/schemas/site-config.json
@@ -476,7 +476,7 @@
 				},
 				"href": {
 					"type": "string",
-					"description": "Path to the content. Resolved relative to the nearest ancestor root."
+					"description": "Path to the content, relative to the inherited root (page.root + enclosing group.root). Two normalization rules: (1) a leading '/' is equivalent to no leading slash — both mean 'relative to the inherited root', never the OS filesystem root, so '/intro' and 'intro' are the same. (2) If the href starts with segments that duplicate the end of the inherited root (e.g. href: 'guides/intro' inside group{root:'guides'} under page{root:'/docs'}), the duplicated prefix is stripped — preventing a phantom intermediate folder in the sidebar."
 				},
 				"source": {
 					"type": "string",
--- a/cli/src/site/ContentPlanner.ts
+++ b/cli/src/site/ContentPlanner.ts
@@ -4,6 +4,7 @@ import { dirname, join } from "node:path";
 import { hasIncompatibleImports, rewriteRelativeImagePaths, stripIncompatibleContent } from "./ContentMirror.js";
 import { slugify } from "./openapi/Slug.js";
 import type { ContentRules } from "./renderer/SiteRenderer.js";
+import { normalizeHrefSegments } from "./StructureParser.js";
 import type { Navigation, NavigationArticle, NavigationGroup, NavigationPage } from "./Types.js";
 
 export interface PlannedMarkdownPage {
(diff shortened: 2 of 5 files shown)


aidancrosbie (approved):
Validated against a phantom-folder fixture — sidebar renders flat under the `━ Guides ━` separator with no phantom `▶ Guides` folder, dedup correctly skips non-matching prefixes (intermediate `deployment/` folder preserved for Docker/Kubernetes), leading-slash equivalence works, and all four URLs resolve. 55/55 unit tests passing locally with 100% statement/function/line coverage on the new code.

One non-blocking follow-up worth considering: when `normalizeHrefSegments` returns `[]` (href exactly equals the inherited root, e.g. a single-segment article filename that happens to match the last segment of a multi-segment group root), the article is silently dropped from both sidebar and plan. Edge case, but the failure mode is exactly the class of "silent disappearance" this PR exists to fix. A defensive guard ("if dedup would empty the segments, don't dedup") is a one-liner.

aidancrosbie (approved):

PR #170, opened 2026-05-30
site-core: close sanitizeUrl backslash bypass on scheme-relative URLs

summerfang-jolli:
## Summary

`SAFE_URL_PATTERN` in `site-core/src/Sanitize.ts` blocked `//evil.com` but **not** the backslash variant `/\evil.com`. Browsers normalise the backslash to `/`, resolving the URL as `//evil.com` — the exact scheme-relative cross-origin vector the function documents protecting against.

## Why this matters

`sanitizeUrl` is the single allow-list for every customer-supplied URL that hits the generated site — `header.items`, `footer.links`, `socialLinks`, page-tab links materialised via `MetaGenerator`. A bypass here is a generic XSS / open-redirect / phishing vector against any site built by Jolli CLI / web tool.

This is **already live** in the shipped `@jolli.ai/cli` (PR #168 merged this code into main). The merge of the cross-repo `tools/site-core` import (jolliai/jolli#965) inherited the same bug. Same patch should be applied in jolli (separate PR, follow-up).

## Fix

```diff
-const SAFE_URL_PATTERN = /^(?:https?:|mailto:|tel:|[#?]|\/(?!\/)|\.\.?\/)/i;
+const SAFE_URL_PATTERN = /^(?:https?:|mailto:|tel:|[#?]|\/(?![/\\])|\.\.?\/)/i;
```

The lookahead now rejects both `/` and `\` immediately after the leading slash.

## Tests

| Input | Before | After |
|---|---|---|
| `//evil.com` | clamp to `#` ✅ | clamp to `#` ✅ |
| **`/\evil.com`** | **pass through (bug)** | **clamp to `#` ✅** |
| `/\\evil.com` | pass through | clamp to `#` ✅ |
| `\\evil.com` | clamp (no leading `/`) | clamp ✅ |
| `/foo/bar` | pass through ✅ | pass through ✅ |
| `/` | pass through ✅ | pass through ✅ |

Adds 2 new test groups (backslash bypass + mixed variants) on top of the existing scheme-relative tests.

## Validation

- `npm run all` ✓ (passes site-core's 597 tests, CLI's full suite, VS Code's full suite)
- Sanitize.test.ts: 32 tests passing (was 28, +4 new)

## Credit

Caught by @aidancrosbie in review of jolliai/jolli#965.

## Follow-up

Apply same patch in jolli/tools/site-core/src/Sanitize.ts (separate PR — that file diverged from this one when PR #965 landed, so it carries the same bug).

### DIFF
--- a/site-core/src/Sanitize.test.ts
+++ b/site-core/src/Sanitize.test.ts
@@ -106,9 +106,28 @@ describe("sanitizeUrl — scheme-relative URLs", () => {
 		expect(sanitizeUrl("///evil.com")).toBe("#");
 	});
 
+	it("clamps the backslash variant `/\\evil.com` to '#'", () => {
+		// Browsers normalise the leading backslash to a forward slash and resolve
+		// the URL as `//evil.com` — the exact scheme-relative cross-origin
+		// vector this allow-list documents protecting against. Without the
+		// `[/\\]` lookahead, `\/(?!\/)` was satisfied (because the next char is
+		// `\`, not `/`) and the value passed through unchanged.
+		expect(sanitizeUrl("/\\evil.com")).toBe("#");
+		expect(sanitizeUrl("/\\evil.com/track?u=1")).toBe("#");
+	});
+
+	it("clamps the mixed slash/backslash variants `/\\\\foo` and `\\\\foo` to '#'", () => {
+		// Defence-in-depth: any combination of leading slashes/backslashes that
+		// browsers parse as scheme-relative must clamp. We can't trust a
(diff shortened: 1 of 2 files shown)


summerfang-jolli:
Superseded — Sanitize fix will land via the jolli-side PR (#966 in that repo, consolidated). jolliai/site-core/ is being deleted in the consolidated jolliai PR, so an in-place fix here is redundant. The fix flows to customers via the next `@jolli.ai/site-core` npm release published from jolli.

PR #277, opened 2026-07-04
Paginate the commit list 6 at a time with a Show more link

summerfang-jolli:
## Quick recap

The commit list in both the IntelliJ and VS Code plugins now shows a maximum of six entries at a time. A "Show more" link appears below the visible entries whenever additional commits exist; each click reveals the next six, and the link disappears once every commit is on screen. The list resets to six entries when the user switches branches or a new commit appears, but stays expanded during background activity such as an AI summary being generated in the background — so scrolling down to read older commits is not interrupted by a routine refresh.

Both plugins received the change in the same session and behave consistently. The IntelliJ side reused an existing styling helper that was already used for capping sub-sections within an expanded commit row, while the VS Code side introduced a dedicated renderer to keep the new incremental behavior separate from the existing binary show-all toggle used by those same sub-sections. A subtle build-time defect — where a separator character was being interpreted as a literal newline inside the generated script — was caught by an existing automated check and corrected before the tests were run.

---

## Topic (1)
<details>
<summary><strong>01 · Paginate commit list to show six at a time with incremental reveal</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

When working in a git repository with no remote upstream configured, the commit list grew unbounded — every commit on the branch was shown at once with no way to limit the view. The user wanted the list capped at six entries initially, with a "Show more" link that reveals the next six on each click until all commits are visible.

- **Render-time cap over data-layer pagination**: All commits are still fetched from git in a single call; the six-at-a-time limit is applied purely when building the UI. This avoids adding offset/limit parameters to the git-log layer and keeps the two plugins' data paths unchanged, at the cost of loading all commits into memory upfront. For the typical branch-commit counts seen here, the memory trade-off is negligible and the implementation stays simple.
- **Incremental reveal instead of binary show-all toggle**: The existing "Show more" helpers in both plugins are binary — one click shows everything, another hides it again. The new commit-list behavior grows the visible window by exactly one page per click and never collapses, matching the user's explicit spec. This required a separate counter variable and a distinct click-target marker rather than reusing the subsection toggle machinery, keeping the two behaviors independent and avoiding regressions in the subsection expand/collapse logic.
- **Reset on sequence change, preserve on content-identical refresh**: The visible count resets to 6 only when the set of commit hashes actually changes (branch switch or new commit). A background AI-summary refresh that leaves the commit list unchanged does not snap the list back to six entries. This distinction required comparing the incoming hash sequence against the stored one before deciding whether to reset, adding a small amount of logic but meaningfully improving the experience for users who scroll down while a summary is being generated.

**✅ What Was Implemented**

- In the IntelliJ plugin, `CommitsPanel.kt` gained a `visibleCommits` counter (starting at 6, matching the existing page-size constant in `CappedRows`). The `updateCommitList()` method now renders only the first `visibleCommits` entries and appends a "Show N more" row — built with the newly public `CappedRows.showMoreRow()` helper — that increments the counter by one page and re-renders on click. The counter resets to 6 inside the existing "commit sequence changed" guard in `refreshFromGit()`, so a branch switch or new commit collapses the list while a background summary refresh leaves the expanded state intact.
- In the VS Code extension, `SidebarScriptBuilder.ts` gained a `COMMITS_PAGE = 6` constant and a transient `commitsVisible` variable. The `renderSection()` function now slices the commits array to `commitsVisible` entries and appends a dedicated `renderCommitsMoreRow()` node (carrying a `data-commits-more` marker distinct from the existing subsection toggle). A new click-delegation branch grows `commitsVisible` by one page and re-renders. The `branch:commitsData` message handler resets the counter only when the incoming commit hash sequence differs from the stored one, preserving the expanded state across content-identical refreshes.
- A regression test was added to `SidebarScriptBuilder.test.ts` asserting that the emitted script contains the page-size constant, the slice logic, the more-row renderer, the click handler, and the reset-on-change guard. A build-time bug was caught and fixed: joining commit hashes with `'\n'` inside the template-literal script body produced a real newline mid-string; switching to `'|'` (safe because hashes are hexadecimal) resolved the parse failure caught by the existing smoke test.

**📋 Future Enhancements**

IntelliJ foreign read-only mode (browsing another repo or branch's memories) uses a separate rendering path that was intentionally left uncapped in this commit. Paginating it for parity with the VS Code foreign-mode path was discussed but deferred.

**📁 FILES**
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/CommitsPanel.kt`
- `intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/CappedRows.kt`
- `vscode/src/views/SidebarScriptBuilder.ts`

</blockquote>
</details>

---

*Generated by Jolli Memory · July 4, 2026 at 10:47 AM · via Anthropic*

### DIFF
--- a/intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/GitOps.kt
+++ b/intellij/src/main/kotlin/ai/jolli/jollimemory/bridge/GitOps.kt
@@ -331,6 +331,53 @@ class GitOps(private val projectDir: String) {
         return lines.last().substringBefore(" ").takeIf { it.isNotBlank() }
     }
 
+    /** Reflog-derived merged-history result: log-range base + whether the branch authored anything. */
+    data class MergedHistory(val base: String, val hasOwnCommit: Boolean)
+
+    /**
+     * Resolves what the merged-mode history view needs from [branch]'s reflog: the
+     * log-range base (where the branch was created) and whether the branch ever
+     * committed anything of its own.
+     *
+     * Used when HEAD is fully contained in the mainline (`merge-base HEAD base ==
+     * HEAD`) — e.g. on `main` in a repo with no remote — so the panel can still show
+     * the user's own commits (`<base>..HEAD --author=<you>`) instead of an empty
+     * list. [hasOwnCommit] keys on a `commit` reflog op: a branch showing only
+     * creation + rebase/reset/checkout never authored anything itself, so the caller
+     * shows an empty panel rather than re-listing commits that belong to the base.
+     *
+     * Returns null for detached HEAD or when the reflog is unavailable/expired —
+     * same graceful-degradation boundary [findBranchCreationPoint] uses. Mirrors the
+     * VS Code bridge's resolveMergedHistory.
+     */
+    fun resolveMergedHistory(branch: String): MergedHistory? {
+        if (branch == "HEAD") return null
+        val reflog = exec("reflog", "show", branch, "--format=%H %gs") ?: return null
+        val lines = reflog.lines().filter { it.isNotBlank() }
+        if (lines.isEmpty()) return null
+
+        // Base: the explicit "branch: Created from …" entry if present (scan from
+        // oldest), else the oldest surviving entry — the same best-effort fallback
+        // the non-strict findBranchCreationPoint uses.
+        var base: String? = null
+        for (i in lines.indices.reversed()) {
+            if (lines[i].contains("branch: Created from")) {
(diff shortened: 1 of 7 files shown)


Finn-coder2026:
When I switch branch to main:
Vscode plugin:
image
Intellij plugin:
image
When I stay in work branch:
Intellij plugin:
image
Vscode plugin:
image
the main branch view is aligned

Finn-coder2026 (approved):

PR #297, opened 2026-07-09
IntelliJ: port context relevance filtering (#294) and Slack references (#295)

summerfang-jolli:
Ports two CLI/VS Code features into the native Kotlin IntelliJ plugin. Both are native reimplementations — IntelliJ runs its own Kotlin summary pipeline (not the Node CLI), so neither feature flowed to IntelliJ automatically. Two DCO-signed commits, one per feature.

Full suite green: **1600 tests, 0 failures** (`./gradlew test --rerun-tasks`, independently re-run on the final tree).

## Commit 1 — AI context relevance filtering (ports #294, full scope)
- Brings **plans and notes into the summary prompt** (IntelliJ was references-only). New `PlanPromptFormatter`/`NotePromptFormatter` render `<plans>`/`<notes>` blocks, caller-order + budget.
- New `ContextRelevance.kt`: one batched `rank-context` LLM call ranking plans/notes/references against the change (changed files + declared symbols); tier + auto-exclude driven by **rank position**, not raw score; **fail-open** (any failure keeps everything).
- Wired into `PostCommitHook.processCommitOp` after user hard-excludes; soft-excluded items are dropped from the prompt **and kept in the working area** (not archived).
- `excludedContext` recorded on `CommitSummary`, rendered in folder markdown **and** the committed-summary detail view.

## Commit 2 — Slack thread references (ports #295)
- Slack as a 5th reference source. `SlackAdapter` + permalink parse/correlate + blob normalize.
- Permalink is the primary URL anchor; `slack.workspaceUrl` config is the fallback; **linkless capture** otherwise → `url` made optional across the reference model, omitted from markdown when absent.
- Tool_use input (`channel_id`/`message_ts`) threaded through the Claude envelope parser (Slack-only); role:user permalinks scanned up front and correlated by channel:ts.
- `slack.workspaceUrl` validated in settings (HTTPS + `slack.com` host) and persisted as a normalized origin. Slack renders in committed-memory + working-memory views with null-url guards.

## Deviations from the CLI (intentional, IntelliJ-specific)
- **Relevance applies to the normal commit path only** — IntelliJ's amend path is a no-LLM 1:1 summary migration, so there is no prompt to filter (unlike the CLI's regenerating amend).
- **Reference *ordering*** by relevance is not applied to the source-grouped, age-ordered reference blocks (soft-excluded refs are still dropped); plans/notes get full relevance ordering.
- **Per-call `timeoutMs` skipped** (IntelliJ's `LlmClient` exposes no per-call timeout) — the fail-open wrapper covers a wedged call. Global Anthropic timeout untouched.
- **Slack binding is Claude-only** (no Codex binding); single `mcp__claude_ai_Slack__` prefix, matching the CLI definition.
- **Pre-commit tier-chip UI not included** — showing tiers *before* commit requires running the ranker in the working-memory panel (a per-refresh LLM call + caching), which is the deferred authoritative-pre-commit-panel subsystem. Backend records everything; this is a follow-on.

Source PRs: #294 (`feature/context-relevance-filtering`), #295 (`feature/slack-mcp-integration`).


summerfang-jolli:
Follow-up fixes after live testing of the Slack capture path in a real project:

- **Fix reference discovery dropping tool calls split across a scan boundary** — incremental discovery started its scan at the cursor, so a `tool_use` written just before the cursor (a discovery tick firing mid-turn, between a `tool_use` and its `tool_result`) was never registered and its result could never pair, silently dropping the reference. Now scans `tool_use`s from line 0 and only emits results at/after the cursor. Affects every MCP reference source, not just Slack. Regression test added.
- **Read Slack permalinks from string-form user message content** — a directly-typed prompt serializes `message.content` as a plain string, not a text-block array, so a pasted thread permalink (the intended primary URL anchor) was never harvested and captures fell back to linkless. `scanUserPermalinks` now handles both shapes. Test added.

Both verified end-to-end on a real transcript: the Slack thread captured with its title, fields, full body, and the permalink-derived URL. Plugin bumped to 0.99.5.4.

Finn-coder2026 (approved):

Finn-coder2026 (approved):

PR #302, opened 2026-07-10
Keep platform-tool internal metadata off tools/list and degrade a bad binding

dougschroeder:
## Commits in this PR (4)

1. Add manifest-driven Jolli-platform tools to the jolli mcp server (`42e7e74`)
2. Add the jolli menu prompt to the jolli mcp server (`cd48de5`)
3. Add a /jolli umbrella-menu skill over the Jolli skills and MCP tools (`c4c1a22`)
4. Keep platform-tool internal metadata off tools/list and degrade a bad binding (`27ac1e5`)

---

*Generated by Jolli Memory · July 12, 2026 at 11:39 AM*

### DIFF
--- a/cli/src/Types.ts
+++ b/cli/src/Types.ts
@@ -1053,6 +1053,16 @@ export interface JolliMemoryConfig {
 	readonly compileExcludeFolders?: ReadonlyArray<string>;
 	/** Jolli Space API key for pushing summaries and proxy LLM calls (sk-jol-...) */
 	readonly jolliApiKey?: string;
+	/**
+	 * Opt-in gate for the manifest-driven Jolli-platform MCP tools. When `true`,
+	 * the `jolli mcp` server also registers the backend-defined platform tools
+	 * (fetched from the tenant's tool manifest at startup) alongside the built-in
+	 * git-memory tools. Off by default — when absent or `false`, the server stays
+	 * git-memory-only and never contacts the backend for a manifest. Opt-in
+	 * polarity, unlike the `*Enabled` source-discovery flags below which default
+	 * to on. The `JOLLI_MCP_PLATFORM_TOOLS=1` env var overrides it at read time.
+	 */
+	readonly mcpPlatformToolsEnabled?: boolean;
 	/** Enable Codex CLI session discovery at post-commit time (default: auto-detect) */
 	readonly codexEnabled?: boolean;
 	/** Enable Gemini CLI session tracking via AfterAgent hook (default: auto-detect) */
--- a/cli/src/commands/ConfigureCommand.test.ts
+++ b/cli/src/commands/ConfigureCommand.test.ts
@@ -448,4 +448,40 @@ describe("ConfigureCommand — settable keys", () => {
 			await expectSetRejected("-30", /positive integer/);
 		});
 	});
+
+	describe("mcpPlatformToolsEnabled boolean coercion via --set", () => {
+		it("accepts true/false/yes/no/1/0 forms", async () => {
+			for (const truthy of ["true", "yes", "1"]) {
+				mockSaveConfig.mockClear();
+				await runConfigure(["--set", `mcpPlatformToolsEnabled=${truthy}`]);
+				expect(mockSaveConfig).toHaveBeenCalledWith(expect.objectContaining({ mcpPlatformToolsEnabled: true }));
+			}
+			for (const falsy of ["false", "no", "0"]) {
+				mockSaveConfig.mockClear();
+				await runConfigure(["--set", `mcpPlatformToolsEnabled=${falsy}`]);
+				expect(mockSaveConfig).toHaveBeenCalledWith(
+					expect.objectContaining({ mcpPlatformToolsEnabled: false }),
+				);
+			}
+		});
+
+		it("rejects garbage boolean values", async () => {
+			const errorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+			const prev = process.exitCode;
+			try {
+				await runConfigure(["--set", "mcpPlatformToolsEnabled=maybe"]);
+				expect(errorSpy).toHaveBeenCalledWith(expect.stringMatching(/true\/false/));
+				expect(process.exitCode).toBe(1);
+				expect(mockSaveConfig).not.toHaveBeenCalled();
+			} finally {
+				errorSpy.mockRestore();
+				process.exitCode = prev;
+			}
+		});
+
+		it("lists mcpPlatformToolsEnabled in --list-keys output", async () => {
+			const help = await runConfigureHelp();
+			expect(help).toContain("mcpPlatformToolsEnabled");
+		});
+	});
 });
--- a/cli/src/commands/ConfigureCommand.ts
+++ b/cli/src/commands/ConfigureCommand.ts
@@ -50,6 +50,7 @@ const VALID_CONFIG_KEYS = [
 	"openCodeEnabled",
 	"cursorEnabled",
 	"copilotEnabled",
+	"mcpPlatformToolsEnabled",
 	"globalInstructions",
 	"logLevel",
 	"excludePatterns",
@@ -122,6 +123,7 @@ function coerceConfigValue(key: ConfigKey, raw: string): string | number | boole
 		key === "openCodeEnabled" ||
 		key === "cursorEnabled" ||
 		key === "copilotEnabled" ||
+		key === "mcpPlatformToolsEnabled" ||
 		key === "syncTranscripts" ||
 		key === "syncOnPush"
 	) {
@@ -201,6 +203,11 @@ const CONFIG_KEY_INFO: ReadonlyArray<{ key: ConfigKey; type: string; description
 		key: "copilotEnabled",
 		type: "boolean",
 		description: "Enable Copilot CLI session discovery (true/false; requires Node 22.5+ at runtime)",
+	},
+	{
+		key: "mcpPlatformToolsEnabled",
+		type: "boolean",
+		description: "Register backend-defined Jolli-platform tools in the MCP server (true/false; off by default)",
 	},
 	{ key: "logLevel", type: "enum", description: "Log level: debug | info | warn | error" },
 	{ key: "excludePatterns", type: "string[]", description: "Glob patterns for file exclusion (comma-separated)" },
--- a/cli/src/core/JolliMemoryPushClient.test.ts
+++ b/cli/src/core/JolliMemoryPushClient.test.ts
@@ -586,12 +586,6 @@ describe("fetchManifest", () => {
 						description: "required has non-string members",
 						inputSchema: { type: "object", properties: {}, required: [123] },
 					},
-					{
-						name: "bad_binding",
-						description: "binding missing path",
-						inputSchema: { type: "object", properties: {} },
-						binding: { method: "POST" },
-					},
 					"not-an-object",
 					TOOL_B,
 				],
@@ -652,6 +646,57 @@ describe("fetchManifest", () => {
 		expect(tool.binding).toEqual({ method: "POST", path: "/api/mcp/tools/list_workflow_definitions" });
 	});
 
+	it("drops a malformed binding WITHOUT dropping the tool (like menu)", async () => {
+		// `binding` is internal routing metadata, never advertised, so a bad one can
+		// never poison tools/list — drop only the binding and keep the tool callable
+		// (the generic executor then falls back to POST /api/mcp/tools/<name>).
+		const c = client(async () =>
+			jsonResponse(200, {
+				tools: [
+					{
+						name: "not_object",
+						description: "d",
+						inputSchema: { type: "object", properties: {} },
+						binding: "nope",
+					},
+					{
+						name: "arr_binding",
+						description: "d",
+						inputSchema: { type: "object", properties: {} },
+						binding: [],
+					},
+					{
+						name: "no_path",
+						description: "d",
+						inputSchema: { type: "object", properties: {} },
+						binding: { method: "POST" },
+					},
+					{
+						name: "no_method",
+						description: "d",
+						inputSchema: { type: "object", properties: {} },
+						binding: { path: "/api/mcp/tools/x" },
+					},
+					{
+						name: "non_string_method",
+						description: "d",
+						inputSchema: { type: "object", properties: {} },
+						binding: { method: 1, path: "/api/mcp/tools/x" },
+					},
+				],
+			}),
+		);
+		const tools = await c.fetchManifest();
+		expect(tools.map((t) => t.name)).toEqual([
+			"not_object",
+			"arr_binding",
+			"no_path",
+			"no_method",
+			"non_string_method",
+		]);
+		expect(tools.every((t) => t.binding === undefined)).toBe(true);
+	});
+
 	it("preserves a valid menu block on a fetched entry", async () => {
 		const withMenu = {
 			name: "create_ticket",
@@ -681,7 +726,7 @@ describe("fetchManifest", () => {
 		expect(tool.menu).toEqual({ label: "Just a label" });
 	});
 
-	it("drops a malformed menu block WITHOUT dropping the tool (unlike binding)", async () => {
+	it("drops a malformed menu block WITHOUT dropping the tool (like binding)", async () => {
 		// Each of these entries has a bad menu but must remain a usable tool with no
 		// menu — a partially-rolled-out backend must never lose a working tool.
 		const c = client(async () =>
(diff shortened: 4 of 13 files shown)


flyerli2025:
## Review

Overall this is high quality: opt-in and off by default, the asymmetric best-effort-manifest / loud-fail-invocation design is well reasoned and documented, field-level validation, and broad test coverage. One blocker: the same-origin safeguard introduced in Topic 03 — arguably the whole point of that topic — is bypassable, so the bearer token can still be redirected off-origin.

### 🔴 Critical — same-origin guard is bypassable; Bearer token can be exfiltrated to a third-party host

`cli/src/core/JolliMemoryPushClient.ts` (`resolveToolEndpoint`, ~L453-459):

```ts
if (!binding || !binding.path.startsWith("/") || binding.path.startsWith("//")) {
    return { method: "POST", path: `/api/mcp/tools/${encodeURIComponent(tool.name)}` };
}
return { method: binding.method, path: binding.path };
```

`path` then flows into `new URL(path, origin)` in `call()` (~L413) with an `Authorization: Bearer <apiKey>` header. A string-prefix check can't gate this, because the WHATWG URL parser normalizes the input first. Verified in Node:

| `binding.path` from the manifest | passes the guard? | request actually goes to |
|---|---|---|
| `//evil.example/steal` | ❌ rejected (tested) | — |
| `https://evil.example` | ❌ rejected (tested) | — |
| `/\evil.example/steal` | ✅ **passes** | `https://evil.example/steal` |
| `/⇥/evil.example` (embedded TAB/CR/LF) | ✅ **passes** | `https://evil.example/` |

For http/https, WHATWG turns `\` into `/` and strips tab/newline/CR before parsing, so a single-slash path becomes protocol-relative and picks up an attacker host. This defeats exactly the threat Topic 03 claims to defend against ("prevent a poisoned binding from redirecting an authenticated request, including its access credential, elsewhere").

**Failure scenario**: with the feature enabled, a manifest entry advertising `binding.path: "/\attacker.com/x"` causes the next call to that tool to send `sk-jol-…` to `attacker.com`.

**Fix**: resolve, then compare origin — don't inspect the raw string. This matches the pattern already used in `assertJolliOriginAllowed`:

```ts
function resolveToolEndpoint(tool, origin) {
    const fallback = { method: "POST", path: `/api/mcp/tools/${encodeURIComponent(tool.name)}` };
    if (!tool.binding) return fallback;
    try {
        const resolved = new URL(tool.binding.path, origin);
        if (resolved.origin !== origin) return fallback;   // compare AFTER normalization
        return { method: tool.binding.method, path: resolved.pathname + resolved.search };
    } catch {
        return fallback;
    }
}
```

`resolveToolEndpoint` doesn't currently have `origin` — it needs to be threaded in from `resolveAuth`/`call`, or the check moved down into `call()`.

### 🟠 High — the SSRF test only covers the obvious payloads, giving false confidence

`cli/src/core/JolliMemoryPushClient.test.ts` (`ignores an off-origin binding path`, ~L314-326) only exercises `https://evil.example/steal` and `//evil.example/steal` — it misses the `\` and embedded-control-char vectors that actually get through. When fixing the Critical, please add those vectors to the parametrized case to pin the regression.

### 🟡 Low — `binding.method` is not validated

`resolveToolEndpoint` (~L458) passes the backend-supplied method string straight to `fetch`, and `call()`'s type was widened from `"GET" | "POST" | "DELETE"` to `string`. Once the origin issue is fixed, an invalid method just throws and gets wrapped as `isError`, so the impact is limited — but since both method and path come from the manifest, consider an uppercase allowlist (`GET/POST/PUT/PATCH/DELETE`). Related informational note: a GET-bound tool would still send `args` as a body (`call()` unconditionally `JSON.stringify(body)`), which undici rejects for GET — again a silent failure, not a security issue.

### ✅ What's good

- The opt-in gate (config flag + strict `=== "1"` env override) is consistent with the repo's other `*Enabled` semantics, and the doc comment calls out that the polarity is inverted.
- The `fetchManifest` best-effort-collapses-to-`[]` vs `invokePlatformTool` loud-fail asymmetry is deliberate and well explained (startup-time discovery vs. a live user action).
- Built-in-wins-on-name-collision, and the static `TOOL_DEFINITIONS` is never mutated (with no platform tools the static array is returned directly). Clean.
- Field-level `isPlatformToolEntry` validation with "drop the bad entry, keep the good ones" is thoroughly tested.

**Bottom line**: solid design and engineering, but the Critical (same-origin bypass) should be fixed before merge — it's one of the security controls this PR exists to add, and today it's effectively a no-op. Recommend switching to origin comparison and adding the missing test vectors.

flyerli2025 (commented):
Two more issues I found beyond the origin-bypass concern already discussed:

1. `cli/src/core/JolliMemoryPushClient.ts` (`isPlatformToolEntry`, around L482-L510): the manifest validator is out of sync with MCP `ToolSchema`. It currently *requires* `inputSchema.properties`, so a valid no-arg schema like `{ type: "object" }` gets dropped, and the new test at `cli/src/core/JolliMemoryPushClient.test.ts` even locks that behavior in (`no_props`). At the same time, the validator never checks that `required` is an array of strings and it accepts any object-ish `properties` value, so malformed entries like `required: "title"` or `properties: []` still survive. Because `startMcpServer` advertises these tools verbatim, one malformed backend entry can make the whole `tools/list` response fail schema validation for MCP clients, while a valid zero-arg tool silently disappears. I verified against the SDK schema locally that `{ type: "object" }` is valid and those malformed variants are not.

2. `cli/src/mcp/McpServer.ts` (around L226-L231): duplicate platform tool names produce an advertise/call mismatch. `platformByName` is built with `new Map(...)`, so the last duplicate wins for execution, but `toolDefinitions` still includes every duplicate from the manifest. If the backend emits two `create_ticket` entries with different schemas or bindings, clients can see and choose the first one from `tools/list`, but `tools/call` will always invoke the last one. That leaves one advertised tool unreachable and can execute a different backend action than the schema/description the caller selected.

Focused validation still passed under Node 24.10.0 for the touched Vitest suites.

dougschroeder:
Thanks for the thorough review — all five points addressed.

**🔴 Critical (same-origin guard bypass).** Fixed. Replaced the raw-string prefix check with origin comparison *after* normalization: `resolveToolEndpoint` now resolves `binding.path` against the tenant origin via `new URL()` and honors it only when `resolved.origin === origin`, otherwise falls back to `POST /api/mcp/tools/<name>`. The method + path are resolved in `invokePlatformTool` (origin from the resolved auth) before the request is issued, so `/\host`, embedded tab/CR/LF, protocol-relative, and absolute off-origin paths all fall back and the bearer token stays on-origin.

**🟠 High (shallow SSRF test).** Fixed. The off-origin test is now parametrized over the bypass vectors too — `https://evil.example/steal`, `//evil.example/steal`, `/\evil.example/steal`, and tab/CR/LF variants — each asserted to fall back to the conventional endpoint. Added positive cases for a same-origin absolute-URL binding and a same-origin non-POST method.

**🟡 Low (unvalidated method).** Fixed. `resolveToolEndpoint` uppercases the binding method and honors it only against an allowlist (`GET/POST/PUT/PATCH/DELETE`); anything else falls back to the conventional POST endpoint. On the GET-with-body note: a GET-bound tool with args isn't part of the contract (invocations are POST), and were one advertised the request error surfaces through the existing loud-fail / `isError` path rather than silently — so I left `call()` as-is.

**Validator out of sync with the tool schema.** Fixed. Replaced the strict filter with a normalizer: `type` must be `"object"`; `properties` is now optional and defaulted to `{}` (a valid zero-arg `{ type: "object" }` is kept — the old drop-behavior test is replaced with one asserting it's kept and normalized); `properties` must be a plain object when present (an array is rejected); `required` must be an array of strings when present (`required: "title"` and non-string members are rejected); other schema keywords are preserved.

**Duplicate tool names.** Fixed. Platform tools are deduped by name (first wins) before both the advertised list and the dispatch map are built, so `tools/list` and `tools/call` can no longer disagree; a dropped duplicate is logged.

Verification: `npm run all` green across cli + vscode; coverage holds (cli 99.31% / 98.18% / 99.14% / 99.51%).

flyerli2025:
## Code review

Read the full diff (13 files, +1586/-17) and verified the MCP SDK's actual outgoing-message behavior. Overall a high-quality PR — the SSRF guard, dedup/collision handling, best-effort degradation, and field-level tolerance are all well thought through, with strong test coverage. A few things below.

### 🟡 Medium — `binding` / `menu` internal metadata leaks into `tools/list`

`McpServer.ts` builds the advertised list by spreading the full objects:

```ts
platformTools.length > 0 ? [...TOOL_DEFINITIONS, ...platformTools] : TOOL_DEFINITIONS;
```

`platformTools` are `PlatformToolManifestEntry[]`, which carry `binding` (`{method, path}`) and `menu` on top of `name/description/inputSchema`. But the doc comment on `PlatformToolManifestEntry` states binding is *"internal routing metadata, not part of the advertised tool schema"* — the implementation advertises it anyway.

I checked this isn't silently stripped by the SDK: `ToolSchema` is a plain `z.object` (no `.passthrough()`), **but the SDK doesn't validate outgoing results** — the request handler's return value is sent verbatim (`protocol.js`: `{ result, jsonrpc, id }` → `transport.send(response)`, JSON-serialized as-is). So every MCP client sees the backend routing path (`binding.path`) and menu curation in `tools/list`.

Not a security hole (same-origin backend paths, no secrets), but it violates the stated contract and exposes internal routing/curation. Current tests only assert names + length, so this isn't caught.

**Suggestion** — project when appending, and add an assertion that the advertised entry has no `binding`/`menu`:

```ts
const advertised = platformTools.map(({ name, description, inputSchema }) => ({ name, description, inputSchema }));
const toolDefinitions = advertised.length > 0 ? [...TOOL_DEFINITIONS, ...advertised] : TOOL_DEFINITIONS;
```

(`platformByName` keeps using the original `platformTools` for dispatch, so routing is unaffected.)

### 🟢 Low — inconsistent degradation: a malformed `binding` drops the whole tool, a malformed `menu` doesn't

`toPlatformToolEntry` returns `null` (drops the entire tool) when `binding` is present but `method`/`path` aren't both strings (see the `bad_binding` test). A malformed `menu`, by contrast, only drops the menu and keeps the tool — justified as *"a partially-rolled-out backend must never lose a working tool."*

The tension: `resolveToolEndpoint` already handles bad bindings gracefully at call time (unknown method / off-origin path both fall back to `POST /api/mcp/tools/<name>`). Since the runtime already tolerates a bad binding, rejecting the whole tool at validation time contradicts the "never lose a working tool" principle applied to `menu`. A more consistent choice would be to degrade a bad binding too (drop the binding, keep the tool → falls back). Design call, not a bug — flagging to confirm it's intentional.

### ⚪️ Nit — `invokePlatformTool` resolves auth/origin twice

`invokePlatformTool` calls `resolveAuth()` + `parseBaseUrl` to get the origin for `resolveToolEndpoint`, then `this.call()` does `resolveAuth()` + `parseBaseUrl` again internally. Correct, just one redundant round-trip. No real impact.

### Highlights

- **SSRF guard is correct**: `resolveToolEndpoint` compares the *resolved* WHATWG URL origin, not a raw string prefix, with tests for parser-bypass vectors (`/\host`, embedded tab/CR/LF, protocol-relative). This is exactly what prevents a manifest from redirecting the bearer token off-origin — consistent with the save-time origin allowlist.
- **Collision/dedup**: built-ins always win a name collision; duplicate manifest names are first-wins; `tools/list` and `tools/call` views stay consistent — all tested.
- **Gate + prompt capability**: when the menu is empty, no prompt is registered and the `prompts` capability isn't advertised — byte-identical to a tools-only server.

Recommend addressing the Medium item (align with the stated contract + add a test); the Low item is a judgment call for the author.

dougschroeder:
Thanks — all three addressed.

**🟡 Medium (binding/menu in `tools/list`).** Fixed; you're right it broke the stated contract. The advertised list now projects each platform tool to its public schema (`name`/`description`/`inputSchema`) instead of spreading the full manifest entry, so `binding` and `menu` never reach a client's `tools/list`. Dispatch still resolves against the full entries (kept in the by-name map), so routing and menu-building are unchanged. Added a test asserting the advertised entry is exactly `{name, description, inputSchema}` (no `binding`, no `menu`) while the executor still receives the full entry with its binding.

**🟢 Low (inconsistent degradation).** Agreed — made it consistent. A malformed `binding` no longer drops the whole tool; it's discarded and the tool stays callable, falling back to the conventional `POST /api/mcp/tools/<name>` endpoint at call time — the same fallback a structurally-valid but off-origin / unknown-method binding already takes. This mirrors the field-granular `menu` degradation: `binding` is internal metadata that never reaches `tools/list`, so a bad one can't poison it and there's no reason to lose a working tool over it.

**⚪️ Nit (double auth/origin resolution).** Left as-is by choice. The call needs the origin *before* it can hit the shared request path (to resolve + same-origin-check the binding), and that shared path re-resolves auth internally for every method. Both are local config parses (no network), so threading a pre-resolved origin through the shared path would widen its signature for every caller to save microseconds — not worth the coupling.

flyerli2025:
## Review

Overall this is high quality — thorough docs, opt-in gating, an SSRF guard with bypass-vector tests, and complete best-effort degradation. One issue worth resolving before merge, plus a few observations.

### 🟡 Medium — a `GET` binding always throws at call time (declared-supported but unusable)

`ALLOWED_TOOL_METHODS` lists `GET` as a valid binding method, but `invokePlatformTool` forwards `args` as a body unconditionally:

```ts
const { status, json, parseFailed } = await this.call<unknown>(method, path, args); // args always passed
```

`McpServer` invokes with `args ?? {}`, so the body is at minimum `"{}"` and always `!== undefined` — `call()` then `JSON.stringify`s it onto `body`. Node's `fetch` rejects a GET with a body:

```
TypeError: Request with GET/HEAD method cannot have body.   (verified locally)
```

So any tool advertising `binding: { method: "GET", ... }` throws before the request even hits the network (caught and surfaced as an MCP error response). This is true even for a zero-arg GET tool, since the body is always `"{}"`. And even if it didn't throw, there's no path that encodes args into the query string for GET — the args would never reach the backend. GET is effectively doubly-unusable.

The `invokePlatformTool` tests don't catch this because they all inject a `fetchImpl` stub (no GET-body enforcement) and none exercises a GET binding.

Suggested directions (your call):
- Drop `GET`/`HEAD` from `ALLOWED_TOOL_METHODS` (args can't be carried anyway) so they safely fall back to `POST /api/mcp/tools/<name>`; **or**
- Special-case GET/HEAD: omit the body and encode args as a query string.

Adding a GET repro test that runs against real `fetch` (or asserts no body is set) would pin the contract first.

### 🟢 Low / observations (non-blocking)

- `startMcpServer` now always calls `loadConfig()` even when the gate is off. Confirmed safe — `loadConfigFromDir` returns `{}` on a missing file, and gate-off behavior is otherwise byte-identical (same `TOOL_DEFINITIONS` reference, no prompts capability).
- Gate-on + slow backend adds up to `MANIFEST_TIMEOUT_MS` (5s) to startup; documented and degrades to "no platform tools". Acceptable.
- `call()`'s method param widened from the `"GET" | "POST" | "DELETE"` union to `string`, so existing internal callers lose literal narrowing. Acceptable since the platform path is validated via `ALLOWED_TOOL_METHODS`.

### ✅ Nicely done

- `resolveToolEndpoint` compares the **normalized `resolved.origin`** rather than a raw string prefix, blocking `/\host`, embedded tab/CR/LF, and similar URL-parser rewrite vectors that would otherwise leak the bearer token off-origin — with matching tests.
- `fetchManifest` collapses every failure mode to `[]` and never throws; a single malformed entry degrades at field granularity so a bad `binding`/`menu` never drops an otherwise-valid tool.
- Internal `binding`/`menu` metadata is stripped in the `advertisedPlatformTools` projection so it never reaches `tools/list`.
- Name-collision handling (built-in wins, manifest de-dup) logs a `warn` with clear rationale rather than dropping silently.

dougschroeder:
Thanks — the GET-binding issue is real and now fixed.

**🟡 GET binding throws at call time** — confirmed valid. The tool-call path always relays `args` as a JSON body, and since the layer above passes `args ?? {}`, the body is at minimum `"{}"`; Node's `fetch` then rejects the GET before it hits the network. I took your first suggested direction: dropped `GET` (and `HEAD`, which wasn't actually in the set) from `ALLOWED_TOOL_METHODS`, so a GET-natured binding now falls back to the conventional `POST /api/mcp/tools/<name>` endpoint that every tool supports — consistent with how an unknown-method or off-origin binding already degrades. I chose this over special-casing GET with query-string args because that conventional POST endpoint is the universal contract, and query-string encoding of arbitrary nested args would be lossy with no current consumer. Added a test pinning a GET binding to the POST fallback (asserts method, URL, and that the args body still rides along).

On the non-blocking observations — no action needed, and thanks for confirming each: the gate-off `loadConfig()` call is intentional (returns `{}` on a missing file, byte-identical behavior), the manifest-timeout startup cost is documented and degrades to no platform tools, and widening `call()`'s method param to `string` is fine since the platform path is validated through `ALLOWED_TOOL_METHODS`.

flyerli2025 (dismissed):

jeverest (approved):

PR #315, opened 2026-07-13
Fix pre-push worker entrypoint detection and destroy stdin stream

Finn-coder2026:


Finn-coder2026:
## Review summary

This PR fixes a `pre-push` hook blocking issue caused by incorrect worker entrypoint detection in bundled builds.

### Before the fix

When the VS Code extension was bundled as CommonJS, `PrePushWorker` could be inlined into `PrePushHook.js`. The previous `isMainScript()` check only compared `process.argv[1]` with `import.meta.url`, so the inlined worker could incorrectly identify itself as the directly executed entrypoint.

As a result, `processPushPending()` could run in the foreground inside the `pre-push` hook. A `git push` could therefore become noticeably slow or appear stuck while waiting for Jolli Space synchronization, especially when the network or service was slow or unavailable.

The stdin stream also remained open after the hook finished reading its input, which could prevent the process from exiting cleanly.

### After the fix

The entrypoint checks now also validate the executable filename. `PrePushWorker` only starts automatically when the actual entry file is `PrePushWorker.js` or `PrePushWorker.ts`, and the same protection is applied to `PrePushHook`.

The hook now performs only the lightweight work needed for the push, records the pending operation, and starts the worker as a detached background process. The stdin stream is explicitly destroyed after it reaches `end`.

The user-visible result is that `git push` no longer waits for Jolli Space synchronization to finish, while the synchronization still continues in the background.

### Review notes

The implementation is consistent with the existing `QueueWorker` entrypoint guard, and I did not find a blocking correctness issue.

One remaining coverage gap is that there is no direct regression test proving that the bundled `PrePushWorker` does not auto-start from inside `PrePushHook.js`. It would also be useful to add `PrePushHook` and `PrePushWorker` to the runtime bundle lists in `QueueWorker.bundle.test.ts` and `BuildEntryContract.test.ts` so this packaging contract is checked going forward.

flyerli2025 (approved):

PR #324, opened 2026-07-15
Add host-side local workflow-run orchestration

dougschroeder:
## Quick recap

This change lands the host side of running a Jolli workflow locally: a person's
own AI client runs the workflow's recipe on their machine (no Jolli LLM budget),
and the output is published to a git-backed destination space as a branch and
pull request that space-cli opens locally. A helper command reports which
workflows are eligible to run locally, a stub prints an install prompt when the
companion space-cli plugin is missing, and a new guided skill walks an assistant
through starting a run, checking out the right branch, pausing for a human's
review and approval, then publishing the result as a PR that either auto-merges
or waits for team review.

The eligibility parsing was also aligned to the backend's real response shape —
a numeric workflow id and a JRN-based destination identity — after an earlier
assumed shape caused the workflow list to always come back empty. Live
end-to-end against a real backend and space-cli is out of scope here.

---

## Topics (4)
<details>
<summary><strong>01 · Align local workflow-run eligibility to the backend's real response shape</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

The "run a workflow locally" helper always reported zero available workflows
even when an eligible one existed, because the parser assumed a response format
the backend doesn't actually send.

- **Reconcile the CLI to the backend's real, already-shipped shape** rather than
  asking the backend to add fields it doesn't have — the live response is
  treated as ground truth, confirmed against the running system.
- **Fold the correction directly into the still-unreleased code** instead of
  layering a separate patch on top, keeping the history clean.
- **Degrade gracefully where a companion tool isn't ready yet** rather than
  fabricating a workaround — report "nothing runnable" or "install required"
  until that tool catches up.

**✅ What Was Implemented**

- Rewrote the workflow-list parsing to accept a numeric workflow id (carried
  as-is instead of forced to text), and to read the destination's sync protocol
  and JRN from the fields the backend actually returns.
- Re-keyed the eligibility and offer logic off the space JRN (or the slug
  encoded in it) — the backend emits no numeric space id — matching the full
  identifier or the short name against the machine's cloned spaces.
- Rewrote the affected tests against the real captured response shape,
  preserving every malformed-data and failure-mode case.

**📁 FILES**
- `cli/src/core/JolliMemoryPushClient.ts`
- `cli/src/core/LocalRunEligibility.ts`
- `cli/src/core/LocalRunOffer.ts`
- `cli/src/commands/LocalRunOfferCommand.ts`

</blockquote>
</details>
<details>
<summary><strong>02 · Build the host eligibility helper command</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Expose the eligibility check to the guiding recipe as a single command an AI
client can call to learn which workflows are runnable locally right now.

- **A CLI helper the recipe shells, not a built-in platform tool** — the
  tool-dispatch path had no clean way to inject the network client, and spawning
  a subprocess from inside the stdio server risks corrupting its JSON-RPC
  channel.
- **Short-circuit to an empty offer before reading clones** when there are no
  candidate workflows at all, so the install prompt appears only when workflows
  actually exist.

**✅ What Was Implemented**

- A pure eligibility function that marks a workflow runnable only when its
  destination is git-backed and already cloned locally, echoing whether the
  resulting PR would auto-merge as an informational signal (never a factor in
  runnability itself).
- A best-effort workflow-list fetch that fails safe to an empty list under any
  error condition (feature disabled, tool missing, transport failure) rather
  than throwing.
- A `jolli local-run-workflows` command that reads the machine's cloned spaces
  and prints the offerable set as JSON; the clones read spawns through the
  run-cli indirection (or node on the resolved `Cli.js` on Windows) rather than
  a bare `jolli` on `PATH`, which matters for sandboxed and GUI-launched
  environments.

**📁 FILES**
- `cli/src/core/LocalRunEligibility.ts`
- `cli/src/core/LocalRunOffer.ts`
- `cli/src/commands/LocalRunOfferCommand.ts`
- `cli/src/install/McpRegistration.ts`

</blockquote>
</details>
<details>
<summary><strong>03 · Add a docs command stub for the space-cli plugin</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

Local runs depend on a `docs` command from the companion space-cli plugin to
pull and publish files; without the plugin installed, the command previously
failed with a confusing generic error instead of a helpful install prompt.

- **Reuse the existing generic stub mechanism** rather than bespoke handling —
  it already provides argument forwarding, help-menu grouping, and
  collision-tolerant registration for the other space-related stubs.

**✅ What Was Implemented**

- Added a `docs` stub alongside the existing space-related command stubs so
  that, when the plugin isn't installed, `jolli docs pull --branch <b>` and
  `jolli docs publish` print a clear install instruction and exit non-zero
  instead of a Commander "unknown command/option" error.

**📁 FILES**
- `cli/src/commands/SpaceCommandStubs.ts`

</blockquote>
</details>
<details>
<summary><strong>04 · Add the guided skill that walks an assistant through a local run</strong></summary>
<br>
<blockquote>

**⚡ Why This Change**

With the eligibility check and command groundwork in place, the final piece is
step-by-step guidance an AI assistant can follow to carry out a local workflow
run from start to finish.

- **Guide an existing assistant through the steps rather than build a fully
  automated command** — writing arbitrary workflow output requires genuine
  reasoning ability that only an assistant already capable of writing code and
  files has; a plain command can't generate that content itself.
- **Send status pings only right before and right after the human review, not
  throughout** — an assistant's turn is fully paused while awaiting a person, so
  periodic pings during the wait aren't possible; the design relies on the wait
  window being generous enough to cover typical review time.

**✅ What Was Implemented**

- Authored a new installable skill that walks an assistant through the full
  sequence: check which workflows can run locally, announce whether the result
  will auto-merge or need review, start the run, check out the correct branch,
  write the output, pause for human approval with status pings before and after
  (not during) the wait, then publish and complete the run.
- Explicitly warned against a destructive branch-checkout mode that wipes
  uncommitted files, and clarified that the file-pull step handles
  authentication internally so the assistant never fetches or handles any access
  token itself.
- Listed the new skill in the umbrella `/jolli` menu alongside the existing
  recall, search, and pull-request skills.

**📁 FILES**
- `cli/src/install/SkillInstaller.ts`

</blockquote>
</details>

---

*Generated by Jolli Memory · July 14, 2026 at 11:13 PM · via Jolli proxy*

### DIFF
--- a/cli/src/core/LocalRunOffer.ts
+++ b/cli/src/core/LocalRunOffer.ts
@@ -1,0 +1,123 @@
+/**
+ * Combines the candidate workflow list, the machine's cloned-space ids, and the
+ * pure {@link evaluateLocalRunEligibility} rule into the offer the local-run
+ * recipe consumes: the set of workflows that can be run locally right now.
+ *
+ * The I/O is injected ({@link LocalRunOfferDeps}) so the branching is fully
+ * unit-testable without a live backend or a real space-cli spawn. The command
+ * surface (`local-run-workflows`) wires the real deps and prints the result.
+ */
+
+import { evaluateLocalRunEligibility, type WorkflowSummary } from "./LocalRunEligibility.js";
+
+/** The combined install hint surfaced when space-cli is required but absent. */
+export const SPACE_CLI_INSTALL_HINT = "npm i -g @jolli.ai/cli @jolli.ai/space-cli";
+
+/** One offerable workflow: the id the recipe passes to `start_local_run`, plus its advisory auto-merge signal. */
+export interface OfferableWorkflow {
+	/** Carried verbatim from the backend (a numeric id) so it stays usable as the integer `start_local_run` expects. */
+	readonly id: string | number;
+	/**
+	 * The workflow's human-readable name, when the backend supplied one, so the
+	 * recipe can present a multi-workflow choice by name instead of an opaque id.
+	 * Advisory display only — never a factor in what is offered.
+	 */
+	readonly name?: string;
+	/** Whether an approved PR will auto-merge (`true`) or open for team review (`false`). */
+	readonly autoMerges: boolean;
+}
+
+/**
+ * Type-tagged offer outcome. `workflows` (possibly empty) is the normal success —
+ * an empty list is a legitimate "nothing to offer" state (platform tools off, or
+ * no eligible workflow), never an error. `space_cli_required` is a "needs input"
+ * result (mirrors the `binding_required` idiom): space-cli must be installed
+ * before eligibility can be computed.
+ */
+export type LocalRunOfferResult =
+	| { readonly type: "workflows"; readonly workflows: OfferableWorkflow[] }
+	| { readonly type: "space_cli_required"; readonly message: string; readonly install: string };
+
+export interface LocalRunOfferDeps {
+	/** Best-effort candidate workflow list (empty on any degrade; never throws). */
+	readonly listWorkflows: () => Promise<WorkflowSummary[]>;
+	/** The cloned-space keys (JRNs and/or slugs) on this machine, or `null` when space-cli is unavailable. */
+	readonly readClonedSpaceKeys: () => Promise<Set<string> | null>;
+}
+
+/**
+ * Resolves the offer. When there are no candidate workflows the result is an
+ * empty offer WITHOUT consulting the clones — the blocker there is the absent
+ * workflow list (platform tools off / none served), not space-cli, so surfacing
+ * an install prompt would mislead. Only when candidates exist does a missing
+ * space-cli become the `space_cli_required` needs-input result.
+ */
+export async function resolveLocalRunOffer(deps: LocalRunOfferDeps): Promise<LocalRunOfferResult> {
+	const workflows = await deps.listWorkflows();
+	if (workflows.length === 0) {
+		return { type: "workflows", workflows: [] };
+	}
+	const clonedSpaceKeys = await deps.readClonedSpaceKeys();
+	if (clonedSpaceKeys === null) {
+		return {
+			type: "space_cli_required",
+			message: "Running a workflow locally needs the Jolli space-cli plugin. Install it with:",
+			install: SPACE_CLI_INSTALL_HINT,
+		};
+	}
+	const offerable = evaluateLocalRunEligibility(workflows, clonedSpaceKeys)
+		.filter((verdict) => verdict.runnable)
+		.map((verdict) => ({ id: verdict.id, name: verdict.name, autoMerges: verdict.autoMerges }));
+	return { type: "workflows", workflows: offerable };
+}
+
+/**
+ * Parses the cloned-space keys from `jolli space clones --json` output. Accepts a
+ * bare array or a `{ clones: [...] }` envelope; each entry contributes its space
+ * `jrn` and/or `slug` (whichever it exposes) as a string key, matched against a
+ * workflow destination's JRN or its encoded slug. Defensive by contract: a
+ * non-JSON body (a broken space-cli, a proxy page) yields an empty set — treated
+ * as "no clones" (nothing runnable) rather than an error — and entries with no
+ * usable string key are skipped.
+ *
+ * NOTE: the exact `jolli space clones --json` output shape is owned by
+ * `@jolli.ai/space-cli` (a separate package). This parser keys on the space
+ * identity the backend actually emits (JRN / slug), never a numeric id the
+ * backend does not produce; a clones command that exposes neither yields no
+ * matches (and must gain a JRN/slug key for local runs to be offerable).
+ */
+export function parseClonedSpaceKeys(stdout: string): Set<string> {
+	const keys = new Set<string>();
+	let json: unknown;
+	try {
+		json = JSON.parse(stdout);
+	} catch {
+		return keys;
+	}
+	for (const entry of extractCloneArray(json)) {
+		if (entry !== null && typeof entry === "object") {
+			const { jrn, slug } = entry as { jrn?: unknown; slug?: unknown };
+			if (typeof jrn === "string" && jrn.trim() !== "") {
+				keys.add(jrn);
+			}
+			if (typeof slug === "string" && slug.trim() !== "") {
+				keys.add(slug);
+			}
+		}
+	}
+	return keys;
+}
+
+/** Pulls the clone array out of the body — a bare array or a `{ clones: [...] }` envelope. */
+function extractCloneArray(json: unknown): unknown[] {
+	if (Array.isArray(json)) {
+		return json;
+	}
+	if (json !== null && typeof json === "object") {
+		const clones = (json as { clones?: unknown }).clones;
+		if (Array.isArray(clones)) {
+			return clones;
+		}
+	}
+	return [];
+}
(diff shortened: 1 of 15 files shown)


cli/src/install/SkillInstaller.ts
```diff
@@ -934,6 +935,149 @@ different phrasing. Do NOT mention BM25 or index internals.
...
+- \`{ "type": "error", "message": "..." }\` — report the message and stop.
+
+Have the user pick one workflow (use your host's interactive single-select tool
+if it has one — e.g. AskUserQuestion on Claude Code — otherwise list them as
+text).
+
+## Step 2 — start the run
+
+Call the \`start_local_run\` tool (on Claude Code
+\`mcp__jollimemory__start_local_run\`) with the chosen \`{ "id": "<workflow id>" }\`.
```
flyerli2025: **[Low — doc clarity]** The `start_local_run` call template quotes the id as a string: `{ "id": "<workflow id>" }` (Step 1's example likewise shows `{ "id": "..." }` at line ~982). This is inconsistent with the rest of the PR, which deliberately preserves workflow ids as numbers because `start_local_run` expects an integer (`LocalRunEligibility.ts` WorkflowSummary doc, `JolliMemoryPushClient.ts` `resolveWorkflowId`, and the test asserting `typeof id === "number"`).

Note this is an LLM instruction, not a deterministic coercion — at call time the tool's own inputSchema is authoritative, and the runtime `local-run-workflows` output prints `"id": 7` (a number), so a model will most likely pass a number regardless. So it's unlikely to actually fail schema validation. But the string-quoted placeholder mirrors the genuinely-string `{ "runId": "<runId>" }` style and is misleading. Suggest dropping the quotes: `{ "id": <workflow id> }`, and fixing the Step 1 example, with a note to pass the id verbatim as returned.

cli/src/install/SkillInstaller.ts
```diff
@@ -934,6 +935,149 @@ different phrasing. Do NOT mention BM25 or index internals.
...
+Drive the steps below in order. Prefer the Jolli MCP tools for the run lifecycle;
+the eligibility check and the git operations go through the \`jolli\` CLI (via the
+run-cli entry script the sibling skills also use).
+
+## Step 1 — discover the runnable workflows
+
+Run the eligibility helper and read its JSON:
+
+\`\`\`bash
+"$HOME/.jolli/jollimemory/run-cli" local-run-workflows
```
flyerli2025, line 1026: **[Medium]** This skill shells `"$HOME/.jolli/jollimemory/run-cli"` (here, and at the `docs pull` / `docs publish` blocks) but the template carries no Windows shell prerequisite. The sibling shell-backed skills go through `heredocInvocation`, which prepends a "Shell prerequisite" block pinning **Git Bash** on Windows and explaining that `run-cli` is written to `%USERPROFILE%\.jolli\...` — only Git Bash's `$HOME` aligns; PowerShell / WSL bash see a different home and won't find the entry script.

Without that block, on a Windows host defaulting to PowerShell or WSL bash this skill instructs commands that can't locate the installed Jolli entry script. Suggest adding the same (or a trimmed) prerequisite block before Step 1.

(Minor accuracy note: the prerequisite isn't universal even today — the `push` fallback one-liner omits it too — but recall/search/pr-description all carry it, and local-run's bash steps genuinely hit this on Windows.)

cli/src/core/LocalRunOffer.ts
```diff
@@ -0,0 +1,117 @@
...
+	if (clonedSpaceKeys === null) {
+		return {
+			type: "space_cli_required",
+			message: "Running a workflow locally needs the Jolli space-cli plugin. Install it with:",
+			install: SPACE_CLI_INSTALL_HINT,
+		};
+	}
+	const offerable = evaluateLocalRunEligibility(workflows, clonedSpaceKeys)
+		.filter((verdict) => verdict.runnable)
+		.map((verdict) => ({ id: verdict.id, autoMerges: verdict.autoMerges }));
```
flyerli2025: **[Medium — UX]** The offer is reduced to `{ id, autoMerges }` only. The real `list_workflows` response includes a human-readable workflow `name` (e.g. "Impact Analysis"), but it's dropped twice — first at parse time in `JolliMemoryPushClient` `mapWorkflowEntry` (keeps only `{id, destination}`), then here in the offer map. The skill then asks the user to "pick one workflow" (Step 1) from this reduced payload.

With a single runnable workflow this is fine, but a tenant with 2+ runnable workflows leaves the agent presenting only opaque numeric ids (7, 12, …) — selecting the right one becomes guesswork. The code comment marks the reduction as intentional, but intentional minimalism here costs usability. Suggest threading `name` through `WorkflowSummary` → `OfferableWorkflow` so the skill can present workflows by name.

flyerli2025 (commented):
Review of the local-run skill + offer wiring. Three findings on the `jolli-local-run` skill template and the offer payload — details inline. Severities: one Low (doc clarity), two Medium.

flyerli2025 (dismissed):

dougschroeder:
Thanks for the review — all three were valid and are addressed in the updated commit.

**1. `start_local_run` id shown as a quoted string (Low).** Fixed. The recipe now passes the id exactly as the helper returns it — the backend's numeric id stays an unquoted number (`{ "id": <workflow id> }`), a string id/slug stays quoted — and the discovery example shows the real numeric shape (`"id": 7`) instead of `"id": "..."`.

**2. No Windows shell prerequisite (Medium).** Fixed. The Git Bash prerequisite is extracted to a single shared block and included in the local-run recipe, whose `run-cli` bash steps hit `%USERPROFILE%` the same way the sibling shell-backed skills do. Sibling skill output is unchanged.

**3. Workflow `name` dropped, leaving opaque numeric ids (Medium).** Fixed. The human-readable `name` is now threaded through the candidate parser → eligibility verdict → offer → the `local-run-workflows` JSON, and the recipe presents workflows by `name` (falling back to the id when absent). It stays advisory display only — never an eligibility input — and a missing/blank/non-string name is dropped.

Verification: full build + lint + tests green across both workspaces; coverage holds, with new tests covering each layer.

cli/src/install/SkillInstaller.ts
dougschroeder: Fixed — the recipe now passes the id exactly as the helper returns it: the backend's numeric id stays an unquoted number (`{ "id": <workflow id> }`), a string id/slug stays quoted, and the Step 1 example now shows the real numeric shape (`"id": 7`) instead of `"id": "..."`.

cli/src/install/SkillInstaller.ts
dougschroeder: Fixed — extracted the Git Bash shell-prerequisite block into a single shared source and included it in the local-run recipe before Step 1, since its `run-cli` bash steps hit `%USERPROFILE%` the same way the sibling shell-backed skills do. The sibling skills reuse the same shared block, so their output is unchanged.

cli/src/core/LocalRunOffer.ts
dougschroeder: Fixed — the human-readable `name` is now threaded through the candidate parser → the eligibility verdict → the offer → the `local-run-workflows` JSON, and the recipe presents workflows by `name` (falling back to the id when absent). It stays advisory display only — never an eligibility input — and a missing/blank/non-string name is dropped.

flyerli2025 (approved):

PR #327, opened 2026-07-15
Add canPush check to warn on read-only Space bindings

Finn-coder2026:


Finn-coder2026:
## Summary

This PR makes the CLI accurately report whether the current repository can sync memories to its bound Jolli Space. It detects bindings that still exist but are no longer writable, warns before the next push would fail with a 403, and offers an interactive rebind when another writable Space is available.

## What changed

### Front-door client contract

- Extended bound responses with `binding.canPush`, which mirrors the server-side `articles.edit` permission enforced by the push endpoint.
- Treats `canPush: false` as a read-only binding and treats a missing or non-boolean value as `null` so older servers remain compatible and do not produce false warnings.
- Preserves `spaces` and `defaultSpaceId` on degraded bound responses, allowing the CLI to receive rebind choices in the same front-door request.
- Added optional `replace: true` support to `createBinding` for the server-controlled rebind flow.

### Guided `jolli` Space-sync flow

- Healthy bindings now render a clear status such as `✓ syncing · Space "Acme Core"`.
- A binding whose Space details are hidden because the caller lacks `spaces.view` now shows a no-access warning instead of claiming that syncing is active.
- A visible binding with `canPush: false` now shows a read-only warning and explains that memories will not sync.
- When the server returns writable alternatives, the CLI offers an opt-in rebind:
  - defaults to No;
  - handles a single destination directly;
  - supports choosing among multiple Spaces and honors the default Space;
  - sends `replace: true` only after explicit confirmation;
  - treats a concurrent 409 that already selected the same Space as success;
  - fails safely with a retry hint when the rebind cannot be confirmed.
- Space names are explicitly labelled and quoted, and availability wording is caller-relative because the server filters Spaces by permission.

### `jolli status`

- Added a human-readable `Jolli Space:` row backed by one best-effort front-door request with a five-second timeout.
- Reports healthy, read-only, no-access, unbound, no-Spaces, missing-key, rejected-key, outdated-client, and unreachable states.
- Shows `run jolli to rebind` only when a bindable alternative is available; otherwise it directs the user to request access.
- Performs no Space request when no Jolli API key is configured.
- Keeps `--json` unchanged and network-free, preserving existing machine-readable behavior.

### Resilience and test coverage

- Empty unbound Space lists are folded into the no-Spaces state to handle contract drift safely.
- Network, authentication, and compatibility failures remain best-effort and do not break the guided front door or status command.
- Added coverage for older-server compatibility, malformed `canPush` values, permission loss, single- and multi-target rebinds, default-No behavior, 409 races, retry behavior, all status variants, timeout/network failures, and `--json` isolation.

The DCO sign-off has also been corrected, and the updated commit is now on the PR branch.

flyerli2025 (approved):

PR #366, opened 2026-07-22
Move workflow commands to a plugin seam and namespace the skill recipes

dougschroeder:
## Summary

Converts the host CLI's workflow-run surface into a discovered plugin seam, so the `workflow` commands are provided by the `@jolli.ai/workflow-cli` plugin rather than being unconditional host builtins.

- Adds a `KNOWN_PLUGINS` entry for the workflow plugin with a new "Jolli Workflows" `--help` group, and a stub (`WorkflowCommandStubs`) registered when the plugin is absent. The `workflow local-run` stub emits a machine-readable `{"type":"workflow_cli_required","installHint":...}` object on stdout (parsed by the local-run recipe); `workflow runs` / `workflow run-status` print a prose install hint and exit non-zero. The stub keeps `workflow` visible in `--help` either way.
- Removes the moved builtins: the `local-run` / `runs` / `run-status` and `verify-publish-branch` commands, their command/core files and tests, and the workflow-only methods, run-payload parsers, and tool-name constants from the memory-push client.
- Points the first-party skill recipes at the namespaced names (`workflow local-run` / `workflow runs` / `workflow run-status`, and `space verify-publish-branch`) and bumps the edited skills' revisions so existing installs are rewritten. `open-url` remains a flat host builtin.

## Verification

- `npm run all` green (CLI + VS Code); CLI coverage thresholds held.
- Manual e2e both paths: plugin absent (stub emits the `workflow_cli_required` JSON on `local-run`, prose + non-zero on `runs`/`run-status`, `workflow` still shown under "Jolli Workflows") and plugin present (real `local-run`/`runs`/`run-status` subcommands, handled by the plugin).

> **Ordering:** ship this host change only after `@jolli.ai/workflow-cli` and the updated `@jolli.ai/space-cli` are published to npm, so no host skill calls a name no installed plugin provides.

---

*Generated by Jolli Memory · July 22, 2026*

---

## Follow-up — review round 2

- **Stale recipes on a CLI-only upgrade** — the CLI now refreshes on-disk skill recipes once per version on startup (version-guarded, fail-soft, skipped for lifecycle commands and un-enabled repos), so an upgrade without a re-enable self-heals; the removed flat command names emit a "moved / refreshed / re-run" notice instead of an unknown-command error.
- **Remote-run orphaning** — the remote-run recipe confirms the workflow-cli monitor is installed *before* triggering the backend run, and stops without creating a run when it's missing.


flyerli2025 (dismissed):

cli/src/install/SkillInstaller.ts
```diff
@@ -1200,7 +1200,7 @@ name: jolli-local-run
 description: Run a Jolli workflow locally — your own agent executes the workflow's recipe (no Jolli LLM budget) and its file writes land in a git-backed Jolli Space via a branch and pull request that space-cli opens on this machine. Use when the user wants to run a Jolli workflow locally.
 metadata:
   version: "${SKILL_VERSION}"
-  revision: 4
+  revision: 5
```
summerfang-jolli, line 1213: [P1] Preserve an upgrade path for already-installed recipes. Bumping this revision only helps when `updateSkillIfNeeded()` runs again (for example, via `jolli enable` or an IDE activation); a CLI-only user can update the global package without re-enabling, leaving the revision-4 local-run recipe and revision-2 remote-run recipe on disk. Those recipes still call `local-run-workflows`, `workflow-run-status`, and `workflow-runs`, all of which this PR removes, so workflow actions break immediately after the host upgrade even when the new plugins are installed. Please retain deprecated forwarding aliases for at least a compatibility window, or add an automatic skill-refresh path that runs as part of the upgrade before removing the old commands.

cli/src/install/SkillInstaller.ts
```diff
@@ -1492,6 +1499,14 @@ could not be reached — platform tools off, or a transport failure), tell the u
 the run status could not be retrieved and stop. That is a degraded outcome, not a
 crash — the run may still be progressing server-side.
 
+If instead the command exits non-zero and prints a prose install hint naming
```
summerfang-jolli, line 1541: [P1] Check for `@jolli.ai/workflow-cli` before triggering the remote run. The recipe calls `run_remote_workflow` in Step 2, so by the time this missing-plugin branch is reached the backend run has already been created and may be spending budget, while the user has no monitor and is told to stop. Please preflight the plugin/monitor before Step 2 (or keep a host-side monitor available during the migration) so a missing dependency cannot leave an active run orphaned.

summerfang-jolli (dismissed):
Two blocking migration issues remain: existing installed skills have no guaranteed refresh when the old host commands disappear, and the remote-run recipe discovers a missing monitor only after starting a server-side run. I also ran the branch's focused workflow/skill tests successfully. The full `npm run all` reached the CLI suite but was blocked by four Windows-environment failures in unchanged symlink/path tests (`PluginLoader.test.ts` and `UninstallScan.test.ts`).

dougschroeder:
Thanks — both points were valid and are addressed in `29f1db40` (each verified against the code first).

**1. Stale recipes after a CLI-only upgrade.** Correct: refresh only happened on the next enable/activation, so upgrading the global package without re-enabling left recipes calling the removed flat commands. Fixed at the source — the CLI now refreshes the on-disk skill recipes once per version on startup: version-guarded by a per-repo marker (a single small file read on the hot path), fail-soft, and skipped both for the lifecycle commands and for repos where the tool was never enabled (so it never creates recipes uninvited). As a belt-and-suspenders for the in-flight call that triggers the upgrade, the three removed flat names (`local-run-workflows`, `workflow-run-status`, `workflow-runs`) now resolve to a hidden notice that says the command moved, that skills were just refreshed, and to re-run — instead of a bare "unknown command" error. A CLI-only upgrade now self-heals on the next invocation, no re-enable required.

**2. Remote run orphaned when the monitor is missing.** Correct — the run trigger is a backend tool that spends budget even when the monitor isn't installed, and the old check only fired after the run was created. The remote-run recipe now confirms the monitor is present *before* triggering, using the local-run eligibility helper purely as a presence probe (only the absent-plugin stub emits the "plugin required" signal; any other result means it's installed). When it's missing the recipe tells the user to install it and stops **without creating a run**, so a missing dependency can no longer orphan an active run. The post-trigger check stays as a backstop.

Verification: build, lint, and typecheck are green across the affected packages, and the CLI suite passes with coverage held above threshold (the new code is fully covered). Unrelated note: the full suite has a pre-existing failure in a cursor-session test whose fixed fixture timestamp is now older than its 48-hour recency window — it fails identically on the prior commit, independent of these changes.

summerfang-jolli (approved):

PR #420, opened 2026-07-31
Fix Windows-only test failures so npm run all passes cross-platform

summerfang-jolli:
## Summary

`npm run all` failed on Windows with **18 test failures** across 9 files. Every one stemmed from host-vs-target platform assumptions in the tests (and one source helper), not real regressions — the branch is a clean checkout of `main`, so these are the current state of `main` on a Windows dev box.

## Changes

**Source (1 file, no production behavior change):**
- `UninstallScan.getVendorConfigRoot` now joins with the path flavor matching the target `platform` argument (`path.win32`/`path.posix`) instead of the host `path`, following the same convention `ExecutableResolver`'s `knownPaths` already documents. In production `platform === host`, so it's a no-op there; it just makes the JetBrains/vendor root paths deterministic under a platform-pinned unit test on any OS.

**Tests (9 files):**
- **Path-separator assertions** (`IdeBridgeCommand`, `PostInstall`, `SharedBranchImporter`) — build expected paths with `join(...)` instead of hardcoded POSIX literals.
- **`CreatePrWebviewPanel` `openDiff`** — pass `resolve("/repo")` as the workspace root so the inside-workspace guard matches on Windows, consistent with the sibling tests.
- **`ClaudeExecutableResolver`** — return a host-appropriate finder path so win32 `discover()`'s `.exe` filter accepts it.
- **`UninstallScan` `scanCliGlobal`** — create a regular file (the scanner only `lstat`s the shim) instead of a symlink, which needs Developer Mode on Windows.
- **`PluginLoader`** symlink-forensics test — use the existing `itIfSymlinks` guard (its sibling already did).
- **`GitHookInstaller`** (exec-bit repair) and **`Installer`** (read-only rename) — `skipIf(win32)`; both assert POSIX-only filesystem semantics Windows fundamentally lacks.

## Verification

`npm run all` exits **0** (clean → build → lint → test):
- CLI: 9063 passed / 57 skipped
- vscode: 4430 passed / 10 skipped
- acceptance: 9 passed
- Coverage floors met (CLI branches 96.34% ≥ 96%)

No production behavior changes; the source edit is a no-op wherever `platform === host`.

### DIFF
--- a/cli/src/install/Installer.test.ts
+++ b/cli/src/install/Installer.test.ts
@@ -3403,29 +3403,39 @@ describe("Installer", () => {
 	});
 
 	describe("reconcileClaudeAgentHooks — atomic write survives a read-only settings file", () => {
-		it("enable still succeeds when .claude/settings.local.json is read-only (atomic tmp+rename)", async () => {
-			// install() writes the canonical Claude hooks via reconcileClaudeAgentHooks,
-			// which uses atomicWriteFile (temp file + rename). rename replaces the target
-			// based on the PARENT DIRECTORY's permissions, not the target file's — so a
-			// read-only settings FILE does not block the rewrite. This guards that
-			// resilience (a plain writeFile to the same path would EACCES). It does NOT
-			// exercise a "writeFile catch" — reconcileClaudeAgentHooks has none on its
-			// write path; the resilience comes from the atomic rename.
-			await install(tempDir);
-			const settingsPath = join(tempDir, ".claude", "settings.local.json");
-			// Drop SessionStart so the reconcile has a real change to write, then lock the file.
-			const settings = JSON.parse(await readFile(settingsPath, "utf-8"));
-			delete settings.hooks.SessionStart;
-			await writeFile(settingsPath, JSON.stringify(settings), "utf-8");
-			await chmod(settingsPath, 0o444);
+		// POSIX-only: this exercises the fact that rename replaces a target based on
+		// the PARENT DIR's permissions, so a read-only target FILE doesn't block the
+		// rewrite. On Windows this can't hold: rename over a read-only file fails with
+		// EPERM, and although atomicWriteFile (core/AtomicWrite.ts) catches EPERM/EACCES
+		// from rename and retries with a plain writeFile, that fallback ALSO fails against
+		// a read-only target — so the win32 path has no way through, which is a different
+		// filesystem contract, not what this resilience guard is about.
+		it.skipIf(process.platform === "win32")(
+			"enable still succeeds when .claude/settings.local.json is read-only (atomic tmp+rename)",
+			async () => {
+				// install() writes the canonical Claude hooks via reconcileClaudeAgentHooks,
+				// which uses atomicWriteFile (temp file + rename). rename replaces the target
+				// based on the PARENT DIRECTORY's permissions, not the target file's — so a
+				// read-only settings FILE does not block the rewrite. This guards that
+				// resilience (a plain writeFile to the same path would EACCES). It does NOT
+				// exercise a "writeFile catch" — reconcileClaudeAgentHooks has none on its
+				// write path; the resilience comes from the atomic rename.
+				await install(tempDir);
+				const settingsPath = join(tempDir, ".claude", "settings.local.json");
+				// Drop SessionStart so the reconcile has a real change to write, then lock the file.
+				const settings = JSON.parse(await readFile(settingsPath, "utf-8"));
+				delete settings.hooks.SessionStart;
+				await writeFile(settingsPath, JSON.stringify(settings), "utf-8");
+				await chmod(settingsPath, 0o444);
 
-			try {
-				const result = await install(tempDir);
-				expect(result.success).toBe(true);
-			} finally {
-				await chmod(settingsPath, 0o755);
-			}
-		});
+				try {
+					const result = await install(tempDir);
+					expect(result.success).toBe(true);
+				} finally {
+					await chmod(settingsPath, 0o755);
+				}
+			},
+		);
 	});
 
 	describe("manual-disable persistence is fail-safe (behavioral)", () => {
--- a/cli/src/install/UninstallScan.test.ts
+++ b/cli/src/install/UninstallScan.test.ts
@@ -12,6 +12,15 @@ import { mkdirSync, mkdtempSync, readFileSync, rmSync, symlinkSync, writeFileSyn
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { symlinksSupported } from "../testUtils/symlinkSupport.js";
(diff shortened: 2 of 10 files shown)


cli/src/install/UninstallScan.test.ts
```diff
@@ -296,7 +296,10 @@ describe("scanCliGlobal", () => {
 		const root = join(tempDir, "lib", "node_modules");
 		mkdirSync(join(root, "@jolli.ai", "cli"), { recursive: true });
 		mkdirSync(join(tempDir, "bin"), { recursive: true });
-		symlinkSync(join(root, "@jolli.ai", "cli", "dist", "Cli.js"), join(tempDir, "bin", "jolli"));
+		// scanCliGlobal only `lstat`s the shim — a regular file exercises the
+		// same "found the bin" branch as a POSIX symlink would, and stays
+		// creatable on a Windows box without Developer Mode / symlink privilege.
+		writeFileSync(join(tempDir, "bin", "jolli"), "#!/usr/bin/env node\n");
```
aidancrosbie: **Nice to have:** the symlink this replaces was *dangling* — it pointed at `<root>/@jolli.ai/cli/dist/Cli.js`, which the test never creates (only the `@jolli.ai/cli` dir is `mkdir`'d). Dangling is the one state where `lstat` and `stat` diverge: `stat` throws ENOENT, `lstat` succeeds. A regular file satisfies both, so this was the only fixture that would fail if `scanCliGlobal`'s deliberate `lstat` (UninstallScan.ts:387, with a comment explaining the choice) were ever changed to `stat`.

**Why it matters:** `jolli uninstall` would silently stop listing the global `jolli` shim whenever the link is dangling — a partially-removed or mid-upgrade npm global install, i.e. exactly when uninstall matters. No behavior changes here; it's the future regression guard that's lost.

**Suggested fix:** keep the symlink and gate it the same way this PR gates the PluginLoader case:

```ts
import { symlinksSupported } from "../testUtils/symlinkSupport.js";
const itIfSymlinks = symlinksSupported ? it : it.skip;
```

That helper is account-level, so an elevated / Developer-Mode Windows box still runs it. If you'd rather keep the regular file, worth softening the comment — it's accurate for branch coverage but not for the behavioral contract.

cli/src/install/Installer.test.ts
```diff
@@ -3403,29 +3403,36 @@ describe("Installer", () => {
...
-			await install(tempDir);
-			const settingsPath = join(tempDir, ".claude", "settings.local.json");
-			// Drop SessionStart so the reconcile has a real change to write, then lock the file.
-			const settings = JSON.parse(await readFile(settingsPath, "utf-8"));
-			delete settings.hooks.SessionStart;
-			await writeFile(settingsPath, JSON.stringify(settings), "utf-8");
-			await chmod(settingsPath, 0o444);
+		// POSIX-only: this exercises the fact that rename replaces a target based on
+		// the PARENT DIR's permissions, so a read-only target FILE doesn't block the
+		// rewrite. On Windows, rename over a read-only file fails (EPERM), which is a
```
aidancrosbie: **Nice to have:** the rationale is slightly incomplete. `atomicWriteFile` (`cli/src/core/AtomicWrite.ts:28-35`) already catches `EPERM`/`EACCES` from `rename` and retries with a plain `writeFile` — that fallback exists for exactly this. The real reason the premise can't hold on win32 is that the *fallback* also fails against a read-only target.

**Why it matters:** as written, a future maintainer could read this as "the atomic-write helper is POSIX-naive" and add a redundant EPERM fallback that's already there.

**Suggested fix:** note that the EPERM path is handled and it's the plain-`writeFile` fallback that a read-only target defeats on Windows.

cli/src/install/UninstallScan.ts
```diff
@@ -198,13 +198,18 @@ export async function pruneVscodeExtensionManifest(extensionDir: string): Promis
 
 /** Per-platform config root for a JetBrains-platform vendor (`JetBrains`, `Google`). */
 export function getVendorConfigRoot(vendor: string, home: string, platform: NodeJS.Platform): string {
+	// Build the path with the flavor matching the TARGET platform, not the host,
+	// so a `platform`-pinned unit test yields the same string on any OS (the same
+	// convention ExecutableResolver's `knownPaths` follows). In production
+	// `platform === osPlatform()`, so this is a no-op there.
```
aidancrosbie: **Nice to have:** worth naming the concrete anchors for "the same convention ExecutableResolver's `knownPaths` follows" — the rule is written down in `ExecutableSpec.knownPaths`' docstring (`cli/src/core/localagent/ExecutableResolver.ts:34-39`), and `discoverPresence` (same file, ~line 323) already contains the identical `joinFor` line.

**Why it matters:** this is the only production line in the PR, so making the precedent one-click verifiable is worth the few extra characters.

**Suggested fix:** add the file/symbol references to the comment.

aidancrosbie (dismissed):
Approving. Verified the two claims that matter: the `getVendorConfigRoot` edit is a genuine no-op in production (`scanUninstallInventory` defaults `platform` to `osPlatform()` and the only production caller, `UninstallCommand.ts:139`, never overrides it), and it follows a convention already written down in `ExecutableSpec.knownPaths`' docstring and used verbatim by `discoverPresence`. CI is ubuntu-only across every workflow, so neither `skipIf(win32)` guard fires there and the coverage floors are untouched.

Three non-blocking suggestions inline — all Nice to Have, none should hold up the merge.

cli/src/install/UninstallScan.test.ts
summerfang-jolli: Good catch on the dangling-symlink contract — restored it. The fixture is a dangling symlink again (target Cli.js is never created) and gated with the existing `itIfSymlinks` helper, so it runs on Linux CI and elevated/Developer-Mode Windows and skips only on a plain Windows account. That keeps the lstat-vs-stat regression guard intact. Comment updated to explain the dangling intent. (8139c3b)

cli/src/install/Installer.test.ts
summerfang-jolli: You're right, the rationale was incomplete. Updated the comment to note that atomicWriteFile (core/AtomicWrite.ts) already catches EPERM/EACCES from rename and retries with a plain writeFile, and that the real reason win32 can't pass is that this plain-writeFile fallback ALSO fails against a read-only target — so there's no way through, which is a different filesystem contract than the parent-dir-permission resilience this guards. (8139c3b)

cli/src/install/UninstallScan.ts
summerfang-jolli: Added the concrete anchors to the comment: the convention is written down in `ExecutableSpec.knownPaths`' docstring (core/localagent/ExecutableResolver.ts:34-39) and applied verbatim by `discoverPresence` in the same file, whose `joinFor` line this mirrors. (8139c3b)

aidancrosbie (approved):

PR #426, opened 2026-08-02
cli: honor JOLLI_TELEMETRY_ENV=sandbox to self-tag E2B telemetry

summerfang-jolli:
## What

Part B of tagging E2B-sandbox telemetry as `env="sandbox"`. Gives the CLI a way to self-tag when it runs inside a sandbox, so sandbox traffic can be excluded downstream with a simple `WHERE env <> 'sandbox'`.

E2B sandbox runs mint fresh `install_id`s and emit `prod`-tagged telemetry (`app_installed` + a burst of `space *` / `ide-bridge` commands), inflating install / DAU / retention metrics — ~72% of prod installs measured are these sandboxes. Until now `env` was 100% origin-derived (`resolveTelemetryEnv(origin)`) with no override hook.

## Change

- `resolveTelemetryEnv` takes an optional `env: NodeJS.ProcessEnv = process.env` and checks `JOLLI_TELEMETRY_ENV === "sandbox"` **before** the origin logic.
- Both call sites — `initTelemetry` and `jolli telemetry status/inspect` — inherit it via the default param, no other edits.
- `env` is stamped once at init and flows to every event (incl. `app_installed`), so this one guard tags the whole session.
- Adds `"sandbox"` to the `TelemetryEnv` union.

## Tests

Added cases to `Telemetry.test.ts`: `JOLLI_TELEMETRY_ENV=sandbox` overrides a prod origin (and works with no origin); a non-`"sandbox"` value (and absent var) still falls through to origin logic. Full `npm run all` gate passes.

## Rollout note

This must only reach a running sandbox **after** Part A (backend `env` allowlist, separate repo `~/jolli.ai/jolli`) is deployed — otherwise events tagged `env="sandbox"` are silently dropped at ingest. Part C (sandbox env var / Dockerfile) also lives in that repo. Forward-looking only; historical rows stay `prod`.


jeverest (dismissed):

flyerli2025 (changes requested):
1. `cli/src/core/Telemetry.ts:87`
`initTelemetry()` still computes the cached telemetry environment with `resolveTelemetryEnv(init.origin)` and drops the `init.env` override that the same function already uses for consent. That means the new `JOLLI_TELEMETRY_ENV=sandbox` path only works when the flag is present in `process.env`; any caller that bootstraps telemetry with an injected env map will still cache origin-derived values for the whole session.

2. `cli/src/core/Telemetry.ts:34` and `cli/src/core/Telemetry.ts:191`
This module is explicitly documented as lockstep with the IntelliJ telemetry implementation, but the PR only changes the TS side. The Kotlin mirror still accepts only origin-derived environments in `intellij/src/main/kotlin/ai/jolli/jollimemory/core/telemetry/Telemetry.kt`, and its tests have no `sandbox` case in `intellij/src/test/kotlin/ai/jolli/jollimemory/core/telemetry/TelemetryTest.kt`. Shipping this as-is would diverge the shared telemetry contract across surfaces.

flyerli2025 (approved):

PR #429, opened 2026-08-03
Fix six drift defects found reviewing #424; raise 51 CLI files to 100% coverage

flyerli2025:
## Summary

Fixes the six defects found while reviewing #424 — each one a place where two lists that had to agree drifted apart — and raises 51 CLI files to 100% statement/branch/function/line coverage.

## Motivation

The six defects share a shape: a second, redundant copy of a decision that nobody updated when the first one changed. A stale `data-context` allowlist made the skills hover card unreachable; a letter-keyed color map collided once two sources stamped the same letter; three push-error classifiers disagreed about HTTP 401. None of them fail loudly — they degrade silently, which is why they survived until a review.

The coverage work is the other half of the same problem. Every gap closed here was a branch no test had ever taken, which is exactly where this class of drift hides.

## Changes

**Defect fixes**

- **Onboarding ledger writes into a disabled repo** ([`OnboardingFunnel.ts`](cli/src/core/OnboardingFunnel.ts)) — the ledger is a new repo-local file, not an append to the buffer the telemetry primitive owns, so the "telemetry recording is ungated" carve-out never covered it. The Disable command's own status refresh was writing into the repo it had just disabled. Gated on the in-memory `isManuallyDisabled()` mirror rather than the durable reader, which can itself persist a migration decision — a write, inside a zero-write gate.
- **Amend stranded its skill archive** ([`QueueWorker.ts`](cli/src/hooks/QueueWorker.ts)) — `consumeWorkspaceContext` had already guarded the registry rows and written the orphan bytes, but the fresh leaf spread only plans/notes/references. The rows came off the panel and nothing addressed the stored files.
- **Skills hover card was unreachable** ([`SidebarScriptBuilder.ts`](vscode/src/views/SidebarScriptBuilder.ts)) — a `data-context` allowlist on the mouseover handler was a second list of hover kinds beside the dispatch below it, and it was the one that went stale. Removed rather than extended: `lookupBranchHoverById` already answers "does this row have hover data". The tooltip was no fallback — the extension contributes only a webview view, so `SerializedTreeItem.tooltip` is inert for every kind.
- **Pinned reference badges painted the wrong color** ([`PinnedPanel.kt`](intellij/src/main/kotlin/ai/jolli/jollimemory/toolwindow/PinnedPanel.kt)) — colors were keyed on the letter tag, which is not a unique key: Jira and Jolli Memory both stamp "J", Zoom Doc and Zoom Meeting both stamp "Z", and "N"/"S" collide with the note/snippet letters the panel does own. Notion pins painted note-green, Slack pins snippet-amber, and the six letters added since the map was written fell through to gray. References now resolve through `SourceDisplay`, so a new source cannot desync this panel again.
- **MCP `status` could not explain a refused push** ([`McpTools.ts`](cli/src/mcp/McpTools.ts)) — added `pushDisabled` / `pushDisabledError`. Both halves or neither: a bare boolean reads as "you turned this repo off" even when the store was simply unreadable, which fails closed machine-wide.
- **401 classified three different ways** ([`PushRefusal.ts`](cli/src/core/PushRefusal.ts), [`JolliPushOrchestrator.kt`](intellij/src/main/kotlin/ai/jolli/jollimemory/services/JolliPushOrchestrator.kt)) — `NotAuthenticatedError` (CLI) / `UnauthorizedError` (IntelliJ) are the same server response. `CreatePrPanel.repoWideStopReason` already stopped the whole loop and reported "sign-in rejected", while the three attachment loops collected N separate `plan "X" failed` lines and fired N doomed requests on the way.

**Coverage — 51 files to 100%**

Session discovery and transcript readers (Antigravity, Cline, Cline CLI, Cursor CLI, Devin), the references DSL (envelope parsers, bindings, normalizers), the skills capture pipeline (scanners, attribution, store, archive), local-agent backends, and push-control. Mostly untaken defensive branches: malformed JSONL, absent optional fields, degraded host shapes.

A handful of gaps turned out to be unreachable rather than untested, and were removed rather than papered over — each with a comment recording the invariant the type system cannot express:

- `sort` comparators with an equal-case arm over a `Map` keyed on the sort field (`SkillStore`, `PushControlStore`).
- A `typeof === "string"` re-check on `JSON.parse` of a regex-captured JSON string literal (`SkillStore`).
- A second `lastRowId !== undefined` test at a return the first one already proved (`OpenCodeSkillScanner`).
- `bodyChars?: number` on a command entry that only exists because its body was measured — now a `CompletedCommandEntry` with the field required (`ClaudeSkillScanner`).

Four genuinely defensive guards were kept and marked `/* v8 ignore */` with the reason.

One test defect fixed along the way: `ClaudeSkillScanner.test.ts`'s "takes the plugin from attributionPlugin" asserted a value that actually came from the id-prefix fallback — `ATTRIBUTED_TURN` matches none of the scanner's pre-filter needles, so it never reaches `captureAttribution`. Split into two tests: one using a nested Skill call (which carries both the attribution fields and a `"name":"Skill"` needle) to exercise the real path, one pinning the prefix fallback.

## Testing

- [x] Added/updated unit tests — 51 files to 100%, verified against a full-suite coverage run, not per-file runs
- [x] CLI suite: 360 files / 9786 tests, **99.21% statements / 97.49% branches / 99.21% functions / 99.41% lines** (floor is 97/96/97/97)
- [x] VS Code suite: 109 files / 4511 tests
- [x] `npm run typecheck` and `npm run lint` clean
- [ ] `npm run all` green end-to-end in a single run — see below

Two runs of `npm run all` each hit a different real-`git` test timing out under full fan-out with coverage (`JolliMemoryBridge.integration.test.ts`, then `GitClient.test.ts` at 341s for the file). Both pass in isolation — 7/7 and 135/135, the latter in 48s — which is the load signal [`AGENTS.md`](AGENTS.md) describes rather than a regression, and neither file is touched by this PR. Worth a look at CI's result before merging.

`intellij/` is unchanged in its build; the two Kotlin edits are covered by `JolliPushOrchestratorTest` and a new `PinnedPanelBadgeColorTest`.

## Checklist

- [x] Code follows the existing style
- [ ] Documentation updated (README / CHANGELOG) if user-facing — the six fixes are all repairs to documented behavior, so no changelog entry
- [x] No new warnings from Plugin Verifier

### DIFF
--- a/cli/src/core/OnboardingFunnel.ts
+++ b/cli/src/core/OnboardingFunnel.ts
@@ -46,11 +46,12 @@
 
 import { mkdir, readFile } from "node:fs/promises";
 import { join } from "node:path";
-import { getJolliMemoryDir } from "../Logger.js";
+import { getJolliMemoryDir, isManuallyDisabled } from "../Logger.js";
 import type { JolliMemoryConfig, StatusInfo } from "../Types.js";
 import { atomicWriteFile } from "./AtomicWrite.js";
 import { isInsideGitRepo } from "./GitOps.js";
 import { resolveLlmCredentialSource } from "./LlmClient.js";
+import { readManualDisableFlagSync } from "./RepoProfile.js";
 import { type BucketLabel, bucket, getTelemetryContext, track } from "./Telemetry.js";
 
 /** Which route (if any) the user has configured to capture memory. */
@@ -210,6 +211,46 @@ export async function maybeEmitOnboardingProgress(opts: ResolveFunnelOptions): P
 		// Short-circuit before any git/status work — and before taking the lock —
 		// when telemetry is off, so an opted-out user pays nothing.
 		if (!getTelemetryContext()?.enabled) return;
+		// Same, for a repo the user durably opted out of (spec 304). The ledger is a
+		// NEW repo-local file, not an append to the buffer the telemetry primitive
+		// already owns, so the "telemetry recording is ungated" carve-out does not
+		// cover it — without this gate the Disable command's own status refresh
+		// writes into the repo it just disabled.
+		//
+		// BOTH readers are consulted because neither alone covers every trigger:
+		//   - `isManuallyDisabled()` is the free in-memory mirror the editor host
+		//     seeds at activate() and flips in lockstep with enable/disable. It is
+		//     the only reader available on VS Code's synchronous write paths, but
+		//     it is process-local — CLI processes never set it (see Logger.ts), so
+		//     on its own it left `jolli status` / bare `jolli` / `jolli enable` /
+		//     the QueueWorker's IngestRunStore recreating the ledger in a repo the
+		//     user had already disabled.
+		//   - `readManualDisableFlagSync` is the durable, disk-backed truth those
+		//     CLI processes need. It must be the SYNC variant: the async
+		//     `readManualDisableFlag` persists a legacy-marker migration decision,
+		//     i.e. a write, which is exactly what this gate exists to prevent.
+		//     Steady-state cost is one `readFileSync`: the `git rev-parse` it needs
+		//     to anchor on the main worktree runs once per `cwd` and is then served
+		//     from `_mainRootCache` (see `RepoProfile.ts`), because this gate is NOT
+		//     a once-per-process seed — VS Code reaches it from every
+		//     `StatusStore.refresh()`, including two file watchers that fire
+		//     repeatedly while an AI session is live, so an unmemoized subprocess
+		//     spawn per refresh would block the extension host's event loop. What
+		//     is left is charged only after the telemetry-consent short-circuit,
+		//     and is small next to the `isInsideGitRepo` / `getStatus` work
+		//     `resolveOnboardingFunnel` runs immediately after it.
+		//
+		// KNOWN TRADE-OFF, deliberate: this sits ahead of `track()`, not merely ahead
+		// of the ledger write, so disabling a repo silences the funnel INCLUDING the
+		// `repo_enabled: false` snapshot that would have recorded the disable itself.
+		// The last snapshot on record for that repo says "enabled" and nothing ever
+		// supersedes it — so the drop-off this funnel exists to observe is the one
+		// event it cannot see. Gating only the persist is NOT the fix: the dedup
+		// ledger would go unread, turning every status refresh into a fresh emit, and
+		// the zero-write contract (spec 304) is the stronger of the two promises.
+		// Recovering the signal needs a one-shot emit at the disable GESTURE itself,
+		// which is a separate design — not a relaxation of this gate.
+		if (isManuallyDisabled() || readManualDisableFlagSync(opts.cwd)) return;
 		ledgerPath = join(getJolliMemoryDir(opts.cwd), LEDGER_FILE);
 		const path = ledgerPath;
 		run = (ledgerLocks.get(path) ?? Promise.resolve()).then(() => emitOnce(opts, path));
--- a/cli/src/core/PushRefusal.ts
+++ b/cli/src/core/PushRefusal.ts
@@ -42,11 +42,39 @@
  * Error `name`s that mean "this repo cannot push right now", for any credential
  * and for every document in it.
  *
- * Both spellings of the outdated-client condition are listed on purpose: the CLI
- * raises `ClientOutdatedError` while the VS Code and IntelliJ clients raise
- * `PluginOutdatedError` for the identical server response (HTTP 426). Since this
- * set is shared and errors also cross the IDE bridge by name, matching only one
- * spelling would silently classify the other as a per-item failure.
+ * Two conditions are listed under BOTH of their spellings on purpose, because the
+ * surfaces name the same server response differently and errors cross the IDE
+ * bridge by name — matching one spelling would silently classify the other as a
+ * per-item failure. HTTP 426: `ClientOutdatedError` (CLI) / `PluginOutdatedError`
+ * (VS Code + IntelliJ). HTTP 401: `NotAuthenticatedError` (CLI) /
+ * `UnauthorizedError` (IntelliJ).
+ *
+ * Note that 403 maps to whichever of the two auth names the endpoint can justify,
+ * and both are listed here because both are repo-wide: the read-shaped endpoints
+ * (`frontDoor`, `deleteDoc`) cannot tell a rejected credential from a forbidden
+ * repo and fold 403 into `NotAuthenticatedError`, while the push/bind endpoints
+ * distinguish it and raise `PermissionDeniedError`. Don't go looking for a single
+ * 403 branch — there isn't one (see `JolliMemoryPushClient`).
+ *
+ * A rejected credential is repo-wide for the same reason the others are: it is a
+ * property of the repo + credential, so every remaining document in a loop gets
+ * the identical rejection. It was the last classifier disagreement — IntelliJ's
+ * `CreatePrPanel.repoWideStopReason` already stopped the whole loop on it and
+ * reported "sign-in rejected", while `JolliPushOrchestrator.isFatalPushError` and
+ * the CLI's three attachment loops collected it as N separate `plan "X" failed`
+ * lines and fired N doomed requests on the way.
+ *
+ * **That 401 promotion reaches the CLI and IntelliJ, NOT VS Code** — membership
+ * here is necessary but not sufficient, because a surface also has to PRODUCE one
+ * of these names. `vscode/src/services/JolliPushService.ts` branches on 426, 412,
+ * 409 and 403 only, so a 401 falls through to its generic non-2xx arm and rejects
+ * with a plain `Error`, whose `name` is `"Error"` — a string this set cannot
+ * match. Its attachment loops therefore still collect a rejected credential as N
+ * per-item failures, exactly the shape the promotion removes elsewhere. A
+ * `status === 401` branch in that client, raising `UnauthorizedError`, is all it
+ * takes for VS Code to inherit the behaviour; until it exists, read "shared
+ * classifier" as shared membership, not as identical behaviour on this one
+ * condition.
  *
  * `BindingRequiredError` is deliberately ABSENT: it is recoverable — the caller
  * runs the binding chooser and retries — so it is fatal only to the loop that
@@ -56,6 +84,8 @@
 export const REPO_WIDE_REFUSAL_NAMES: ReadonlySet<string> = new Set([
 	"ClientOutdatedError",
 	"PluginOutdatedError",
+	"NotAuthenticatedError",
+	"UnauthorizedError",
 	"PushDisabledError",
 	"PermissionDeniedError",
 ]);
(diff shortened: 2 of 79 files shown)


cli/src/core/PushRefusal.ts
```diff
@@ -56,6 +64,8 @@
 export const REPO_WIDE_REFUSAL_NAMES: ReadonlySet<string> = new Set([
 	"ClientOutdatedError",
 	"PluginOutdatedError",
+	"NotAuthenticatedError",
```
dougschroeder, line 87: **Should fix — comment accuracy.** This promotion reaches the CLI and IntelliJ, but not VS Code, and the surrounding comments now read as though all three agree.

`vscode/src/services/JolliPushService.ts` branches on 426, 412, 409 and 403 only. A 401 falls through to the generic `status < 200 || status >= 300` arm at `:333` and is rejected as a plain `Error`, whose `name` is `"Error"`. `isRepoWideRefusal` matches on `name`, so it never fires there — the VS Code attachment loops still collect N per-item failures on a rejected credential, which is the exact behaviour this change sets out to remove.

**Why it matters:** the defect itself is pre-existing and that file is not in this diff, so leaving it is fine. What is new is the claim. This header, and `JolliPushOrchestrator.kt:50` ("the shared source of truth the CLI and VS Code loops both read"), both read as three-surface parity. That is true of the classifier and false of the classification — and a comment asserting a fix that isn't there is the same stale-second-list shape this PR is about.

**Suggested fix:** narrow both comments to name the two surfaces the promotion actually reaches, and note that VS Code's client needs its own 401 branch before it inherits the behaviour. Add that branch in a follow-up.

cli/src/core/OnboardingFunnel.ts
```diff
@@ -210,6 +211,29 @@ export async function maybeEmitOnboardingProgress(opts: ResolveFunnelOptions): P
...
+		//     user had already disabled.
+		//   - `readManualDisableFlagSync` is the durable, disk-backed truth those
+		//     CLI processes need. It must be the SYNC variant: the async
+		//     `readManualDisableFlag` persists a legacy-marker migration decision,
+		//     i.e. a write, which is exactly what this gate exists to prevent.
+		//     Its cost (one `git rev-parse` + one `readFileSync`) is charged only
+		//     after the telemetry-consent short-circuit, and is small next to the
+		//     `isInsideGitRepo` / `getStatus` work `resolveOnboardingFunnel` runs
+		//     immediately after it.
+		if (isManuallyDisabled() || readManualDisableFlagSync(opts.cwd)) return;
```
dougschroeder, line 253: **Should fix — performance.** This adds a blocking subprocess to a hot path on the VS Code extension host.

In an enabled repo (the common case) `isManuallyDisabled()` is `false`, so the gate falls through to `readManualDisableFlagSync`, which runs `execFileSync("git", ["rev-parse", "--git-common-dir"])`. `StatusStore.refresh()` calls this emitter, and refresh is wired to the sessions watcher, the HEAD watcher and roughly eighteen command paths. Telemetry is on by default, so the consent short-circuit above does not spare it. Before this PR the sync reader ran once per session, at `Extension.ts:464`.

**Why it matters:** each call blocks the extension host event loop for a process spawn — tens of milliseconds on macOS and worse on Windows — which stalls every other extension in the window, not only this one.

**Suggested fix:** the durable read is redundant on this surface, since inside the editor host the in-memory mirror is authoritative by construction. Either memoise the sync read per `cwd`, or skip it once the mirror has been seeded. Either keeps the CLI coverage this gate needs without paying on every refresh.

dougschroeder, line 253: **Should fix — a decision worth confirming.** This gate sits ahead of `track()` as well as the ledger write, so disabling a repo permanently silences the funnel, including the `repo_enabled: false` snapshot that records the disable itself. The last snapshot on record for that repo says it is enabled, and nothing ever supersedes it.

**Why it matters:** the funnel exists to observe drop-off, and disabling is the drop-off — so it becomes invisible in the data. Gating the whole emit is the right call for the zero-write contract (gating only the persist would break the dedup ledger and turn every status refresh into a fresh emit), so this is a trade rather than a mistake. But it is a real one, and it is not mentioned anywhere in the PR.

**Suggested fix:** confirm it is intended and say so in the description. If the drop-off signal is worth keeping, it needs a design that emits once at the disable gesture without recreating the ledger.

dougschroeder (dismissed):
## Review — six drift fixes, plus 51 files to 100%

Six well-chosen fixes, each aimed at a real duplicated-decision bug, and the coverage work is unusually disciplined: dead branches are removed with the invariant written down, and the four genuinely-defensive ones are kept behind an ignore directive rather than deleted to make a number go up. The amend, hover-card and pinned-badge fixes are each correct and consistent with how the sibling paths already behave.

Approving — nothing here blocks. Three follow-up items are left inline, and all three are either a comment correction or a decision to confirm rather than a defect this PR introduces.

### What's good

- **`bodyChars?: number` becoming a `CompletedCommandEntry` with the field required.** The best change in the PR: it converts "this branch is unreachable" from a comment into something the compiler enforces.
- **The amend fresh-leaf fix is consistent with its three siblings.** The amend short-circuit, the full amend path and squash all already set `skills: consumed.skillRefs`; the fresh leaf was the only one that didn't, and it is also the one path with no re-association backstop — so nothing else could have attached those refs later. Conditional emission (`length > 0`) is preserved.
- **The sidebar fix removes the second list instead of extending it**, and the `mouseout` symmetry note shows the half-fix was considered — a kind that can open a card but not close it. The reasoning about the native tooltip holds for these rows in particular, since they deliberately clear `title=` so it cannot compete with the card.
- **`PinnedPanel` recovers the source from the pin's own key**, so no change to the stored shape was needed, and the new test pins each collision class separately: Notion/note, Slack/snippet, Jira/Jolli, hyphenated wire names, and the six letters that fell through to grey. `zoom-doc` in particular would defeat a naive `valueOf`.
- **The test defect is real and honestly fixed.** `ATTRIBUTED_TURN` matches none of the scanner's pre-filter needles, so that assertion was passing via the id-prefix fallback. Splitting it in two is the kind of thing a coverage PR usually papers over.
- **The two `npm run all` timeouts are reported rather than hidden**, and correctly read as the known real-`git` contention signal. Neither file is touched here.

### Worth exercising by hand before merge

1. **Skills survive an amend** — use a skill, commit, then `git commit --amend`, and open the commit's memory. The skill should still be listed; previously the row left the panel while the archived file stayed on the orphan branch, unreferenced.
2. **Skills hover card** — hover the "Skills used" row for about a second, move onto the card, then away. Confirm the plan / note / reference cards are unchanged too, since the allowlist removal touches all of them.
3. **A disabled repo writes nothing** — in a repo with no ledger file, run `jolli disable`, then `jolli status` and bare `jolli`. No `onboarding-progress.json` should appear. Re-enable and confirm it comes back.
4. **401 mid-push** — invalidate the credential and push from the CLI, from IntelliJ, and from VS Code. The first two should stop once; VS Code will still emit one failure per attachment (see the note on `PushRefusal.ts`).
5. **Pinned badges** — pin a Notion reference, a Slack reference, a Jira reference, a Jolli Memory reference, a note and a snippet, then compare each badge against its own source row. Check light and dark themes; the new source table carries identical values for both, where the old map had distinct dark hues for four entries.
6. **Non-git dead end** — run `jolli` outside a repo. Exactly one "not a git repository" message, with no `fatal:` line leaking out from underneath it.

cli/src/core/PushRefusal.ts
flyerli2025: Narrowed both comments as suggested; the VS Code 401 branch itself is left as the follow-up you described.

**`PushRefusal.ts`** — the header now carries a dedicated paragraph at `:67` stating the promotion reaches the CLI and IntelliJ and *not* VS Code, plus why: membership in this set is necessary but not sufficient, because a surface also has to PRODUCE one of these names. It names `JolliPushService`'s 426/412/409/403-only branching, that a 401 therefore rejects with a plain `Error` whose `name` this set cannot match, and that its attachment loops consequently still collect N per-item failures. It closes by telling the reader to read "shared classifier" as shared membership, not identical behaviour on this one condition, and records the one-line fix that would change that: a `status === 401` branch raising `UnauthorizedError`.

**`JolliPushOrchestrator.kt`** — kept the "shared source of truth" sentence (it is accurate about the set) but qualified it immediately after, at `:54`: sharing the set is not the same as sharing the behaviour; VS Code reads it but has no 401 branch, so a rejected credential still reaches its loops as a plain `Error` and is collected per attachment. The 401 stop is stated to be live on IntelliJ and the CLI only.

Your framing was the useful part here — a comment asserting a fix that isn't there is the same stale-second-list shape this PR is about, so the claim is now scoped to what the code actually does.

cli/src/core/OnboardingFunnel.ts
flyerli2025: Took your first option — memoised per `cwd`, so the gate keeps the CLI coverage it needs without a spawn per refresh.

`readManualDisableFlagSync` (`RepoProfile.ts:310`) now resolves the main worktree root through `resolveMainRootSync` (`:266`), backed by `_mainRootCache` (`:258`). The memo is written on the failure path too (`:284-288`): a non-git `cwd` falls back to `cwd` itself and is cached, which matters because that is the case paying the full subprocess cost — the guided front door's non-git dead end reaches this gate. So `git rev-parse --git-common-dir` runs at most once per `cwd` per process.

One deliberate narrowing of your wording: what is memoised is the **root resolution**, not the read. `profile.json`'s `readFileSync` stays uncached, because the common dir is fixed for the life of a process while `manuallyDisabled` is exactly the value that flips under enable/disable — caching the read would stale the answer this gate exists to consult. Steady-state cost per `StatusStore.refresh()` is therefore one `readFileSync` (plus a `statSync` only on the legacy-marker fallback), no process spawn. Rationale is written down at `RepoProfile.ts:252-256`.

While in there: the `execFileSync` also captures git's stderr (`:274-279`) so a non-git `cwd` stops leaking `fatal: not a git repository` under jolli's own message.

flyerli2025: Confirmed: intended, and your reading of the trade is the one we took.

It is now written down in code, at the gate itself (`OnboardingFunnel.ts:243`, `KNOWN TRADE-OFF, deliberate`). It records all three parts: that the gate sits ahead of `track()` and so silences the funnel including the `repo_enabled: false` snapshot that would have recorded the disable, leaving "enabled" as the last snapshot on record with nothing superseding it; that gating only the persist is **not** the fix, because the dedup ledger would go unread and every status refresh would become a fresh emit, and the zero-write contract (spec 304) is the stronger of the two promises; and that recovering the signal needs a one-shot emit at the disable **gesture**, which is a separate design rather than a relaxation of this gate.

Two things still outstanding on the description side, both fair hits:

1. It says nothing about this trade-off, which is the half of your request the code comment does not satisfy.
2. Its `OnboardingFunnel` bullet is now stale in a way this thread's sibling caused — it reads "gated on the in-memory `isManuallyDisabled()` mirror **rather than** the durable reader", but the code consults both (`isManuallyDisabled() || readManualDisableFlagSync(opts.cwd)`) after the memoisation fix. The description was not updated alongside it.

Both belong in the description, not just the code, so it is not being carried only in a comment.

foster-han (approved):

PR #443, opened 2026-08-06
Collapse duplicate KB folders on sidebar Refresh (jolliai + jolliai-2 symptom)

summerfang-jolli:
## Summary

Refresh now runs a single sweep — `KbFoldersService.archiveUnusedFolders()` — that archives Memory Bank folders holding **nothing**: no `index.json` summaries, no `manifest.json` rows, no visible files or branch dirs, and nothing under `.jolli/` beyond the inert stubs `MetadataManager.ensure()` seeds. One click both re-lists the Folders tab and clears the rows the user can't act on. A toast reports how many were archived and offers a **Reveal Archive** action.

## Why

`resolveKBPath` **claims** the folder it returns (writes `.jolli/config.json`) using whatever cwd the calling process had. Every jollimemory process that ran outside a real project therefore materialized a folder named after that cwd — `system32`, `unknown`, `tmp.XXXXXX`, an agent's temp dir, a doc title a local-agent run used as its working directory — plus checkouts opened once and never committed to. `checkClaimable` now gates new ones, but it can't retract the pile already on disk, and users have had to clear it by hand. Measured on a real Memory Bank: **25 of 57 rows**, every one with zero memories.

## Emptiness is the whole criterion (deliberately not a dedupe pass)

An earlier revision (`808c5c22`) wired a duplicate-collapsing pass into Refresh; `b60ac2c9` **withdrew it entirely**. Archiving is a plain `renameSync`, and a move cannot merge — collapsing two populated folders means burying one. The only correct survivor is one rebuilt from the orphan branch (the system of record), which is **Migrate's** capability, not Refresh's. So Refresh stays a lossless action with one stated invariant: it never moves a folder holding memories.

Keying on emptiness alone is what makes that safe:

- **It cannot lose memories.** An empty folder has nothing to move, and `resolveKBPath` re-claims the path on the next write.
- **It needs no name heuristics.** Guessing junk from a name would eventually archive a legitimately-named repo; a populated `system32` folder means the user really does keep memories for a project by that name.
- **"Keep unless proven empty" at every exit** — unreadable dir, unparseable manifest, unrecognized `.jolli/` entry (including the `shadow-status.json` dirty marker) all resolve to *keep*.

Never touched: the current project's folder, or any folder carrying its `repoName` (a fresh install's own folder is legitimately empty until the first commit). Folders without `.jolli/config.json` are invisible to `discoverRepos` and out of scope by construction. Extras are **moved** to the hidden, recoverable `<kbParent>/.jolli/archive/` (same path as Migrate), never hard-deleted.

## Follow-up (review fixes, `c5cdcf60`)

Addresses the review on `b60ac2c9`: corrected the stale `dedupeFolders` comments and the misleading `async` signature on `handleRefresh`; extended the OS-noise skip into the `.jolli/` loop and made it case-insensitive (lowercase Windows `thumbs.db`); fixed the `shadow-status.json` docstring (it records a *failed* shadow write, written from `DualWriteStorage`'s `catch`, not one in flight); and added the **Reveal Archive** toast action.

## Scope note

This sweep is VS Code-only. The IntelliJ Memory Bank sidebar gets no equivalent yet, so the junk rows persist there — a deliberate follow-up, not part of this PR.

## Verification

- `node scripts/run-vitest.mjs` on the two changed test files → **2 files / 340 tests passed**.
- Full `npm run test:vscode` (with coverage) → **112 files / 4620 passed / 16 skipped**, coverage above the 97/97/97/97 floor.
- `npm run lint:vscode` (biome `--error-on-warnings`) → clean, 229 files.
- `npm run typecheck:vscode` → exit 0.

### DIFF
--- a/cli/src/core/FolderConsolidation.test.ts
+++ b/cli/src/core/FolderConsolidation.test.ts
@@ -1,0 +1,419 @@
+import { existsSync, mkdirSync, readdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { FileWrite, SummaryIndex } from "../Types.js";
+import {
+	__setSotResolverForTests,
+	type ConsolidationPlan,
+	classifyDuplicateFolders,
+	executeConsolidation,
+} from "./FolderConsolidation.js";
+import { __setSshRunnerForTests } from "./SshAliasResolver.js";
+import type { StorageProvider } from "./StorageProvider.js";
+
+// Keep visible-markdown generation trivial so the orphan-superset rebuild does
+// not depend on the full CommitSummary rendering surface.
+vi.mock("./SummaryMarkdownBuilder.js", () => ({
+	buildMarkdown: vi.fn().mockReturnValue("# Mock Markdown\n\nBody"),
+}));
+
+vi.spyOn(console, "log").mockImplementation(() => {});
+vi.spyOn(console, "warn").mockImplementation(() => {});
+vi.spyOn(console, "error").mockImplementation(() => {});
+
+/** In-memory StorageProvider standing in for the orphan branch. */
+class InMemoryStorage implements StorageProvider {
+	private files = new Map<string, string>();
+	private present: boolean;
+	constructor(present = true) {
+		this.present = present;
+	}
+	async readFile(path: string): Promise<string | null> {
+		return this.files.get(path) ?? null;
+	}
+	async writeFiles(files: FileWrite[], _msg: string): Promise<void> {
+		for (const f of files) {
+			if (f.delete) this.files.delete(f.path);
+			else this.files.set(f.path, f.content);
+		}
+	}
+	async listFiles(prefix: string): Promise<string[]> {
+		return [...this.files.keys()].filter((k) => k.startsWith(prefix)).sort();
+	}
+	async exists(): Promise<boolean> {
+		return this.present;
+	}
+	async ensure(): Promise<void> {}
+}
+
+function makeTmpParent(): string {
+	const dir = join(tmpdir(), `kb-consol-${Date.now()}-${Math.random().toString(36).slice(2)}`);
+	mkdirSync(dir, { recursive: true });
+	return dir;
+}
+
+function summaryJson(hash: string): string {
+	return JSON.stringify({
+		version: 3,
+		commitHash: hash,
+		commitMessage: "Test commit",
+		commitAuthor: "Alice",
+		commitDate: "2026-01-15T10:00:00Z",
+		branch: "main",
+		generatedAt: "2026-01-15T10:00:00Z",
+		topics: [{ title: "T", trigger: "t", response: "r", decisions: "d" }],
+		stats: { filesChanged: 1, insertions: 1, deletions: 0 },
+	});
+}
+
+function indexJson(hashes: string[]): string {
+	const index: SummaryIndex = {
+		version: 3,
+		entries: hashes.map((h) => ({
+			commitHash: h,
+			parentCommitHash: null,
+			commitMessage: "Test",
+			commitDate: "2026-01-15T10:00:00Z",
+			branch: "main",
+			generatedAt: "2026-01-15T10:00:00Z",
+		})),
+	};
+	return JSON.stringify(index);
+}
+
+/** Creates `<parent>/<name>/.jolli/{config,index,summaries}` for a repo. */
+function makeKbFolder(
+	parent: string,
+	name: string,
+	opts: { remoteUrl: string; repoName: string; hashes: string[] },
+): string {
+	const root = join(parent, name);
+	const jolli = join(root, ".jolli");
+	mkdirSync(join(jolli, "summaries"), { recursive: true });
+	writeFileSync(
+		join(jolli, "config.json"),
+		JSON.stringify({ version: 1, sortOrder: "date", remoteUrl: opts.remoteUrl, repoName: opts.repoName }),
+	);
+	for (const h of opts.hashes) writeFileSync(join(jolli, "summaries", `${h}.json`), summaryJson(h));
+	writeFileSync(join(jolli, "index.json"), indexJson(opts.hashes));
+	return root;
+}
+
+function summaryHashesOnDisk(root: string): string[] {
+	try {
+		return readdirSync(join(root, ".jolli", "summaries"))
+			.filter((n) => n.endsWith(".json"))
+			.map((n) => n.slice(0, -5))
+			.sort();
+	} catch {
+		return [];
+	}
+}
+
+const REMOTE = "git@github.com:acme/app.git";
+
+function assertPlan(plan: ConsolidationPlan | null): asserts plan is ConsolidationPlan {
+	expect(plan).not.toBeNull();
+	if (!plan) throw new Error("expected a consolidation plan");
+}
+
+describe("FolderConsolidation", () => {
+	let parent: string;
+
+	beforeEach(() => {
+		parent = makeTmpParent();
+		// Default: orphan branch empty (present but no summaries) → union-largest path.
+		__setSotResolverForTests(async () => new InMemoryStorage(true));
+	});
+
+	afterEach(() => {
+		__setSotResolverForTests(null);
+		__setSshRunnerForTests(null);
+		try {
+			rmSync(parent, { recursive: true, force: true });
+		} catch {
+			/* ignore */
+		}
+	});
+
+	describe("classifyDuplicateFolders", () => {
+		it("returns null when fewer than two folders exist for the repo", async () => {
+			makeKbFolder(parent, "app", { remoteUrl: REMOTE, repoName: "app", hashes: ["a1", "a2"] });
+			expect(await classifyDuplicateFolders("/cwd", "app", REMOTE, parent)).toBeNull();
+		});
+
+		it("classifies identical folders and keeps the shortest name as survivor", async () => {
+			makeKbFolder(parent, "app", { remoteUrl: REMOTE, repoName: "app", hashes: ["a1", "a2"] });
+			makeKbFolder(parent, "app-2", { remoteUrl: REMOTE, repoName: "app", hashes: ["a1", "a2"] });
+			const plan = await classifyDuplicateFolders("/cwd", "app", REMOTE, parent);
+			expect(plan?.kind).toBe("identical");
+			expect(plan?.survivor).toBe(join(parent, "app"));
+			expect(plan?.archived).toEqual([join(parent, "app-2")]);
+			expect(plan?.counts.added).toBe(0);
+		});
+
+		it("classifies union-largest when a folder holds summaries beyond the orphan branch", async () => {
+			makeKbFolder(parent, "app", { remoteUrl: REMOTE, repoName: "app", hashes: ["a1", "a2", "a3"] });
+			makeKbFolder(parent, "app-2", { remoteUrl: REMOTE, repoName: "app", hashes: ["a1", "z9"] });
+			const plan = await classifyDuplicateFolders("/cwd", "app", REMOTE, parent);
+			expect(plan?.kind).toBe("union-largest");
+			expect(plan?.survivor).toBe(join(parent, "app")); // 3 > 2
+			expect(plan?.counts.union).toBe(4);
+			expect(plan?.counts.survivor).toBe(3);
+			expect(plan?.counts.added).toBe(1);
+		});
+
+		it("classifies orphan-superset when the orphan branch already covers every folder", async () => {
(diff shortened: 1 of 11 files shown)


flyerli2025:
## Review

Reviewed at `b60ac2c9`. Gates I actually ran locally (not inferred):

- `node scripts/run-vitest.mjs` on both changed test files → **2 files / 335 tests passed**, exit 0
- `npm run lint` (vscode, biome `--error-on-warnings`) → clean, 229 files
- `npm run typecheck` (vscode) and `npm run typecheck:cli` → both exit 0
- Did **not** run the full `npm run all` (build + full CLI suite)

**Verdict: good to merge once the PR description is rewritten.** One High (process), three Medium, four Low. Nothing blocking in the production code.

The most valuable part of this PR is the half that undoes itself. `808c5c22` collapsed duplicate folders by repo identity; `b60ac2c9` withdraws that entirely and keeps only "archive folders that hold nothing." The reason is a real invariant: archiving is a `renameSync`, and **a move cannot merge** — collapsing two populated folders means burying one. The only correct survivor is one rebuilt from the orphan branch, which is Migrate's capability, not Refresh's. That call is right, and it's recorded in the two places that will actually be read later (the service docstring, and the "I used to be exported, don't re-export me" note on `normalizeRemoteUrl`).

---

### High

**1. The PR description describes the withdrawn design, and carries a forbidden footer**

The body is entirely about `dedupeFolders()` and its grouping rules (base + `-N`, equal remotes, distinct remotes, lone local-only). All of that was deleted in `b60ac2c9`. What ships is `archiveUnusedFolders()`, keyed on emptiness alone, with duplicate folders explicitly out of scope. Anyone reading this PR later gets the opposite of what merged.

Separately, the body ends with a `🤖 Generated with [Claude Code]` footer. `AGENTS.md` states this as a critical rule covering commit messages **and PR descriptions**. The commits themselves are clean (both carry only `Signed-off-by:`) — it's only the PR body.

Also stale: "Full-project `typecheck` needs CLI package deps installed in the worktree." Both `typecheck` and `typecheck:cli` pass clean on this branch now.

Suggestion: replace the body with the `b60ac2c9` commit message — it is a considerably better description than the current one.

### Medium

**2. `SidebarWebviewProvider.ts:1173` — the comment is wrong in two ways**

```
// Now async (the KB scope awaits `dedupeFolders` before
// re-listing). It swallows its own failures, so the caller
// stays fire-and-forget like every other message handler.
```

`dedupeFolders` no longer exists, and the KB scope specifically does **not** await — `handleRefresh` calls `void this.tidyAndRelistKbFolders()`. The commit message explains exactly why awaiting is wrong (it pushes the branch/status refreshes past a microtask and breaks the same-tick observability of the Codex discovery kick that a `scope="all"` refresh owes `pushConversations` — caught by an existing test).

This is the one comment in the PR that would actively lead the next person to re-add an `await` and walk back into a trap a test already caught. Worth fixing on its own.

The method's JSDoc just below (L1822) also still says "we call `handleExpandFolder("")` directly", which is now `tidyAndRelistKbFolders`. And `KbFoldersService.test.ts:1796` has one more leftover `dedupeFolders` reference.

**3. `handleRefresh` is now `async` with no `await` in its body**

Every call inside it is `void`-ed, so the returned promise is always already-resolved, and the call site voids it anyway. The `async` keyword changes no behaviour — it only implies an ordering guarantee that does not exist, which is precisely the misreading in finding 2. Either drop it back to `: void`, or make the comment say the signature is incidental.

**4. Design: Refresh is now a button that mutates the disk**

The service docstring says it outright — "This is the ONLY mutation the sidebar's Refresh performs." Refresh reads as read-only in every IDE, and this version renames directories with no preview and no confirmation. The toast says "(recoverable)" but offers no action to get there, and the archive lives in a hidden `<kbParent>/.jolli/archive/` the user will not find by following that sentence.

The rationale for folding it into Refresh (one click both re-lists and clears rows the user can't act on) is documented and defensible, so this is a second opinion rather than an objection. Two cheap improvements: add a `Reveal Archive` action button to the toast, or expose the sweep as its own "Clean up unused Memory Bank folders" command and let Refresh stay read-only. Your call.

Related: the IntelliJ Memory Bank sidebar gets no equivalent sweep, so the 25-of-57 junk rows measured in the commit message stay there. Worth one line in the description saying whether that's a deliberate follow-up.

### Low

**5. OS-noise handling is asymmetric between the two loops**

`OS_NOISE_FILES` is only consulted in the top-level loop; the `.jolli/` loop uses the strict allowlist. So a `.jolli/.DS_Store` counts as content and pins the junk folder permanently — exactly the failure the top-level skip exists to prevent, and one that a single Finder visit with hidden files shown will produce on macOS. One-line fix: apply the same skip inside the `.jolli` loop.

**6. `OS_NOISE_FILES` is case-sensitive**

`thumbs.db` (lowercase) is common on Windows. Compare lowercased.

**7. `shadow-status.json`'s protection is overstated in the docstring**

The docstring reads: "means a folder write was attempted and hasn't landed… Empty-looking, but mid-write — not disposable." But `DualWriteStorage.writeFiles` calls `markDirty` only in the `catch`, after a shadow write has **failed** — it is not a pre-write marker. So it covers "already failed", not "currently in flight."

Narrow but real consequence: another window landing the very first memory for repo Y (folder still empty at that instant) while you click Refresh here can have that folder renamed into the archive mid-write; on POSIX the writer's data follows the inode into the archive and becomes invisible. I don't think this warrants locking, but the docstring shouldn't claim coverage it doesn't have.

**8. Sync IO inside an async method**

`isEmptyKbFolder` uses `readdirSync`/`readFileSync` per discovered repo, while the rest of the service is on `fs.promises`. At 57 rows this is irrelevant in practice — consistency nit only.

---

### What's well done

- The withdrawal is the strongest part of the change, and the reasoning is parked where it will be read: the service docstring plus the `normalizeRemoteUrl` note. That note is the most effective available guard against the same idea being re-implemented in six months.
- The "keep unless proven empty" polarity is correct at every exit — unreadable dir, unparseable manifest, unrecognized name all resolve to keep. The `-1` sentinel in `countManifestFiles` is the right call over `0`.
- The two boundary tests (duplicates holding 35 and 751 summaries both survive; an empty duplicate sibling is still archived) pin "the test is emptiness, not uniqueness" as an executable invariant. That's the guard that stops a dedupe pass being re-added here.
- The `chmod 0o111` tests are a genuinely clever way to reach both `readdir` catch paths without a filesystem race — and execute-without-read is the real shape of a permission-restricted folder, which is the case that must be kept.

summerfang-jolli:
Thanks for the thorough pass — every finding validated against the code. All fixed in `0a1b05b0`, plus the PR description rewrite.

**1 (High) — PR description + forbidden footer + stale typecheck note.** Fixed. Rewrote the body around what actually ships (`archiveUnusedFolders`, keyed on emptiness), dropped the `🤖 Generated with…` footer, and removed the stale "CLI deps not installed" line. Both `typecheck:vscode` and the suite run clean here.

**2 (Medium) — wrong comments at `SidebarWebviewProvider.ts:1173` / JSDoc L1822 / test L1796.** Fixed all three. The `refresh` case comment no longer claims the KB scope awaits `dedupeFolders`; the `handleRefresh` JSDoc now says the KB scope runs `tidyAndRelistKbFolders()` directly; and the leftover `dedupeFolders` reference in `KbFoldersService.test.ts` is now `archiveUnusedFolders`. You're right that this was the one comment that would have walked the next person back into re-adding the `await`.

**3 (Medium) — `handleRefresh` async with no await.** Fixed by dropping it to `: void` (the honest signature) rather than keeping `async` and explaining it away. Body is unchanged — every call stays `void`-ed, and the inline note on *why* the sweep isn't awaited stays put.

**4 (Medium) — Refresh mutates disk.** Kept Refresh's fold-in behavior (agreed it's documented and defensible), and took your first cheap improvement: the toast now offers a **Reveal Archive** action (`KbFoldersService.archiveDir()` → `revealFileInOS`), so "(recoverable)" leads somewhere instead of being a dead-end claim. On IntelliJ parity — added an explicit scope note to the PR description saying the sidebar sweep is VS Code-only for now and the JVM sidebar's junk rows are a deliberate follow-up.

**5 (Low) — OS-noise asymmetry.** Fixed. The `.jolli/` loop now applies the same OS-noise skip as the top-level loop, so a `.jolli/.DS_Store` no longer pins the folder. New test covers it.

**6 (Low) — case-sensitive `OS_NOISE_FILES`.** Fixed. Names are stored lowercase and matched via `isOsNoise(name)` (`name.toLowerCase()`), so a lowercase Windows `thumbs.db` counts as noise. New test pins both the lowercase `thumbs.db` and an uppercase `.DS_STORE`.

**7 (Low) — `shadow-status.json` docstring overstated.** Fixed. Both docstrings (and the matching test comment) now say the marker records a shadow write that **failed** — `DualWriteStorage` writes it only from the write's `catch` — rather than one "in flight". The folder is still kept either way; we just stopped claiming coverage the marker doesn't provide.

**8 (Low) — sync IO in an async method.** Left as-is, per your own read that it's a consistency nit that's irrelevant at this scale. `isEmptyKbFolder` is a standalone sync predicate; converting it to `fs.promises` for 57 rows would add churn for no measurable gain. Happy to revisit if the folder count ever grows.

Verification after the fixes: `test:vscode` → 112 files / 4620 passed / 16 skipped, coverage above the 97/97/97/97 floor; `lint:vscode` and `typecheck:vscode` clean.

summerfang-jolli:
@flyerli2025 all review findings are addressed in `0a1b05b0` (pushed), and the PR description has been rewritten. Per-finding accounting is in the [reply above](https://github.com/jolliai/jolliai/pull/443#issuecomment-5236195352). Findings 1–7 fixed; finding 8 left as-is per your own read that it's a non-issue at this scale. Verification: `test:vscode` 112 files / 4620 passed / 16 skipped (coverage above the 97/97/97/97 floor), `lint:vscode` and `typecheck:vscode` clean. Ready for another look.

summerfang-jolli:
Rebased onto `origin/main` (`f48c1c07`) and force-pushed — clean replay, no conflicts. Branch tip is now `9f7bd5b0`. Re-verified after the rebase: `typecheck:vscode` and `lint:vscode` clean, full `test:vscode` 110 files / 4475 passed / 16 skipped with coverage above the 97/97/97/97 floor.

flyerli2025:
## Code review

The emptiness criterion itself is sound. I traced `MetadataManager.ensure()`, `FolderStorage` and `resolveKBPath` and confirmed a claimed-but-unused folder holds exactly the three allow-listed stubs; the `-1` sentinel on a corrupt manifest correctly resolves to "keep"; the two current-repo guards are right; and archived base slots do not shadow real data (`scanRepoSuffixes` still finds `<repo>-N`).

The three issues below are all about the sweep's interaction with the **vault** it mutates: `<kbParent>` is both the vault git root and the FolderStorage root, so moving a directory there is simultaneously an unlocked working-tree mutation and an implicit `git rm`.

### High — the sweep mutates the vault working tree without `vault-write.lock`

`vscode/src/services/KbFoldersService.ts:212`

The sweep `renameSync`s directories inside the vault git working tree, but QueueWorker, SyncEngine and both compile `writeGuard`s acquire `vault-write.lock` first. Two concrete interleavings:

1. A sync round that has already snapshotted `statusPorcelainZ()` then calls `git add -f -- <repo>/.jolli/config.json` on a path the sweep just moved. `stageAddPaths`' own docstring says ENOENT propagates, so the round fails loudly.
2. A QueueWorker claims `<kbParent>/B` at drain start and leaves it empty for the whole LLM window. The sweep archives it, and the worker's `ensure()` recreates it as an identity-less stub — which is exactly the phantom `<repo>-N` folder this PR is trying to eliminate.

Suggested fix: take the same vault write lock the other writers use around the rename, rather than relying on the sweep being short.

### Medium — "an empty folder has nothing to move" is not true for synced users

`vscode/src/services/KbFoldersService.ts:192`

`.jolli/{config,manifest,index,branches}.json` are in `ALLOWED_AGGREGATE_FILES` and are git-tracked, and `stageVault` runs vault-wide. So the rename becomes a staged `git rm` that gets committed and pushed, stripping that repo's cross-device identity config on peers.

### Low — archived repos stay in the breadcrumb dropdown until reload

`vscode/src/views/SidebarWebviewProvider.ts:1962`

`tidyAndRelistKbFolders` re-lists the tree but never re-pushes `selection:repos` / `selection:branches`, unlike `refreshKnowledgeBaseFolders`. The dropdown keeps offering archived repos; picking one hits `if (!target) return` and silently does nothing until the window is reloaded.

### Checked and cleared

- The client `folderCache` correctly drops archived rows via `mergeFolders`' root replacement.
- `archiveDir()` matches `archiveKBFolder`'s destination — `kbParent` is already `resolveKbParent`-resolved in `Extension.ts`.
- The `void`-not-`await` in `handleRefresh` preserves same-tick ordering for the other scopes.
- The absent-dependency path still re-lists synchronously.
- The manually-disabled case is unreachable: the tab toolbar (and its Refresh button) is hidden when `state.enabled` is false.

flyerli2025 (approved):

PR #448, opened 2026-08-07
Support Kimi Code CLI as a full memory source (discovery, MCP, references, skills)

summerfang-jolli:
Makes Moonshot's Kimi Code CLI (`@kimi-code/cli`, `~/.kimi-code`) a first-class memory source, at parity with Codex.

## What's included

1. **Discovered memory source** — `KimiSessionDiscoverer` scans `~/.kimi-code/sessions/<workDirKey>/<sessionId>/` (honouring `$KIMI_CODE_HOME`), recovers the working dir from `state.json.workDir`, and reads the conversation from `agents/main/wire.jsonl` (Kimi's own JSONL wire protocol — `turn.prompt` = user, `content.part` text = assistant). Wired through the `TranscriptSource` union end to end: config `kimiEnabled`, QueueWorker + ActiveSessionAggregator discovery, message-counter/loader/title dispatch, status/detection (incl. `getStatus` session counts), `jolli configure`, VS Code + IntelliJ UI.

2. **MCP server registration** — `jolli enable` writes Jolli's MCP server into `<KIMI_CODE_HOME>/mcp.json` (`mcpServers`, stdio `{command,args}` — verified against Kimi docs; MCP lives in `mcp.json`, not `config.toml`), so the Kimi agent can `recall`/`search` memory.

3. **References + skill usage** — `KimiEnvelopeParser` correlates `tool.call`/`tool.result` events. Kimi names MCP tools Claude-style (`mcp__server__tool`), so it reuses the Claude match path (`registry.match("claude", …)`) via a shared `McpBusinessNormalize` module (Claude behaviour stays byte-identical). **Any source definition carrying a generic `mcp__<server>__` prefix applies to Kimi with no per-source code — today Linear, GitHub, context7 and Jolli Memory (4 of 12).** The other eight only match the `mcp__claude_ai_*__` first-party-connector namespace (Asana, Confluence, Jira, monday, Notion, Slack, both Zoom), which a Kimi install cannot structurally produce — a known gap; covering one means adding a generic prefix pinned to a real Kimi capture of that server. `KimiSkillScanner` captures the observed `Skill` tool call. Both run via a `KimiDiscovery` orchestrator on the post-commit worker and the VS Code 60s tick (Codex + Kimi now share one discovery deadline rather than running serially).

## Format provenance

Every wire shape was pinned against **real `~/.kimi-code` captures**, not guessed — including deliberately generating a real MCP tool call and a real Skill invocation to confirm naming and payload shapes.

## Verification

- Parser/scanner run against real captured sessions: real MCP call → reference extracted; real Skill call → skill extracted.
- Full Linear pipeline: `mcp__linear__get_issue` → a proper Linear reference (id/title/url); `mcp__linear__list_issues` correctly denied.
- Gate: typecheck + lint clean; CLI tests + coverage above the 97/96/97/97 floor; VS Code tests; build green. IntelliJ compile + enum lockstep tests pass.

## Review round

Addressed a thorough review: Kimi `getStatus` session counts; `KimiSkillScanner` cursor-rewind so an in-flight `Skill` call's `ok` is corrected; `$KIMI_CODE_HOME` now honoured; shared discovery deadline; IntelliJ status HTML + `Kimi Code` labels; and this accurate scoping of which source definitions apply.


flyerli2025:
## Review

Reviewed the full diff locally against `main` and ran `npm run typecheck` (exit 0).

**Overall: high quality — recommend merging after the two Medium items below.** All three lockstep contracts are honoured: the Kotlin `TranscriptSource` enum plus both round-trip tests, the MCP registrar table plus the AGENTS.md table, and the exhaustive `SkillSource` label `Record`. I diffed the `McpBusinessNormalize` extraction line by line — it is a pure move plus a call-site convergence, and Claude's behaviour really is byte-identical (the `CONTEXT_NORMALIZER_IDS` gate stays at the call site, and `toolName` survives via the spread).

---

### Medium

**1. `getStatus()` never discovers Kimi sessions — the new status row always reports zero**

In `cli/src/install/Installer.ts` (~L1195-1300) every hookless source folds its on-demand scan into `allEnabledSessions` (codex, opencode, cursor, devin, cursor-cli, copilot, copilot-chat, cline, antigravity). Kimi does not. Consequences:

- `sessionsBySource.kimi` is permanently `undefined`, so `counts.kimi` in `StatusCommand.ts:492` and `StatusTreeProvider.ts` both come up empty. The Kimi row in `jolli status` renders as `detected & enabled` with no `(N sessions)` suffix — every sibling row has one.
- `filteredMostRecent` (the "most recent session" in `StatusInfo`) never sees a Kimi session at all. A Kimi-only user sees no recent activity in status.

Fix: add a block mirroring the antigravity one. `kimiDetected` is already computed earlier in the same function (~L1177), so it can be reused directly.

**2. `KimiSkillScanner` is missing the unpaired-call cursor rewind, so `ok` can be permanently wrong**

`ClaudeSkillScanner.ts:112-114` documents this explicitly: when the window closes mid-triple, `lastLine` is rewound to just before it, precisely so the optimistic `ok: true` can be corrected on the next pass. `KimiSkillScanner` advances `lastLine` to EOF unconditionally. Its own docstring says Kimi's events are cleaner than Claude's "so this scanner stays simple" — but the thing simplified away is exactly that correction mechanism.

Result: a `Skill` call whose `tool.result` has not landed when the scan window closes is recorded as successful forever, even if the result later carries `isError: true`. The VS Code 60s tick scans live sessions continuously, so hitting an in-flight call is fairly common.

Worth noting that `KimiEnvelopeParser` in the same PR **gets this right** (its `safeCursor` rewinds to the earliest unpaired call). Two new files in one PR give two different answers to the same problem — suggest lifting the `safeCursor` logic into the skill scanner.

---

### Low

**3. `$KIMI_CODE_HOME` is documented in a comment but not honoured by the code**

`HostRegistrars.ts:323` says the user-level file is `~/.kimi-code/mcp.json` "(or `$KIMI_CODE_HOME/mcp.json`)", but that comment is the only mention of the variable in the whole repo. Both the registrar and `KimiSessionDiscoverer` hardcode `homedir()/.kimi-code`. If a user sets it, MCP is written to the wrong file and discovery finds nothing — silently, on both sides. Either implement it or drop the parenthetical; the current state is the worst of the two, since the docs promise behaviour the code lacks.

**4. IntelliJ has two status surfaces; only one was updated**

`StatusPanel.kt` gained a Kimi row, but the status HTML in `JolliMemoryToolWindowFactory.kt:1477` enumerates Copilot CLI / Copilot Chat / Cline / Antigravity and omits Kimi.

**5. IntelliJ labels fall through to `"Kimi"` rather than `"Kimi Code"`**

`ConversationRowComponent.kt:281` (`displayName`) and `WorkingMemoryHtmlBuilder.kt:308` (`sourceLabel`) both hit `else -> replaceFirstChar { uppercase }`. The PR writes `"Kimi Code"` into the CLI's `TranscriptSourceLabel` and three VS Code label maps, so these two are inconsistent. (devin / antigravity have the same pre-existing gap, so this continues an existing pattern rather than introducing one — but since this change standardises on "Kimi Code" everywhere else, it would be nice to close it here.)

**6. Two 3 s discovery deadlines now run serially on the blocking post-commit path**

`CODEX_DISCOVERY_DEADLINE_MS = 3000` (`QueueWorker.ts:225`) carries a docstring stating it is "sized against the post-commit watcher's 15 s agent-session ceiling" — sized as though it were the only one. Codex and Kimi discovery are now awaited **sequentially**, so worst case 6 s of that window is consumed, and a third source would make it 9 s.

Not a correctness problem: `upsertReferenceEntry` runs under `withPlansLock`, and the VS Code 60s tick already fires both concurrently anyway. Suggest either racing both against one shared deadline — `Promise.race([Promise.all([...]), deadline(3000)])` — so the total budget stays 3 s, or at minimum renaming the constant and updating the docstring to say it is a *per-source* budget.

---

### Nits

- `normalizeMcpBusiness`'s return type `unknown | null` collapses to plain `unknown` in TypeScript, so the null case carries no type-level signal. The call site's runtime `canonical === null` check works, but the type does not enforce it.

- The PR description's claim that "**every existing source definition (Linear, GitHub, Jira, Notion, …) applies to Kimi with no per-source code**" is **not accurate**, and it seems worth calling out separately. Only 4 of the 12 definitions carry a generic, non-`claude_ai_` prefix: `linear` (`mcp__linear__`), `github` (`mcp__github__`), `context7` (`mcp__context7__`) and `jollimemory` (`mcp__jollimemory__`). The other eight — Asana, Confluence, Jira, monday, Slack, zoom-meeting, zoom-doc, Notion — only match `mcp__claude_ai_*__`, which is the claude.ai first-party connector namespace and something a Kimi install cannot structurally produce. The Linear tool that was verified end-to-end happens to be one of the four.

  The code itself is fine (a non-match is simply a non-match, and harmless), but the claim will end up in AGENTS.md and mislead whoever reads it next. Covering the rest would mean adding generic prefixes to those eight definitions — the pattern already exists, since `zoom-doc` / `zoom-meeting` disambiguate via `acceptSuffix` under a shared prefix. At minimum, suggest rewording the PR description and AGENTS.md to "any definition carrying a generic `mcp__<server>__` prefix applies to Kimi automatically", and recording the eight as a known gap.

summerfang-jolli:
Thanks for the thorough review — all findings were valid and are addressed in a409b92a (gate re-run green: CLI 10570 tests @ 99.14/97.70/99.05/99.35, VS Code 4442, IntelliJ compile + enum tests, build).

**Medium**
1. **`getStatus()` Kimi count** — added the on-demand Kimi scan into `allEnabledSessions` (mirrors the Codex block); `sessionsBySource.kimi` now populates, so the status row shows its `(N sessions)` and `filteredMostRecent` sees Kimi.
2. **Skill cursor rewind** — `KimiSkillScanner` now rewinds the cursor to before the earliest unpaired `Skill` call, so an in-flight call's optimistic `ok:true` is corrected on the next pass (same mechanism as `ClaudeSkillScanner` and this PR's `KimiEnvelopeParser.safeCursor`). Added tests for the in-flight case and the no-rewind-once-paired case, and updated the docstring that previously claimed the correction was simplified away.

**Low**
3. **`$KIMI_CODE_HOME`** — implemented, not dropped: a shared `kimiCodeHome()` helper now backs both `KimiSessionDiscoverer` and the mcp.json registrar (with an end-to-end discovery test under a relocated home).
4/5. **IntelliJ** — added the Kimi row to `JolliMemoryToolWindowFactory`'s status HTML and `"Kimi Code"` to `ConversationRowComponent.displayName` + `WorkingMemoryHtmlBuilder.sourceLabel`.
6. **Serial deadlines** — Codex + Kimi discovery now run concurrently against one shared deadline (renamed `ARTIFACT_DISCOVERY_DEADLINE_MS`), so the post-commit budget stays bounded regardless of how many hookless sources exist.

**Nits**
- **Coverage claim (b)** — you're right; corrected the PR description and AGENTS.md to "any definition carrying a generic `mcp__<server>__` prefix applies to Kimi" (linear/github/context7/jollimemory — 4 of 12), and recorded the eight `mcp__claude_ai_*__` connector sources as a **known gap**. I deliberately did *not* add generic prefixes for those eight: I only pin a matcher against a real captured tool-naming (as the Linear path was verified), and I haven't captured those servers under Kimi — so guessing their prefixes would repeat the "code and fixtures that agree with each other and nothing real" failure the skill scanners already warn about. Tracked as follow-up.
- **`normalizeMcpBusiness` return type (a)** — acknowledged, left as `unknown | null`: the identity path genuinely returns the raw payload (`unknown`), so a narrower non-null type would be a lie; both call sites runtime-check `=== null`.

flyerli2025 (approved):

PR #452, opened 2026-08-09
dashboard: emit web-local UI telemetry from the local web view

summerfang-jolli:
## What

Wires the local dashboard web view as its own `web-local` telemetry surface, sharing one event catalog with the future hosted `web` frontend.

## Why

The local dashboard was a telemetry blind spot: launching it emits `command_invoked`, but nothing recorded what a user did once the web view was open (which pages, which time range, which chart splits).

## Changes

- **Events**: register `dashboard_opened`, `range_changed`, `chart_split_changed` (reuse `view_switched`); mirror into the Kotlin registry; regenerate `TELEMETRY.md`.
- **Surface**: `trackAs(surface, …)` overrides only the surface dimension, so browser-forwarded events are `web-local` while the hosting process's own events stay `cli`.
- **Endpoint**: `POST /api/telemetry`, placed **before** the mutation-token gate (the token is inlined only into write-surface pages, so a gated beacon would drop events from the stats/memories pages). Cross-site is already blocked by the host/Origin/`frame-ancestors` layers. Bad input always answers 204 — a beacon is never taught to retry.
- **Server lifecycle**: `ServerTelemetry` primes telemetry and runs a periodic flush for the long-lived server process — otherwise `trackAs` is a no-op and the buffer is never drained.
- **Browser**: `JD.track` beacon (`sendBeacon`, survives navigation) wired to `dashboard_opened` (once per session), `view_switched`, `range_changed`, `chart_split_changed`. All content-free — bucketed values and fixed discriminators only.

## Verification

`tsc` + biome clean; 134 telemetry tests green (incl. `trackAs`, `/api/telemetry`, `ServerTelemetry`, `ServerEntry`, and the `TELEMETRY.md` up-to-date check).

## Dependency

`web-local` events are **dropped at ingest** until jolliai/jolli#1247 (surface allowlist) merges. Land that first.

### DIFF
--- a/cli/src/commands/DashboardCommand.ts
+++ b/cli/src/commands/DashboardCommand.ts
@@ -50,12 +50,20 @@ export interface DashboardOptions {
 export interface DashboardDeps {
 	readonly configDir?: string;
 	readonly dbPath?: string;
-	readonly spawnServer?: (port: number | undefined) => void;
+	/**
+	 * Spawns the detached server. `cwd` is the launcher's resolved repo directory
+	 * (`--cwd` or the process cwd): the child runs there so its telemetry buffer
+	 * lands under that repo's `.jolli` rather than wherever the launcher happened
+	 * to be invoked from.
+	 */
+	readonly spawnServer?: (port: number | undefined, cwd: string) => void;
 	readonly openBrowser?: (url: string) => Promise<void>;
 	readonly fetchHealth?: (port: number) => Promise<{ ok: boolean; pid?: number }>;
 	readonly killPid?: (pid: number) => boolean;
 	readonly now?: () => number;
 	readonly sleep?: (ms: number) => Promise<void>;
+	/** Launcher's resolved repo directory, threaded to {@link spawnServer}. Defaults to `process.cwd()`. */
+	readonly cwd?: string;
 }
 
 /* v8 ignore start -- default seams do real process/network/browser work; tests inject fakes */
@@ -73,7 +81,7 @@ async function defaultFetchHealth(port: number): Promise<{ ok: boolean; pid?: nu
 	}
 }
 
-function defaultSpawnServer(port: number | undefined): void {
+function defaultSpawnServer(port: number | undefined, cwd: string): void {
 	// Resolve the entry by directory + filename, not import.meta.url alone —
 	// same rationale as QueueWorker.launchWorker: this function gets inlined
 	// into whichever bundle imports it, and the sibling file is the contract.
@@ -87,6 +95,10 @@ function defaultSpawnServer(port: number | undefined): void {
 	const child = spawnHidden(process.execPath, [scriptPath], {
 		detached: true,
 		stdio: "ignore",
+		// Run the server in the launcher's resolved repo dir so its telemetry
+		// buffer (ServerTelemetry uses process.cwd()) lands under that repo's
+		// `.jolli`, not wherever `jolli dashboard` was invoked from.
+		cwd,
 		env: {
 			...process.env,
 			...(port !== undefined ? { JOLLI_DASHBOARD_PORT: String(port) } : {}),
@@ -133,7 +145,7 @@ function resolveSeams(deps: DashboardDeps): ResolvedSeams {
 /** {@link DashboardDeps} with every optional seam filled in. */
 interface ResolvedSeams {
 	readonly configDir: string;
-	readonly spawnServer: (port: number | undefined) => void;
+	readonly spawnServer: (port: number | undefined, cwd: string) => void;
 	readonly openBrowser: (url: string) => Promise<void>;
 	readonly fetchHealth: (port: number) => Promise<{ ok: boolean; pid?: number }>;
 	readonly killPid: (pid: number) => boolean;
@@ -265,7 +277,7 @@ export async function ensureServerRunning(
 		throw new Error("Another process is starting the dashboard but it never became healthy.");
 	}
 	try {
-		seams.spawnServer(requestedPort);
+		seams.spawnServer(requestedPort, deps.cwd ?? process.cwd());
 		for (let waited = 0; waited < STARTUP_TIMEOUT_MS; waited += 250) {
 			await sleep(250);
 			const state = await readDashboardState(configDir);
@@ -637,7 +649,7 @@ export async function executeDashboard(
 
 	let state: DashboardServerState;
 	try {
-		state = await ensureServerRunning(port, { ...deps, configDir: seams.configDir });
+		state = await ensureServerRunning(port, { ...deps, configDir: seams.configDir, cwd });
 	} catch (err) {
 		console.error(`\n  Error: ${errMsg(err)}\n`);
 		return false;
--- a/cli/src/commands/DashboardCommand.ts
+++ b/cli/src/commands/DashboardCommand.ts
@@ -21,10 +21,11 @@
  * open v11 is pure ESM and the VS Code bundle is CJS.
  */
 
-import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { existsSync, mkdirSync, readFileSync, rmSync, statSync, writeFileSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import type { Command } from "commander";
+import { getProjectRootDir } from "../core/GitOps.js";
 import { getGlobalConfigDir } from "../core/SessionTracker.js";
 import { type BackfillProgress, backfillRepos } from "../dashboard/Backfill.js";
 import { opportunisticSnapshot } from "../dashboard/Backup.js";
@@ -104,6 +105,11 @@ function defaultSpawnServer(port: number | undefined, cwd: string): void {
 			...(port !== undefined ? { JOLLI_DASHBOARD_PORT: String(port) } : {}),
 		},
 	});
+	// A detached spawn emits `error` asynchronously (e.g. a cwd that vanished
+	// between the resolve and the spawn); with no listener Node re-throws it as an
+	// uncaught exception and kills the launcher. Swallow it — a failed spawn means
+	// no dashboard, and the /health probe already reports that to the user.
+	child.on("error", (err) => log.warn("dashboard server failed to spawn: %s", errMsg(err)));
 	child.unref();
 }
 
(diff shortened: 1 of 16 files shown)


cli/src/commands/DashboardCommand.ts
```diff
@@ -87,6 +95,10 @@ function defaultSpawnServer(port: number | undefined): void {
 	const child = spawnHidden(process.execPath, [scriptPath], {
 		detached: true,
 		stdio: "ignore",
+		// Run the server in the launcher's resolved repo dir so its telemetry
+		// buffer (ServerTelemetry uses process.cwd()) lands under that repo's
+		// `.jolli`, not wherever `jolli dashboard` was invoked from.
+		cwd,
```
flyerli2025, line 102: `spawnHidden` returns a bare child and only `unref()` is called — no `error` listener. With `cwd` now supplied from `options.cwd ?? process.cwd()`, `jolli dashboard --cwd /nonexistent` (or a path that is a file) makes `spawn` emit an unhandled `error` event and the launcher crashes with `Error: spawn ... ENOENT` (reproduced locally). Previously the child inherited a valid cwd and a bad `--cwd` was tolerated (`registerRepo` catch logs it).

Suggest attaching `child.on("error", () => {})` (or failing with a clear message), and validating/normalizing the cwd before spawn.

cli/src/commands/DashboardCommand.ts
```diff
@@ -637,7 +649,7 @@ export async function executeDashboard(
 
 	let state: DashboardServerState;
 	try {
-		state = await ensureServerRunning(port, { ...deps, configDir: seams.configDir });
+		state = await ensureServerRunning(port, { ...deps, configDir: seams.configDir, cwd });
```
flyerli2025: `cwd` here is the raw `options.cwd ?? process.cwd()` (line 633) with no git-top-level normalization, but the telemetry buffer identity is the literal cwd (TelemetryBuffer.ts, JOLLI-1957): a repo root and one of its subdirectories are two separate buffers. So launching from `repo/subdir` writes the server's web-local events to `repo/subdir/.jolli/jollimemory/telemetry-queue.ndjson` — not "that repo's .jolli" as the spawn comment claims — and only the server's own 60s flush ever drains it; a SIGKILL/crash strands them. Other surfaces (hooks, QueueWorker, `jolli compile`, VS Code) drain the repo-root buffer. Resolve to the git top level before spawning (e.g. reuse `resolveProjectDir()` from CliUtils, as CleanCommand's `--cwd` does).

cli/src/dashboard/assets/js/stats.js
```diff
@@ -979,6 +979,7 @@ window.JD = window.JD || {};
 		document.querySelectorAll(".chips .chip[data-dim]").forEach((chip) => {
 			chip.onclick = () => {
 				JD.dimension = chip.getAttribute("data-dim");
+				JD.track("chart_split_changed", { card: "spend", split: JD.dimension });
```
flyerli2025: This `card: "spend"` branch can never fire: no element with `data-dim` is rendered anywhere (the `GROUPS` list at line 15 is unused — the "Cost & tokens" group-by chips are gone), so the selector at line 979 never matches. Meanwhile TELEMETRY.md and both registries document `spend` as a `chart_split_changed` card discriminator. Either remove the dead branch or restore the chips.

cli/src/dashboard/assets/js/main.js
```diff
@@ -20,5 +20,18 @@
...
 		window.JD.startRefresh(render);
+		/* Fire once per browser SESSION — nav is a full page reload, so a naive
+		   call would re-fire on every page. `first_run` is the first-ever open in
+		   this browser profile: a content-free proxy for a fresh install. */
+		try {
+			if (window.JD.track && !sessionStorage.getItem("jdOpened")) {
+				sessionStorage.setItem("jdOpened", "1");
+				var everOpened = !!localStorage.getItem("jdEverOpened");
+				localStorage.setItem("jdEverOpened", "1");
+				window.JD.track("dashboard_opened", { first_run: !everOpened });
```
flyerli2025, line 31: TELEMETRY.md documents `first_run` as "first open on this install", but this is per browser profile and per origin — localStorage is keyed by origin including the port, and the dashboard may run on 1818 or 18118, so switching ports, browsers, or clearing storage re-reports `first_run: true`. Align the doc with the implementation ("first open in this browser profile") or make the flag machine-scoped.

cli/src/dashboard/ServerEntry.ts
```diff
@@ -41,20 +42,30 @@ export async function runServerEntry(
...
+	// so the `/api/telemetry` beacon's `web-local` events are actually recorded
+	// and shipped (see ServerTelemetry). Best-effort; never blocks serving.
+	const telemetry = await startServerTelemetry();
 	// Both teardown paths clear the state file guarded by our own pid: if the
 	// record has moved on to another server, it is that server's now and deleting
-	// it would leave a live dashboard no launcher can find.
+	// it would leave a live dashboard no launcher can find. The final telemetry
+	// flush runs first so a clean shutdown ships what the last interval did not.
+	const finish = async (): Promise<void> => {
+		await telemetry.stop();
```
flyerli2025, line 54: `telemetry.stop()` → `flushTelemetryNow(cwd)` with no `timeoutMs`, so the flusher's default 10s network timeout applies. On idle shutdown or SIGTERM, `exit(0)` waits for that and `dashboard.json` isn't cleared until after — a stale-looking record during the wait. The CLI command-exit flush (JOLLI-1955) passes a short timeout; consider passing one here too.

cli/src/dashboard/DashboardServer.ts
```diff
@@ -469,6 +479,35 @@ async function readJsonBody(req: IncomingMessage): Promise<unknown> {
...
+	try {
+		const body = await readJsonBody(req);
+		if (typeof body === "object" && body !== null) {
+			const b = body as Record<string, unknown>;
+			const event = typeof b.event === "string" ? b.event : "";
+			const properties =
+				typeof b.properties === "object" && b.properties !== null
+					? (b.properties as Record<string, unknown>)
+					: {};
+			if (isTelemetryEventName(event)) trackAs(WEB_LOCAL_SURFACE, event, properties);
```
flyerli2025: The handler validates only the event name against the full registry, so a same-origin script (or any local process — a request without an Origin header passes `hasForeignOrigin`) can forge any registered event (e.g. `search_performed`) stamped `web-local`. The comment says the endpoint forwards "ONE content-free event from the local web view"; restricting to an allowlist of the four dashboard events would match that contract.

cli/src/dashboard/assets/js/shell.js
```diff
@@ -310,6 +312,11 @@ window.JD = window.JD || {};
 				return;
 			}
 			button.onclick = () => {
+				/* Emit the user-facing label token (7d/30d/90d), not the internal
+				   DashboardRange token (week/month/3m), so the telemetry discriminator
+				   matches the registry doc and the button labels. `custom` is handled
+				   by its own branch above and reported as "custom" on Apply. */
+				JD.track("range_changed", { range: { week: "7d", month: "30d", "3m": "90d" }[value] || value });
```
flyerli2025: Clicking the already-active range button (and likewise an already-active split at stats.js:1011/1029) still emits `range_changed`/`chart_split_changed` even though nothing changed. A `value !== activeRange` / current-split guard would keep the events honest.

flyerli2025 (changes requested):
## Review: web-local dashboard telemetry

Reviewed live head `d06332bd` (`telemetry-web-local`). Validated with Node 24.10.0: core telemetry tests (97), full `DashboardServer` suite (80), plus `ServerTelemetry`/`ServerEntry`/`DashboardCommand` all pass; the TS/Kotlin/TELEMETRY.md registries are consistent and the doc up-to-date check passes.

The design is sound: `trackAs` overrides only the surface dimension, the beacon sits behind the host-allowlist + Origin + `frame-ancestors` gates ahead of the mutation-token check, and the browser payloads are content-free fixed discriminators.

Findings, most important first. Two are worth fixing before merge (#1 is a new crash path); the rest are cleanup.

1. **Medium — invalid `--cwd` now crashes the launcher** (see inline). `spawnHidden` gets `cwd` but no error handler is attached; a nonexistent/non-directory `--cwd` makes `spawn` emit an unhandled `error` (ENOENT) and the whole `jolli dashboard` process dies with a stack trace (reproduced locally). Before this change the child inherited a valid cwd and the bad flag was tolerated.
2. **Medium — subdir launches still don't land the telemetry buffer under the repo root** (see inline). `cwd = options.cwd ?? process.cwd()` is used verbatim; per the JOLLI-1957 buffer contract the literal cwd IS the buffer identity. Launching from `repo/subdir` writes the server buffer to `repo/subdir/.jolli/jollimemory/telemetry-queue.ndjson`, which no other surface (hooks/QueueWorker/compile/VS Code) drains — events are only shipped by the server's own 60s flush and can be stranded on SIGKILL/crash.
3. **Low — the `card: "spend"` branch of `chart_split_changed` is unreachable** (see inline). Nothing renders `data-dim` (the `GROUPS` list is unused), so this handler can never fire while TELEMETRY.md documents `spend` as a card discriminator.
4. **Low — `first_run` doc drift** (see inline). TELEMETRY.md says "first open on this install", but it's per browser profile AND per origin (port-scoped localStorage, 1818 vs 18118), so it repeats across browsers/ports/storage clears.
5. **Low — final flush can delay shutdown up to ~10s** (see inline). `telemetry.stop()` calls `flushTelemetryNow` without `timeoutMs`, inheriting the flusher's 10s default; idle shutdown/SIGTERM blocks on it and `dashboard.json` isn't cleared until after.
6. **Low — beacon accepts any registered event name** (see inline). `handleTelemetry` only checks `isTelemetryEventName`, so a same-origin script (or an Origin-less local process) can forge other registered events stamped `web-local`. Restrict to the four dashboard events to match the documented contract.
7. **Low — "changed" events fire on no-op clicks** (see inline). Clicking the already-active range button or already-active split emits `range_changed`/`chart_split_changed` without any state change.

cli/src/commands/DashboardCommand.ts
summerfang-jolli: Fixed in a8390b6f. `resolveServerCwd` validates the target before spawning — a nonexistent path or a file falls back to `process.cwd()`, which is always valid — and `defaultSpawnServer` now attaches `child.on("error", …)` so an async spawn failure is logged rather than re-thrown as an uncaught exception that kills the launcher.

cli/src/commands/DashboardCommand.ts
summerfang-jolli: Fixed in a8390b6f. `resolveServerCwd` resolves the raw `--cwd`/process cwd to the git top level via `getProjectRootDir` before spawning, so the detached server shares the one repo-root buffer that hooks/QueueWorker/`jolli compile`/VS Code drain instead of stranding events in `repo/subdir/.jolli`. A non-repo dir (or any git failure) keeps the validated base dir.

cli/src/dashboard/assets/js/stats.js
summerfang-jolli: Fixed in a8390b6f. Removed the unreachable `card:"spend"` track (nothing renders `data-dim` — the `GROUPS` list is unused) and dropped `spend` from the `chart_split_changed` card discriminator in both registries and TELEMETRY.md, so the doc now matches what actually fires (`tokens`/`mcp`).

cli/src/dashboard/assets/js/main.js
summerfang-jolli: Fixed in a8390b6f (doc). `dashboard_opened`s `first_run` now reads "first open in this browser profile; per-origin localStorage, so it re-reports across ports, browsers, or a storage clear" in the TS + Kotlin registries and TELEMETRY.md — aligned with the implementation rather than claiming per-install.

cli/src/dashboard/ServerEntry.ts
summerfang-jolli: Fixed in a8390b6f. The server flush now runs with `timeoutMs: SERVER_FLUSH_TIMEOUT_MS` (2s) via the default flush in `ServerTelemetry`, so idle shutdown / SIGTERM is not held on the flusher’s 10s default before `exit(0)` and `dashboard.json` clears promptly.

cli/src/dashboard/DashboardServer.ts
summerfang-jolli: Fixed in a8390b6f. `handleTelemetry` now gates on a `WEB_LOCAL_EVENTS` allowlist (the four dashboard events) in addition to `isTelemetryEventName`, so a same-origin script or an Origin-less local process can no longer forge another registered event (e.g. `search_performed`) stamped `web-local`. Added a test asserting `search_performed` is dropped (204, nothing buffered).

cli/src/dashboard/assets/js/shell.js
summerfang-jolli: Fixed in a8390b6f. `range_changed` is now gated on `value !== activeRange`, and the tokens/mcp split handlers on a real change of the current split, so clicking the already-active range or split emits nothing.

summerfang-jolli:
Thanks @flyerli2025 — all seven findings were valid; addressed in a8390b6f.

| # | Finding | Fix |
|---|---------|-----|
| 1 | invalid `--cwd` crashes the launcher | `resolveServerCwd` validates the dir before spawn (bad `--cwd` → `process.cwd()`); `defaultSpawnServer` adds a `child.on("error")` guard |
| 2 | subdir buffer not under repo root | `resolveServerCwd` resolves to the git top level (`getProjectRootDir`), so the server shares the repo-root buffer other surfaces drain |
| 3 | `card:"spend"` unreachable | removed the dead track; dropped `spend` from the discriminator in both registries + TELEMETRY.md |
| 4 | `first_run` doc drift | doc now says "first open in this browser profile; per-origin" |
| 5 | shutdown blocks ~10s on final flush | server flush bounded to `SERVER_FLUSH_TIMEOUT_MS` (2s) |
| 6 | beacon accepts any registered event | `handleTelemetry` restricted to a `WEB_LOCAL_EVENTS` allowlist (+ test) |
| 7 | "changed" events fire on no-op clicks | `range_changed`/`chart_split_changed` gated on a real change |

Verified locally: `tsc` clean, biome clean, 147 tests pass across the affected files (incl. new `resolveServerCwd` cases, the anti-forging test, and the TELEMETRY.md up-to-date check). Inline replies left on each thread.

Re: the open question — `surface="web-local"` **is** allowlisted at ingest via backend PR jolliai/jolli#1247 (green). Until that merges + deploys, `web-local` events flush but are dropped, which is the merge-order dependency noted in the PR description.

flyerli2025 (approved):

PR #473, opened 2026-08-12
Emit onboarding_progressed from the Claude and Codex plugin SessionStart hooks

summerfang-jolli:
## Summary

The claude-plugin and codex-plugin surfaces have never emitted the `onboarding_progressed` telemetry event, so every install of either plugin was invisible to the onboarding funnel (in-git-repo / repo-enabled / capture-configured / memories-generated). This PR adds the emit to both plugins' SessionStart bootstraps **and** to `jolli enable --repo-hooks-only` — the mode both plugins' `/jolli:init` runs — with a bounded flush so a one-shot hook process neither loses the event nor stalls session startup.

## Motivation

Both `PluginBootstrapHook` and `CodexPluginBootstrapHook` install repo hooks via `install(..., { repoHooksOnly: true })`, which returns before the report tail that fires `onboarding_progressed` for every other trigger site (`EnableCommand`, `GuidedFrontDoor`, `StatusCommand`, `IngestRunStore`). Neither hook ever bootstrapped telemetry in-process either — they only called `flushTelemetryNow` to drain whatever `Cli.js` already buffered — so there was no path by which either plugin's automatic SessionStart hook could ever `track()` this event. `enable --repo-hooks-only` had the same blind spot: its early-returning branch never reaches `reportEnableResult`.

## Changes

- **Both plugin SessionStart hooks emit the snapshot** (`PluginBootstrapHook.ts` / `CodexPluginBootstrapHook.ts`), on both the success and failure branch of repo-hook reconciliation — a failed install is exactly the "installed but never got into a working state" drop-off this snapshot exists to surface. Events carry `surface: "claude-plugin"` / `"codex-plugin"`.
- **`jolli enable --repo-hooks-only` emits it too** (`EnableCommand.ts`), on both branches, awaited inline. Its explicit flush replaces the exit flush that `markSkipExitFlush()` disarms for this mode.
- **Shared helper**: the bootstrap→emit→flush chain lives once in `PluginBootstrapTelemetry.ts` (three callers). The config is read once and shared through the `deps` seams (pinned by `toHaveBeenCalledTimes(1)`).
- **Bounded flush**: one `BOUNDED_FLUSH_BUDGET_MS` (2 s) budget covers all three call sites — a whole-flush deadline on `performance.now()` (immune to NTP correction / laptop resume), started before the briefing build so the network wait overlaps it.
- **Snapshot ordering**: on the Claude hook's success path the chain starts only after `ensurePluginDefaultProvider` seeds the capture route — any earlier and a fresh install's first snapshot misreports `capture_method: "none"` for a state the same run repairs. The hook awaits `done` in a `finally`, so no throw or early return can orphan the in-flight chain. Paths that skip the seeding (manual disable re-check, `claudeEnabled` off, context-lock contention) still snapshot via the finally-fallback with the unseeded config — the tests assert these paths **do** snapshot.
- **Shared enabled predicate**: `isGitPipelineFullyInstalled` was extracted for the new light status probe (`resolveFunnelStatusLight`, which reads only the two fields the snapshot needs instead of paying for a full `getStatus()`) and is now also used by `GuidedFrontDoor`.
- **Docs**: `TELEMETRY.md` and spec 312 updated; spec 57's `--repo-hooks-only` passages realigned with the bounded flush (they previously said the flush is skipped).
- **Version bumps**: Claude plugin → 1.0.3, Codex plugin → 1.0.2 (both strictly higher than the published 1.0.2 / 1.0.1, so `publish-prod.sh`'s guard accepts them).

## User-visible behavior change

`GuidedFrontDoor` moved from `isGitHookInstalled` (checking `post-commit` alone) to `isGitPipelineFullyInstalled`. A repo carrying only `post-commit` — an interrupted upgrade or partial install — now reports **not enabled** at the front door and gets the enable prompt. This is the definition `getStatus()` has always used; the front door was previously computing a different `repo_enabled` (and dedup signature) from every other trigger site.

## Testing

- [x] Unit tests updated across `PluginBootstrapHook` / `CodexPluginBootstrapHook` / `PluginBootstrapTelemetry` / `EnableCommand` / `OnboardingFunnel` / `TelemetryFlusher` / `GuidedFrontDoor` / `GitHookInstaller` — both hook suites are symmetric (seeded-then-snapshot, lock-contended, host-disabled); the ordering assertion checks completion order so a `Promise.all` rewrite fails it
- [x] Full `npm run all` (clean → build incl. both plugins → typecheck → lint → test) green at the rebased head: cli unit + acceptance passing, vscode 4565 passing / 8 skipped, coverage above the enforced floors

## Checklist

- [x] Code follows the existing style
- [x] Documentation updated — `TELEMETRY.md`, spec 312, spec 57


Finn-coder2026:
Reviewed at `6036316d` (4 commits, 24 files). The follow-up commits close both items flagged in the earlier round, and the `getStatus()` one is fixed more thoroughly than was asked for.

## Verified fixed

- **`--repo-hooks-only` was the funnel blind spot that mattered most** — both plugins' `/jolli:init` runs exactly that mode. `EnableCommand.ts:457` now awaits `capturePluginOnboardingSnapshot(options.cwd).done` on both branches.
- **The lazy fallback no longer pays for a full `getStatus()`.** `resolveFunnelStatusLight` reads only the two fields the snapshot uses, in parallel. Extracting `isGitPipelineFullyInstalled` for it also fixed `GuidedFrontDoor`, which was checking `post-commit` alone and therefore computing a *different* `repo_enabled` (and dedup signature) from every other trigger site. Sharing the predicate is the right fix; copying the light status would have left that drift in place.
- **The flush is bounded.** `deadlineMs` as a whole-flush budget on `performance.now()` — the comment naming NTP correction and laptop resume as the reason not to use `Date.now()` is exactly right — started before the briefing build so the network wait overlaps it, with one `BOUNDED_FLUSH_BUDGET_MS` covering all three call sites.
- **The duplicated block is gone** (`PluginBootstrapTelemetry.ts`, three callers), **the config is read once** (`loadOnce` injected through both `deps` seams, pinned by `toHaveBeenCalledTimes(1)`), and **the ordering assertion can now actually fail** — it asserts completion order, and the comment states plainly that a `Promise.all` rewrite breaks it.
- **Both hook suites are symmetric now** (seeded-then-snapshot, lock-contended, host-disabled). The one remaining asymmetry is `sessionId`, which the Codex hook input genuinely doesn't carry — and it's asserted as `undefined` rather than left implicit.

I ran `tsc --noEmit -p cli/tsconfig.json` (exit 0) and `biome check --error-on-warnings` in `cli/` (971 files, clean) against this head. Worth knowing that the PR's own checks don't cover either — `gh pr checks` shows only CodeQL, `build-intellij`, `build-vscode` and `verify-dco`.

## Blocking

**1. `specs/57-cli-enable-command.md` now contradicts the code, in two places.**

- `:227` — "It also skips the short telemetry exit-flush so it adds no latency to a host's synchronous startup path."
- `:284` — "It prints nothing on success and skips the telemetry exit-flush to keep an automatic caller's startup fast."

That branch now flushes (bounded at 2 s). Spec 312 was updated for exactly this and says so — "its explicit `BOUNDED_FLUSH_BUDGET_MS`-capped flush replaces the exit flush that `markSkipExitFlush()` disarmed for this mode" — so the two specs currently assert opposite things, and 57 is the one that owns this command.

On the trade-off itself: re-arming a flush on a path that deliberately disarmed one is the right call here. A one-shot process gets no later tick, so the alternative is losing the event rather than delaying it, and 2 s is a real ceiling. It just has to be written down where the mode is specified.

**2. The PR description is still the first commit's.** Three commits and twenty files landed after it — `EnableCommand`, the shared helper, the light status probe, `deadlineMs`, spec 312, the version bumps — and none are mentioned. "Testing: 33/33 passing" and the unchecked "Documentation updated — not done; happy to add a `TELEMETRY.md` note if desired" are both stale (`TELEMETRY.md` and spec 312 are in the diff). It also still says the tests assert lock contention does *not* snapshot, which is the reverse of what the code and tests now do.

While updating it, please call out the user-visible behavior change that the current summary gives no hint of: `GuidedFrontDoor` moved from `isGitHookInstalled` to `isGitPipelineFullyInstalled`, so a repo carrying only `post-commit` (interrupted upgrade, partial install) now reports *not enabled* at the front door and gets the enable prompt. That is the correct definition — it is what `getStatus()` has always used — but a reader of this PR's stated scope wouldn't expect it.

## Checked, no action needed

**Plugin versions are fine.** `jolli-claude-plugin` currently publishes `1.0.2` and `jolli-chatgpt-plugin` publishes `1.0.1`, so `1.0.3` / `1.0.2` are both strictly higher and `publish-prod.sh`'s guard will accept them.

## Open, non-blocking

- **Backend `SURFACES` allowlist.** `Telemetry.ts` documents that the backend gates `surface` against its own closed list and drops unknown values there rather than here. `claude-plugin` / `codex-plugin` need to be in it or these events land nowhere — not verifiable from this repo, so please confirm against the backend before merging.
- **Disclosure.** `SessionStartHook.ts` is untouched, so two new emitting surfaces arrive without passing through the CLI's first-run telemetry notice. Fine to defer, but it is a real gap rather than an absent one.
- **Cross-process dedup.** `ledgerLocks` (`OnboardingFunnel.ts:213`) is an in-process `Map`, so two bootstraps racing in separate processes can both emit. Pre-existing, and spec 312 already documents a duplicate as the benign failure direction here.
- **Placement.** `PluginBootstrapTelemetry.ts` sits under `hooks/` but has a `commands/` caller; `core/` would read better. No functional cost — it adds no import weight, since `EnableCommand` already pulls in `OnboardingFunnel`.

Nothing in the code itself needs to change from my side.

Flyfly-1997 (approved):

summerfang-jolli:
Thanks for the thorough review. Both blocking items are addressed (note the branch was also rebased onto main since your pass at `6036316d`; the four commits are content-identical, now ending at `e4d2b1b6`, plus the fix below on top).

**1. Spec 57 realigned** — `4aa47cb0`. Both passages (`:227`, `:284`) now say the exit flush is disarmed by `markSkipExitFlush()` and *replaced* by the explicit onboarding-funnel snapshot flush capped at `BOUNDED_FLUSH_BUDGET_MS` (2 s), on both the success and failure branch — matching spec 312's wording and the code. The trade-off rationale is written down where the mode is specified: a one-shot process gets no later tick, so the alternative to the bounded wait is losing the event, not delaying it.

**2. PR description rewritten.** It now covers all the commits (`--repo-hooks-only` emit, the shared `PluginBootstrapTelemetry` helper, the bounded flush, seeding-then-snapshot ordering, `resolveFunnelStatusLight` / `isGitPipelineFullyInstalled`, docs, version bumps), fixes the reversed lock-contention statement (those paths **do** snapshot via the finally-fallback), refreshes the testing section (full `npm run all` green at the rebased head), and checks the documentation box. The `GuidedFrontDoor` behavior change you asked to be called out has its own section: a repo carrying only `post-commit` now reports not enabled at the front door and gets the enable prompt.

**On the backend `SURFACES` allowlist (your open item):** confirmed against the backend repo — `TelemetryRouter.ts`'s `SURFACES` set includes both `claude-plugin` and `codex-plugin`, committed on main 2026-08-09 (`0e4c0a939`, part of JOLLI-2069). The strings match what the plugins send. Only caveat is that this proves main has it, not that the currently deployed build postdates Aug 9.

The other non-blocking items (first-run disclosure, cross-process dedup, `PluginBootstrapTelemetry.ts` placement) are left as-is per your notes.

Finn-coder2026 (approved):

PR #525, opened 2026-08-20
Add the agent telemetry dimension: which AI host the work happened in

summerfang-jolli:
## What this adds

Telemetry recorded which of our own build artifacts sent an event (`surface`) and nothing about **which AI host the user was actually working in**. `surface` cannot answer that: it is a compile-time constant, so its value set equals the set of build configurations — six, against thirteen supported hosts. Ten of thirteen reported the surface of the observer, and the `run-hook` dist arbitration makes it actively misattributing (measured live during this work: Claude Code serving MCP out of the codex-plugin dist, and Codex out of the cursor-plugin dist).

This branch adds the second dimension — **`agent`**, inside the event `properties` bag (deny-listed server-side, so client-only by design) — plus two sibling closed-set properties, answering four questions:

| Question | Answer |
|---|---|
| Which host ran a CLI command | `agent` from env markers, opt-in per call site (9/13 hosts measured) |
| Which host made an MCP tool call | `agent` from the connection's own `initialize` clientInfo — the only signal that survives the shared per-worktree daemon |
| Who set a commit in motion | `trigger: agent/ui/terminal/unknown`, resolved at enqueue time in the post-commit hook and stamped on the queue entry |
| Was a command skill-driven | `via: skill:<name>`, consumed (read-then-deleted) from the env the skill recipe sets |

## Invariants, enforced not stated

- **Never defaulted** — absent reads as *unknown*; nothing maps to `cli` by omission.
- **Never free-form** — every value passes a closed-vocabulary gate (`TELEMETRY_AGENTS` aliases `TRANSCRIPT_SOURCES`, so a 14th host cannot ship a stale list); an unrecognised value is omitted, never passed through.
- **Never guessed** — two hosts' markers at once answer nothing; a family gate that cannot name its variant abandons the whole answer; an IDE Commit button beats an inherited marker; long-lived processes (extension host, `mcp-serve`, `ide-bridge`) never infer from env at all.

## Everything is measured, including the negatives

Every mapping row traces to a capture made during this work: env probes across nine hosts (agent-spawned shell dumps against a human-shell control), MCP `initialize` captures via a logging probe server (`claude-code`, `codex-mcp-client` — where "codex" would have been the obvious wrong guess), and a real-SDK end-to-end run of the built dist. The recorded negatives matter as much: `CURSOR_TRACE_ID` does not exist in any Cursor context (exposing a pre-existing dead entry in `CaptureProgress`, tracked separately), kimi sets no env var at all, Antigravity sets no `GEMINI_*` (a collision that would have relabelled its work as gemini's), and cursor-agent's clientInfo is a family name deliberately left unmapped until the IDE's own string is captured. devin's probe is quota-blocked and documented for a one-paste retry.

## Public disclosure

`TELEMETRY.md` discloses the new dimension in the same change, and its previously hand-written surface list — stale by three entries, invisible to the drift test because the staleness was in the generator's own string — is now generated and pinned. The watershed (`AGENT_DIMENSION_SINCE_VERSION = "0.99.14"`) is stamped into the doc and guarded by a test that fails if main's version overtakes it before release; **if this ships in a different release, update the constant and regenerate.**

## Explicitly out of scope

`surface` and its allowlist are untouched; no server change; no backfill (pre-watershed rows were never measured, and nothing reconstructs them); the outbound `LocalAgentToolId` dimension stays distinct; skill-attribution on the MCP path is designed and measured but deliberately deferred (the only carrier is model-relayed and can manufacture false positives — the design note records the per-connection-schema mechanism for when the data need is real).

## Verification

`npm run all` green at every commit (CLI 98.41/96.18/98.4/98.83 against the 97/96/97/97 floor; vscode 98.92/97.54/98.33/99.06 — never lowered). IntelliJ telemetry suite green (its Kotlin event-description mirror updated in lockstep; the port deliberately never emits `agent`, with the reasoning pinned by a Kotlin test under an inherited marker). End-to-end verified against the real MCP SDK with every env marker scrubbed — the buffered event carried `agent:"claude"` from the handshake alone.

### DIFF
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -88,7 +88,7 @@ disk **before** they are sent.
 | `ai_provider_selected` | User chose jolli vs anthropic for LLM. Props: provider (discriminator). |
 | `memory_bank_migrated` | Migrate-to-Memory-Bank run. Props: outcome, repos, entries_bucket. |
 | `onboarding_progressed` | Per-install onboarding-funnel snapshot, emitted from a repo context and deduped by state tuple (+ daily heartbeat). Content-free — answers 'after install, where do people stall'. Props: in_git_repo, repo_enabled, capture_configured, capture_method (discriminator: local-agent/anthropic/jolli/none), memories_generated, memories_bucket. |
-| `command_invoked` | Any CLI command ran (auto-emitted). Props: command (discriminator), ok, duration_ms; via (discriminator: skill:<name> from a closed skill-name set — present only when a Jolli skill's recipe invoked the command, omitted for a directly-typed one). MCP tool calls carry a `tool` property and are emitted per call (not per session); the session-level `command:"mcp"` event is suppressed. |
+| `command_invoked` | Any CLI command ran (auto-emitted). Props: command (discriminator), ok, duration_ms; via (discriminator: skill:<name> from a closed skill-name set — present when a Jolli skill's recipe invoked the command; absent means directly typed OR a pre-upgrade skill copy that predates the stamp, so absence is not proof of direct use). MCP tool calls carry a `tool` property and are emitted per call (not per session); the session-level `command:"mcp"` event is suppressed. |
 | `recall_performed` | A recall was run. Props: hit, result_count_bucket. |
 | `search_performed` | A search was run. Props: query_len_bucket, result_count_bucket. |
 | `memory_pushed` | Memories pushed to a Space. Props: kind, created, plans_bucket. |
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -88,7 +88,7 @@ disk **before** they are sent.
 | `ai_provider_selected` | User chose jolli vs anthropic for LLM. Props: provider (discriminator). |
 | `memory_bank_migrated` | Migrate-to-Memory-Bank run. Props: outcome, repos, entries_bucket. |
 | `onboarding_progressed` | Per-install onboarding-funnel snapshot, emitted from a repo context and deduped by state tuple (+ daily heartbeat). Content-free — answers 'after install, where do people stall'. Props: in_git_repo, repo_enabled, capture_configured, capture_method (discriminator: local-agent/anthropic/jolli/none), memories_generated, memories_bucket. |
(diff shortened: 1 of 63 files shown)


summerfang-jolli:
All three review findings verified against the code and fixed in 3464f9e9. Notes on each, including where the review's framing was refined by verification:

**1. ide-bridge-serve inference (P1)** — confirmed and fixed. One scoping note: the JVM UI events the bridge *forwards* were already safe (the `telemetry-track` action re-bootstraps without inference before tracking), so the actual leak was the bridge's own lifecycle events — still a real rule violation. Fixed with a separate `isAgentInferenceExempt` predicate rather than widening `isDetachedDaemonInvocation`, because that predicate also gates the one-time telemetry notice, which the bridge must keep (IntelliJ reads its stderr; the daemons' goes to /dev/null). Regression tests cover the exemption and the prefix-only non-matches.

**2. amend/rebase stamping (P2)** — confirmed, and the review was *more* right than stated: the amend entry is built in PostRewriteHook (the old→new mapping only exists on post-rewrite's stdin), so amend was never stamped anywhere. Origin is now resolved once at that hook's entry — `fromPluginUi` reuses the same plugin-source marker its `commitSource` read consumes, which IntelliJ's Squash action also writes — and stamped on both enqueue paths. Threading pinned by tests with a mocked resolver (the real one reads process env, which differs between dev machines and CI).

**3. via coverage (P2)** — confirmed, plus one site the review missed: the Claude plugin's *committed* recall/search copies, which have no generator and had no pin — exactly how they got skipped. Went with full coverage rather than narrowed semantics: all shipped run-cli invocation sites now carry their skill's bare-name prefix (prose-embedded commands included; `test -f` presence checks excluded — they never run a jolli command, so there is no event to annotate). Menu revisions move 7→8 / 8→9 preserving the strictly-greater contract, both bundle generators re-run, `SKILL_VIA_NAMES` now derives its lockstep pin from the union of all three skill registries, and the Claude bundle files get their own pin so this shape can't recur. TELEMETRY.md wording now states the rollout reality: absence = directly typed OR a pre-upgrade skill copy, not proof of direct use.

Verification: `npm run all` green (CLI 98.41/96.14 vs 97/96 floor), IntelliJ telemetry suite green, all three plugin-bundle drift tests green after regeneration.

jeverest (approved):

PR #558, opened 2026-08-30
feat: allow jollidev.com and jollidev.dev origins

flyerli2025:
## Summary

Adds `jollidev.com` and `jollidev.dev` to the Jolli origin allowlist and telemetry env map, keeping the three parser implementations (CLI TS, VS Code webview, IntelliJ Kotlin) in lockstep.

- **CLI**: `ALLOWED_JOLLI_HOSTS` + the `assertJolliOriginAllowed` error string in `JolliApiUtils.ts`; telemetry env mapping in `Telemetry.ts` (`.com` → prod, `.dev` → dev, consistent by TLD).
- **IntelliJ (Kotlin port)**: matching allowlist + telemetry entries in `JolliAuthUtils.kt` / `telemetry/Telemetry.kt`.
- **VS Code**: the webview validator inherits the list automatically via the inlined `ALLOWED_JOLLI_HOSTS`; this PR also fixes two drifted user-/dev-facing strings the migration missed — the allowlist error message (now lists the two new domains) and a stale lockstep comment that pointed at the retired `JolliApiClient.kt` instead of `JolliAuthUtils.kt`.
- **Docs**: `AGENTS.md` slimmed, with deep rationale moved into `cli/DEVELOPMENT.md` / `cursor-plugin/DEVELOPMENT.md`.

## Intentionally unchanged
- Request paths still trust the saved value — validation stays save-time only.
- `jollidev.com` → prod vs `jollidev.dev` → dev is asymmetric by design (keyed on TLD), asserted by the added telemetry tests.

## Test plan
- `cli/src/core/JolliApiUtils.test.ts` + `Telemetry.test.ts` — new-domain acceptance and env mapping.
- `intellij/.../JolliAuthUtilsTest.kt` + `TelemetryTest.kt` — Kotlin mirror.
- Run `npm run all` before merge.


cli/src/core/Telemetry.ts
```diff
@@ -319,8 +319,10 @@ export function resolveTelemetryEnv(origin?: string, env: NodeJS.ProcessEnv = pr
 	const matches = (h: string): boolean => host === h || host.endsWith(`.${h}`);
 	if (matches("jolli-local.me")) return "local";
 	if (matches("jolli.dev")) return "dev";
+	if (matches("jollidev.dev")) return "dev";
 	if (matches("jolli.cloud")) return "preview";
 	if (matches("jolli.ai")) return "prod";
+	if (matches("jollidev.com")) return "prod";
```
dougschroeder, line 328: **Should Fix:** After this change `prod` means two hosts (`jolli.ai` and `jollidev.com`), and `dev` likewise. The envelope carries `env` but no origin or host field, and `env` is a pure analytics dimension — it selects no endpoint or dataset. So the number this migration is measured by, how many clients have moved, is not answerable from `env`.

**Why it matters:** Worth settling while this map is being written, rather than when someone goes looking for the curve.

**Suggested fix:** The likely answer is server-side. `TelemetryFlusher.ts:79-105` resolves the target from the key's own tenant claim and POSTs to `<that origin>/api/telemetry/events`, so migrated and unmigrated clients already hit different addresses — if the two PROD domains are separate deployments the counts are split already, and if they share one, recording the receiving Host at ingest costs no client release and works for clients already in the field. A client-side field would only be readable from builds shipped after it, which excludes exactly the un-upgraded population being counted. If one is still wanted, keep it bucketed or boolean per `TELEMETRY.md`'s stated posture — never the raw tenant origin, whose subdomain identifies the customer.

Separately: `jollidev.com` → `prod` while `jollidev.dev` → `dev` is keyed on TLD, which reads backwards from the names. The PR description explains it; a one-line comment here would stop the next reader "fixing" it.

cli/DEVELOPMENT.md
```diff
@@ -942,3 +942,316 @@ it in old commits or stale notes, they're outdated.
...
+
+  Concretely, for working-area context (plans / notes / references):
+  - The rules live in [`cli/src/core/PlanService.ts`](cli/src/core/PlanService.ts), [`cli/src/core/NoteService.ts`](cli/src/core/NoteService.ts) and [`cli/src/core/references/ReferenceService.ts`](cli/src/core/references/ReferenceService.ts). `vscode/src/core/{Plan,Note}Service.ts` are **re-export shims** — adding a function to a shim instead of the CLI module makes it invisible to IntelliJ and re-opens the drift.
```
dougschroeder: **Nice to have:** The relocation carried repo-root-relative links into a file one directory down, so they now resolve as `cli/cli/src/...`. Measured: broken link targets in this file go from 18 before the PR (all stale `src/site/*` references) to 91 after; `cursor-plugin/DEVELOPMENT.md` goes from 0 to 5. This file's own pre-existing convention is `src/core/...`.

**Why it matters:** These sections are dense with `file:line` pointers, and that navigability is most of what makes them worth moving. A prefix strip fixes the bulk mechanically.

Also from this commit: three empty `###` headings at lines 1137, 1228 and 1248 sit directly above the moved block's own same-level heading, so they render as empty sections — drop the wrapper, or demote the moved heading to `####`.

dougschroeder (dismissed):
Reviewed against the surrounding auth and telemetry code. The change itself is correct and complete: all eight hand-maintained sites — three enforcement arrays, three user-facing messages, two telemetry classifiers — are updated in both languages with tests on each side, and the suffix-boundary rule is not weakened by either new domain (none of the six entries is a dot-suffix of another, so `x.jollidev.dev` cannot match `.jolli.dev`). Additive only, with `jolli.cloud` retained, which is the right posture for a migration with an overlap window. Approving.

Three things below: one decision worth making while this map is still being written, and two documentation items caused by this PR's own changes.

### Stale allowlist documentation

`vscode/DEVELOPMENT.md:261` still reads "the allowlist (`jolli.ai`, `jolli.dev`, `jolli.cloud`, `jolli-local.me`)" — accurate before this PR, wrong after it. `AGENTS.md` was updated correctly; this one was missed, and it is the document a VS Code contributor reads before editing `SettingsScriptBuilder.ts`, the file this PR had to patch twice.

The repository's behavioral documentation has the same problem in six more lines: the four-row allowlist table, the rejection sentence quoted verbatim as user-facing contract, two restatements in the tenant-resolution topic, and two in the settings-webview topic. Worth folding in here, since the second commit is already a documentation reorganization.

### Two things worth knowing for the release

Both are properties of the credential design rather than of this diff, but they bear on when this can safely ship:

- **This needs to be adopted by clients before the backend mints any key on the new hosts.** An older client that lacks these entries skips the off-allowlist tenant when resolving what to persist, falls back to the sign-in origin, and the tenant-symmetry check then clears a working key — on a save that reports success.
- **The VS Code extension needs a republish, not just the CLI.** `SettingsScriptBuilder.ts:334` inlines the array at extension build time, so an already-installed build keeps rejecting a new-host key in Settings even against an updated CLI.

Nice work on the coverage — the telemetry classifiers are easy to miss and were not, and `Api.bundle.test.ts` needed no edit because it derives its expectation from the source module rather than restating the host list.

cli/src/core/Telemetry.ts
flyerli2025: Thanks — agreed on both points.

**Measurement:** you're right that `env` can't separate the two hosts, and the answer is server-side. `TelemetryFlusher` already POSTs to the key's own tenant origin, so migrated and unmigrated clients hit different addresses today — if the two prod domains are separate deployments the counts are already split, and if they share one ingest, recording the receiving `Host` header there needs no client release and covers clients already in the field. Captured as a follow-up rather than folded into this PR, since the fix lives on the ingest side, not the client. If a client-side signal is ever added it'll stay bucketed/boolean per `TELEMETRY.md`, never the raw tenant origin.

**TLD pairing:** added a comment on both the TS classifier and its Kotlin mirror noting the `jollidev.com → prod` / `jollidev.dev → dev` pairing is keyed on TLD by design, so it isn't "fixed" to match the names (commit 14b10584b).

cli/DEVELOPMENT.md
flyerli2025: Fixed in c5ee242bb.

Stripped the root-relative prefixes the relocation carried in. In `cli/DEVELOPMENT.md`, `cli/…` links now use the file's own `src/…` convention, and `intellij/…`, `vscode/…`, `claude-plugin/…` targets are rewritten to `../…`. Resolving every link against the file's directory now gives **0 PR-introduced broken targets** (the 18 stale `src/site/*` references predate this PR, as you noted, and are left as-is). `cursor-plugin/DEVELOPMENT.md` is likewise back to 0 (was 5).

Also demoted the three moved headings — *Push channel: one watch list…*, *VS Code extension bundles the CLI*, *Pushable context kinds are table-driven* — to `####` so they nest under their `— full rules` wrapper instead of stacking as same-level headings with an empty section between them.

dougschroeder (approved):

PR #559, opened 2026-08-30
docs: prep 0.99.17 release notes and version bump

flyerli2025:
## Summary

Release prep for **0.99.17** — version bumps, changelogs, and a stale-doc fix.

The only functional change since 0.99.16 is the Jolli origin allowlist migration (`jollidev.com` / `jollidev.dev`, PR #558), which is user-invisible, so this release ships as a small maintenance bump.

## Changes

- **Version → 0.99.17** across all three artifacts: `cli/package.json`, `vscode/package.json`, `intellij/build.gradle.kts`. `package-lock.json` carries only the two workspace version fields, edited directly rather than through `npm` (npm's own resolution wants to drop unrelated `peer` markers).
- **Changelogs** — user-facing 0.99.17 entries on `cli/`, `vscode/`, and `intellij/` ("Ready for Jolli's new web addresses" + bug fixes). CLI/VS Code `Last synced commit` markers advanced to `cac3e49a2`.
- **`vscode/DEVELOPMENT.md`** — the shared origin allowlist listing had gone stale (missing `jollidev.com` / `jollidev.dev`); brought back in line with the code and `AGENTS.md`.

## Intentionally unchanged

- **READMEs** — the domain-allowlist change is invisible to their feature descriptions; no README lists the allowlist domains or hardcodes a version, so nothing there needed touching.
- **Root `package.json`** — stays at 0.99.0 (deliberately not tracking artifact versions).

## Notes

- This branch is built on top of the domain-migration commits (PR #558), so the diff overlaps until #558 merges; the 0.99.17 changelog line describes exactly that work.
- `npm run all` was **not** run for this prep (docs/version/lockfile only, no source-logic change; verified no test hardcodes the old version). CI runs the full chain.

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
-<!-- Last synced commit: fb34e0a98 | 2026-08-28 -->
+<!-- Last synced commit: e3287ce7 | 2026-08-31 -->
+
+## 0.99.17
+
+- **Ready for Jolli's new web addresses** — signing in and syncing now work with Jolli's new domains, so keys issued there are accepted.
+- Bug fixes.
 
 ## 0.99.16
 
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
-<!-- Last synced commit: fb34e0a98 | 2026-08-28 -->
+<!-- Last synced commit: e3287ce7 | 2026-08-31 -->
+
+## 0.99.17
+
+- **Ready for Jolli's new web addresses** — signing in now works with Jolli's new domains, so keys issued there are accepted.
+- Bug fixes.
 
 ## 0.99.16
 
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@jolli.ai/cli",
-	"version": "0.99.16",
+	"version": "0.99.17",
 	"description": "Automatic development memory for AI coding agents: turns your Claude Code, Codex, Gemini, Cursor, Copilot, Cline, Devin, OpenCode, Antigravity and Kimi Code sessions into structured commit summaries your agent can recall over MCP",
 	"main": "./dist/Api.js",
 	"types": "./dist/Api.d.ts",
--- a/intellij/CHANGELOG.md
+++ b/intellij/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## 0.99.17
+
+### Changes
+
+- **Ready for Jolli's new web addresses** — signing in now works with Jolli's new domains, so keys issued there are accepted.
+
+### Fixes & Improvements
+
+- Bug fixes.
+
+### Performance
+
+- Performance improvements.
+
 ## 0.99.16
 
 ### Changes
--- a/intellij/build.gradle.kts
+++ b/intellij/build.gradle.kts
@@ -17,7 +17,7 @@ plugins {
 }
 
 group = "ai.jolli"
-version = "0.99.16"
+version = "0.99.17"
 
 repositories {
     mavenCentral()
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
 		},
 		"cli": {
 			"name": "@jolli.ai/cli",
-			"version": "0.99.16",
+			"version": "0.99.17",
 			"hasInstallScript": true,
 			"license": "Apache-2.0",
 			"dependencies": {
@@ -7161,7 +7161,7 @@
 		},
 		"vscode": {
 			"name": "jollimemory-vscode",
-			"version": "0.99.16",
+			"version": "0.99.17",
 			"license": "Apache-2.0",
 			"dependencies": {
 				"@vscode/codicons": "^0.0.45",
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -2,7 +2,7 @@
 	"name": "jollimemory-vscode",
 	"displayName": "Jolli Memory",
 	"description": "VS Code sidebar for Jolli Memory — AI-powered commit summaries and smart git operations",
-	"version": "0.99.16",
+	"version": "0.99.17",
 	"publisher": "jolli",
 	"license": "Apache-2.0",
 	"icon": "assets/icon.png",


cli/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
-<!-- Last synced commit: fb34e0a98 | 2026-08-28 -->
+<!-- Last synced commit: cac3e49a2 | 2026-08-30 -->
```
dougschroeder: **Should fix:** `cac3e49a2` is not reachable from `main`.

#558 was rebase-merged, so main carries `2e693855` — identical content, different SHA. `cac3e49a2` exists only on this branch, and `git merge-base --is-ancestor cac3e49a2 origin/main` fails today. The marker it replaces (`fb34e0a98`) passes that same check.

**Why it matters:** the marker's only purpose is `git log <marker>..HEAD` for the next changelog sync. Once this merges, that command fails with `unknown revision` for anyone who does not have this branch, so the next release's author silently loses the boundary and has to guess where 0.99.17 ended.

**Suggested fix:** use main's own equivalent — `e3287ce7` (main's tip, one commit further along) or `2e693855` for the same content point. Rebasing first, per the review body, makes this value fall out naturally.

vscode/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
-<!-- Last synced commit: fb34e0a98 | 2026-08-28 -->
+<!-- Last synced commit: cac3e49a2 | 2026-08-30 -->
```
dougschroeder: **Should fix:** same unreachable SHA as `cli/CHANGELOG.md` — `cac3e49a2` exists only on this branch; main's equivalent is `2e693855`.

**Why it matters:** after merge, `git log cac3e49a2..HEAD` fails with `unknown revision`, so the marker cannot do the one job it has.

**Suggested fix:** `e3287ce7` (main's tip) or `2e693855`, and keep both changelogs on the same value.

dougschroeder (commented):
## Review: release prep for 0.99.17

Safe to ship once the sync-marker SHA is corrected and the branch is rebased. Nothing here is ship-blocking: the bump is consistent across all three artifacts and the lockfile, the deliberate non-bumps are respected, and the IntelliJ changelog is wired correctly against the gradle plugin. The two items below are both consequences of the same root cause — the branch predates #558's rebase-merge — and rebasing fixes one of them outright while making the marker's correct value obvious.

### A note on the diff you are looking at

#558 was **rebase**-merged, so `main` carries new SHAs (`535b66ee`…`e3287ce7`) for content this branch still holds under its original SHAs (`94520a18`…`c5ee242b`). The merge base is therefore `5b371658` and GitHub replays all of #558, showing 23 files / +514 −266. The actual delta against main's tip is **7 files / +31 −7**: the three artifact versions, two lockfile workspace fields, and the three changelog entries. The overlapping content is byte-identical to main's rebased copies, so there is no revert risk — but the review below is against those 7 files.

---

### Should fix

**1. Changelog sync marker points at a commit that is not on `main`** — see the inline comments on `cli/CHANGELOG.md` and `vscode/CHANGELOG.md`.

**2. Branch is not rebased onto `main`**

Five already-merged commits are replayed, which inflates the PR to 23 files. Two costs: approving this reads as re-approving #558, and the seven lines that are actually new are buried in ~780 lines of noise. A rebase onto `origin/main` plus a force-push reduces this to one screen, and makes the correct value for finding 1 obvious at the same time.

---

### What is good

- **The bump is genuinely complete.** `git grep 0.99.16` at the branch tip returns nothing outside the changelogs and the lockfile — no straggler in a README, a workflow, or a manifest.
- **All three deliberate non-bumps are respected**: root `package.json` stays at 0.99.0, and the Claude / Codex / Cursor `plugin.json` versions (1.0.6 / 1.0.5 / 1.0.3) are untouched.
- **The IntelliJ entry is wired correctly**, which is easy to get wrong: `## 0.99.17` matches `changelog { headerParserRegex }`, and the custom `### Changes` / `### Fixes & Improvements` / `### Performance` groups match `groups = emptyList()`, so `patchPluginXml.changeNotes` renders.
- **`__CLI_PKG_VERSION__` stays coherent** — the CLI, VS Code and Claude-plugin builds all inject it from `cli/package.json`, so dist-path selection and the MCP daemon handshake move together.
- **The lockfile hand-edit is correct and complete.** Those two workspace `version` fields are the only version carriers; the `node_modules/@jolli.ai/cli` and `node_modules/jollimemory-vscode` entries are bare links with no version field.
- **Skipping the full gate was checked rather than assumed** — no test pins the artifact version (the `0.99.1x` hits under `cli/src` are historical constants such as `AGENT_DIMENSION_SINCE_VERSION`).

### Worth verifying before tagging

1. `cd intellij && ./gradlew patchPluginXml`, then confirm the patched `plugin.xml` carries `<version>0.99.17</version>` and a rendered 0.99.17 block in `change-notes`.
2. `npm install` at the repo root, then `git diff --stat package-lock.json` — any churn means the hand edit disagrees with npm's own resolution.
3. `npm run all` on the rebased branch, coverage still at 97/96/97/97 for the CLI.
4. After merge, `git log <marker>..HEAD --oneline` returns a commit list rather than `fatal: unknown revision` — that is finding 1.

cli/CHANGELOG.md
flyerli2025: Fixed in 1e78788 — marker now points at e3287ce7 (main's tip, and this branch's actual merge base), verified reachable with `git merge-base --is-ancestor e3287ce7 origin/main`. Both changelogs carry the same value.

vscode/CHANGELOG.md
flyerli2025: Fixed in 1e78788 — same value as cli/CHANGELOG.md: e3287ce7, reachable from main.

flyerli2025:
Thanks for the thorough review. Both "Should fix" items are addressed in 1e78788:

**1. Sync marker SHA** — updated both `cli/CHANGELOG.md` and `vscode/CHANGELOG.md` from `cac3e49a2` to `e3287ce7` (main's tip). Confirmed `git merge-base --is-ancestor e3287ce7 origin/main` passes, so `git log <marker>..HEAD` will resolve for the next release author.

**2. Rebase** — the branch is now based directly on main's tip `e3287ce7`; the PR is down to the actual delta (1 commit, 7 files, +31 −7). The #558 replay is gone.

Deliberate non-bumps (root `package.json`, the three `plugin.json` versions) are untouched.

Finn-coder2026 (approved):

